PR #6, opened 2017-06-12
Fix nag issue with OpenMP in timedependent routines

aradi:


aradi:
Got answer from NAG with proposed workaround (as in the last commit). This should be the best we can get without complex code modification.

PR #10, opened 2017-06-13
Removed doc material now covered by recipes repository

bhourahine:
Cleaned out doc/modes doc/waveplot/howtos/ and
doc/waveplot/waveplot_in.hsd.

Not sure if tesserals.mws should also be removed.


aradi:
Tesserals should remain. At some point, we need maybe a doc/derivs folder with various mathematical derivations...

PR #22, opened 2017-07-25
Added a script to strain gen files

bhourahine:
Assumes they are oriented correctly in space.


aradi:
Great idea!

Why don't we generalise the interfaces somewhat, so that it can isotropically resize cells in one go? Like `straingen -c xx,yy,zz mygeo.gen 0.2` for uniformly enlarging the cell by 20 percent. (Probably, we should even make xx,yy,zz the default component). I think using symbolic names instead of numbers would make it more user friendly. Also, strain magnitude would fit more as argument than an option as it does not make much sense to leave it away.

On the implementation side: I'd suggest to make its structure more similar to the `repeatgen` script, so that we can easily test it using Pythons unit test. Also the pylint score obtained with `env PYTHONPATH=$PWD/tools/dptools/src pylint3 --rcfile utils/srccheck/pylint/pylintrc-3.ini tools/dptools/bin/straingen` is radically low, mostly due to PEP8 violation.

bhourahine:
Unfortunately, using the strain as an argument (see checked in version that does this) runs into the issues outlined at  https://stackoverflow.com/questions/8741840/optparse-with-integer-argument as the strain can be negative.

aradi:
We could also use a float scaling factor instead: 1.0 would not do anything, 0.98 decrease by 2% 1.02 increase by the same amount. Then, no negative numbers were necessary any more, although it would be then maybe more logical to call the script `scalegen`, as it is not the strain which you give as argument...

bhourahine:
Hi Bálint,

Unfortunately that doesn't work for shear stress.

Cheers

Ben


On September 4, 2017 8:54:11 AM GMT+01:00, "Bálint Aradi" <notifications@github.com> wrote:
>We could also use a float scaling factor instead: 1.0 would not do
>anything, 0.98 decrease by 2% 1.02 increase by the same amount. Then,
>no negative numbers were necessary any more, although it would be then
>maybe more logical to call the script `scalegen`, as it is not the
>strain which you give as argument...
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/dftbplus/dftbplus/pull/22#issuecomment-326891466

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

aradi:
Of course, you are right, it won't work.

bhourahine:
OK, so merged in the change back to it being an option instead of an argument.

One question, is strain < -100% for xx, yy, zz or isotropic inverting the relevant lattice vector(s) a bug or a feature? For shear strains it makes sense.

bhourahine:
Tests now in place. One thing that this script (and repeatgen) aren't testing is for bogus types of optional flags (this would require subclassing OptionParser and overriding its exit() and error() wouldn't it?).

PR #26, opened 2017-08-10
Socket cluster

bhourahine:
Bug fixes cases where the system is not periodic but the socket interface is being used.


prog/dftb+/prg_dftb/dftbplus.F90
```diff
@@ -451,23 +451,30 @@ program dftbplus
 
     if (tSocket) then
       call socket%receive(coord0, latvec)
```
aradi: What about introducing a kind of buffer array receiving the lattice vectors from the socket? We could transfer its content into `latVec` if the system is periodic. This would make the `else` branch and all changes in `initprogram.F90` superfluous, and additionally prevent that `latVec` contains garbage.

test/prog/dftb+/sockets/H2O_cluster/prerun.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def frac2cart(latvecs,coords):
+    newcoords = np.array(coords)
+    for iAt in range(len(coords)):
+        newcoords[iAt] = np.dot(np.transpose(latvecs),coords[iAt])
+    return (newcoords)
+
+
+def readgen(fname):
```
aradi: We have now `frac2cart`, `readgen` and `receive_all` and `a0` 3 times identically in various prerun scripts. What about creating a `socketutils` module (maybe in the directory above?) and symlink it from each socket test dir, so that it can be easily imported in the respective python scripts?

PR #32, opened 2017-09-08
Mainrefact

aradi:
Main DFTB+ file cleaned up and split into subroutines.


bhourahine:
periodic.F90
==========

Any reason why iPair in getSparseDescriptor isn't intent (out), as its always overwritten?

bhourahine:
main.F90 again
============

iAtomStart and iPair aren't particularly consistent names. How about iDnsAtomStrt, iSprsAtomStrt or similar? Eventually, probably should be both abstracted into an ADT and interfaces to get at matrix elements.

Any reason for several of the intent (inout) instead of (out) in handleCoordinateChange, as several of these are just reallocated, so won't have retained data on exit (similarly for  reallocateSparseArrays). Also, again coord and species could be sliced to give coord0 and species0.

Subroutines calcRepulsiveEnergy and calcDispersionEnergy seem overkill, as its a two line operation in each case (also suggests repulsives should be packaged up like dispersion eventually).

Should resetExternalPotentials, resetInternalPotentials and mergeExternalPotentials not be in the module that defines the potential data type (or even just generalize cases under the interface to add potentials together in the latter case)?

Similarly, should getChargePerShell be in the populations module?

addChargePotentials has  species and species0 as inputs. Also some of the update stuff to get atomic, shell and block to synchronise is similar to merge potential stuff above (add mergeInternalPotentials?).

For getSccHamiltonian, worth adding a flag to the potential data structure for the presence of imaginary potentials, allowing to trap for attempts to add imaginary potential parts to the hamitonian if they shouldn't be there (or equivalently that they should be added but iham itsn't allocated).

bhourahine:
main.F90
=======

Not wildly happy about the pointer aliasing in calculateLinRespExcitations (pNaturalOrbs2 and pNaturalOrbs3), as it increases the number of pointers and also looks a bit like a hack. Why not either change the addGradients to match the eigenvector write (will eventually need to if spin polarized forces happen), or pass it a slice of the 3 index array?

bhourahine:
main.F90

Any reason for fixed size arrays like  tLatOptFixLen and stress being sized as unknown on entry to subroutines? So  tLatOptFixLen(3) instead of tLatOptFixLen(:) in the interface, where the latter would  require an assert to test instead of at compile time.

prog/dftb+/lib_dftb/periodic.F90
```diff
@@ -713,19 +715,67 @@ subroutine reallocateArrays3(iNeighbor, neighDist2, mNewNeighbor)
...
+    integer, intent(in) :: nNeighbor(:)
+
+    !> Indexing for mapping image atoms to central cell
+    integer, intent(in) :: img2CentCell(:)
+
+    !> Atomic orbital information
+    type(TOrbitals), intent(in) :: orb
+
+    !> Sparse array indexing for the start of atomic blocks in data structures
+    integer, allocatable, intent(inout) :: iPair(:,:)
```
aradi: Must be intent(inout), as an intent(out) allocatable argument is automatically deallocated when entering the subroutine.

aradi:
As for the general comment: I did not document the helper subroutine arguments, as `runDftbPlus()` is the only routine, which is publicly exported by the module.

bhourahine:
Hi,

The added  public facing routines I was noticing weren't in main.F90 (periodic for example). Several of the internal ones in main probably will become public in the future, either when used in other contexts (linear response) or if migrated to existing modules they have something in common with.

I've commented them up anyway (want to set a good example to external developers...)

Cheers

Ben

On September 11, 2017 9:22:36 AM GMT+01:00, "Bálint Aradi" <notifications@github.com> wrote:
>As for the general comment: I did not document the helper subroutine
>arguments, as `runDftbPlus()` is the only routine, which is publicly
>exported by the module.
>
>-- 
>You are receiving this because your review was requested.
>Reply to this email directly or view it on GitHub:
>https://github.com/dftbplus/dftbplus/pull/32#issuecomment-328455333

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

prog/dftb+/prg_dftb/main.F90
```diff
@@ -0,0 +1,5637 @@
...
+    !> species of atoms
+    integer, intent(in) :: species(:)
+
+    !> atomic orbital information
+    type(TOrbitals), intent(in) :: orb
+
+    !> Index for atomic blocks in sparse data
+    integer, intent(in) :: iPair(:,:)
+
+
```
aradi: I agree, I have moved all IO related routines to`mainio`.

prog/dftb+/prg_dftb/main.F90
```diff
@@ -0,0 +1,5637 @@
...
+      call xlbomdIntegrator%setDefaultSCCParameters(minSCCiter, maxSccIter, sccTol)
+    end if
+
+    ! If the geometry is periodic, need to update lattice information in geometry loop
+    tLatticeChanged = tPeriodic
+
+    ! As first geometry iteration, require updates for coordinates in dependent routines
+    tCoordsChanged = .true.
+
+    ! Belongs in initprogram where initial geometry is set up
```
aradi: I moved it to initprogram, where it actually belongs to.

prog/dftb+/prg_dftb/main.F90
```diff
@@ -0,0 +1,5637 @@
...
+    logical :: tStopSCC, tStopDriver
+
+    !> Whether scc restart info should be written in current iteration
+    logical :: tWriteSccRestart
+
+    !> Whether charges should be written
+    logical :: tWriteCharges
+
+    !> net charge on each atom
+    real(dp), allocatable :: dqAtom(:)
```
aradi: Indeed superfluous, removed.

prog/dftb+/prg_dftb/main.F90
aradi: I've removed superfluous argument. As for passing in `nAtom` instead of array sizing: I wanted to reduce the number of arguments, so I decided to deduce `nAtom` from array sizes in all routines. This is not the most elegant, I agree. Once we have a proper geometry container with coords, species, etc. inside, it could also have a field `nAtom`, and we won't need this hack... :wink:

prog/dftb+/prg_dftb/main.F90
```diff
@@ -0,0 +1,5637 @@
...
+          & energy%EMermin, pressure, energy%EGibbs, coord0, tLocalise, localisation)
+    end if
+    if (tWriteResultsTag) then
+      call writeResultsTag(fdResultsTag, resultsTag, totalDeriv, chrgForces, tStress, totalStress,&
+          & pDynMatrix, tPeriodic, cellVol)
+    end if
+    if (tWriteDetailedXML) then
+      call writeDetailedXml(runId, speciesName, species0, pCoord0Out, tPeriodic, latVec, tRealHS,&
+          & nKPoint, nSpin, size(eigen, dim=1), nOrb, kPoint, kWeight, filling, occNatural)
+    end if
```
aradi: We could/should move this two writing routines into the geometry loop indeed. But as in this branch I did not want to change program behaviour, let's do it later in an other branch.

prog/dftb+/prg_dftb/main.F90
```diff
@@ -0,0 +1,5637 @@
...
+    real(dp) :: diffGeo
+
+    !> Loop variables
+    integer :: iSCCIter
+
+    !> File descriptor for the tagged writer
+    integer :: fdAutotest
+
+    !> File descriptor for the human readable output
+    integer :: fdUser
```
aradi: I've changed it to `fdDetailedOut` for consistency.

prog/dftb+/prg_dftb/main.F90
```diff
@@ -0,0 +1,5637 @@
...
+    real(dp), allocatable :: chrgForces(:,:)
+
+    !> excited state force addition
+    real(dp), allocatable :: excitedDerivs(:,:)
+
+    !> Stress tensors for various contribution in periodic calculations
+    real(dp) :: totalStress(3,3), totalLatDeriv(3,3)
+
+    !> derivative of cell volume wrt to lattice vectors, needed for pV term
+    real(dp) :: derivCellVol(3,3)
```
aradi: @Ben can you suggest a better name? I could come up with `derivPV` or `derivPresVol`, but both look somewhat ugly.

prog/dftb+/prg_dftb/main.F90
```diff
@@ -0,0 +1,5057 @@
...
+    !> species of atoms
+    integer, intent(in) :: species(:)
+
+    !> atomic orbital information
+    type(TOrbitals), intent(in) :: orb
+
+    !> Index for atomic blocks in sparse data
+    integer, intent(in) :: iPair(:,:)
+
+
```
aradi: Changed to intent(in).

aradi: Changed to intent(in)

aradi: I've tamed the monster by decomposing it.

bhourahine:
pdV ?

On September 11, 2017 1:08:41 PM GMT+01:00, "Bálint Aradi" <notifications@github.com> wrote:
>aradi commented on this pull request.
>
>
>
>> +    type(TPotentials) :: potential
>+
>+    !> Energy derivative with respect to atomic positions
>+    real(dp), allocatable :: totalDeriv(:,:)
>+
>+    !> Forces on any external charges
>+    real(dp), allocatable :: chrgForces(:,:)
>+
>+    !> excited state force addition
>+    real(dp), allocatable :: excitedDerivs(:,:)
>+
>+    !> Stress tensors for various contribution in periodic
>calculations
>+    real(dp) :: totalStress(3,3), totalLatDeriv(3,3)
>+
>+    !> derivative of cell volume wrt to lattice vectors, needed for pV
>term
>+    real(dp) :: derivCellVol(3,3)
>
>@Ben can you suggest a better name? I could come up with `derivPV` or
>`derivPresVol`, but both look somewhat ugly.
>
>-- 
>You are receiving this because your review was requested.
>Reply to this email directly or view it on GitHub:
>https://github.com/dftbplus/dftbplus/pull/32#pullrequestreview-61796156

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

prog/dftb+/prg_dftb/main.F90
```diff
@@ -0,0 +1,5057 @@
...
+          call getNextInputCharges(pChrgMixer, qOutput, qOutRed, orb, nIneqOrb, iEqOrbitals,&
+              & iGeoStep, iSccIter, minSccIter, maxSccIter, sccTol, tStopScc, tDftbU, tReadChrg,&
+              & qInput, qInpRed, sccErrorQ, tConverged, qBlockOut, iEqBlockDftbU, qBlockIn,&
+              & qiBlockOut, iEqBlockDftbULS, species0, nUJ, iUJ, niUJ, qiBlockIn)
+          call getSccInfo(iSccIter, energy%Eelec, Eold, diffElec)
+          call printSccInfo(tDftbU, iSccIter, energy%Eelec, diffElec, sccErrorQ)
+          tWriteSccRestart = needsSccRestartWriting(restartFreq, iGeoStep, iSccIter, minSccIter,&
+              & maxSccIter, tMd, tGeoOpt, tDerivs, tConverged, tReadChrg, tStopScc)
+          if (tWriteSccRestart) then
+            call writeRestart(fChargeIn, orb, qInput, qBlockIn, qiBlockIn)
```
aradi: Added explicit file descriptor passing (we should at some point dispose this in favour of `newunit` in the  `open` statement.

prog/dftb+/prg_dftb/mainio.F90
```diff
@@ -35,26 +39,54 @@ module mainio
...
 
+  !> Format string for energy second derivative matrix
   character(len=*), parameter :: formatHessian = '(4f16.10)'
+
+  !> Format for energy values, typically band energies
   character(len=*), parameter :: formatEnergy = '(8f12.5)'
+
+  !> Format for eigenvalue energies
   character(len=*), parameter :: formatEigen = '(8f14.8)'
+
```
aradi: Renamed and moved into the subroutine where it is used

aradi:
A few things I've left untouched:

- Still passing file name to `writeAutotestTag()` and similar routines, since we have to reopen those files, therefore, file name is needed again and again.
- Some variables I can't convert to `intent(in)`, despite that they are only queried but not modified, as every query changes they internal status (e.g. our linked lists change their internal cache when queried...)
- I'd suggest to keep the allocation status of an array as flag, as then it never can happen, that the routine is called with `tHaveSomeFeature = .true.` but the corresponding feature array is not allocated. In the case of stress, we do not allocate the array but use a 3x3 static array, so that won't work. Maybe we should make an allocatable out of that as well...
- `calcRepulsivEnergy` and `calcDispersionEnergy` are an overkill indeed, however, I thought they would improve code readability and abstraction level consistency in `runDftbPlus()`.
- I left all dummy array arguments assumed shape. If we declare them fixed size, the compiler would pass it as an F77 array without size information and create a fixed size descriptor in the routine. So, whichever array size you pass when calling, an `@:ASSERT(size(array) == someValue)` within the subroutine will always pass.

aradi:
I also left the `reset/merge*Potentials` routines in `main.F90` for the moment, as they are quite strongly coupled with it (e.g. `tDualSpinOrbit` and `xi` are needed).

aradi:
Let me know, if I ignored something or there are other issues you find. :smile:

PR #36, opened 2017-09-22
Moved dispersion and Erep after SCC loop

bhourahine:
Not urgent, but might be worth merging in relatively soon.

### DIFF
--- a/prog/dftb+/prg_dftb/main.F90
+++ b/prog/dftb+/prg_dftb/main.F90
@@ -328,13 +328,13 @@ contains
       end if
 
       if (tLatticeChanged) then
-        call handleLatticeChange(latVec, sccCalc, tStress, extPressure, mCutoff,&
-            & dispersion, recVec, invLatVec, cellVol, recCellVol, extLatDerivs, cellVec, rCellVec)
+        call handleLatticeChange(latVec, sccCalc, tStress, extPressure, mCutoff, recVec, invLatVec,&
+            & cellVol, recCellVol, extLatDerivs, cellVec, rCellVec)
       end if
 
       if (tCoordsChanged) then
         call handleCoordinateChange(coord0, latVec, invLatVec, species0, mCutoff, skRepCutoff, orb,&
-            & tPeriodic, sccCalc, dispersion, thirdOrd, img2CentCell, iCellVec,&
+            & tPeriodic, sccCalc, thirdOrd, img2CentCell, iCellVec,&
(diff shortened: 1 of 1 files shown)


prog/dftb+/prg_dftb/main.F90
```diff
@@ -1329,17 +1323,72 @@ subroutine handleCoordinateChange(coord0, latVec, invLatVec, species0, mCutoff,
...
+    
+    ! Notify various relevant modules about current coordinates/boundary conditions
+
+    call calcRepulsiveEnergy(coord, species, img2CentCell, nNeighbor, neighborList, pRepCont,&
+        & energy%atomRep, energy%ERep)
+    
+    if (allocated(dispersion)) then
+      if (tLatticeChanged) then
+        call dispersion%updateLatVecs(latVecs)
+        mCutoff = max(mCutoff, dispersion%getRCutoff())
```
aradi: I think, that won't work robustly enough. The neighbour list for the current geometry has been already created at that point. If (although very unlikely) the dispersion needed a longer real space cutoff as all other modules, it would not contain enough neighbors for the following `updateCoords()` call. We should probaly make the `upateLatVecs()` still at the beginning of the geoopt-loop.

prog/dftb+/prg_dftb/main.F90
```diff
@@ -424,10 +418,10 @@ subroutine runDftbPlus()
...
         end if
 
-        call getEnergies(sccCalc, qOutput, q0, chargePerShell, species, tEField, tXlbomd,&
-            & tDftbU, tDualSpinOrbit, rhoPrim, H0, orb, neighborList, nNeighbor, img2CentCell,&
-            & iSparseStart, cellVol, extPressure, TS, potential, energy, thirdOrd, qBlockOut,&
-            & qiBlockOut, nDftbUFunc, UJ, nUJ, iUJ, niUJ, xi)
+!        call getEnergies(sccCalc, qOutput, q0, chargePerShell, species, tEField, tXlbomd,&
+!            & tDftbU, tDualSpinOrbit, rhoPrim, H0, orb, neighborList, nNeighbor, img2CentCell,&
+!            & iSparseStart, cellVol, extPressure, TS, potential, energy, thirdOrd, qBlockOut,&
+!            & qiBlockOut, nDftbUFunc, UJ, nUJ, iUJ, niUJ, xi)
```
aradi: That would unfortunately screw up the output completely, as the user would see only zeros for the energies during the scc-cycle. We would have probably to separate the energy calculation into an inner- and an outer-scc part.

PR #54, opened 2017-12-07
Fix charge names in detailed.out

aradi:


aradi:
The wrong submodule status came from not having to executed `git submodule update --recursive` when changing branch. I fixed it now, and rebased it on current master.

bhourahine:
Yes, submodules seem to be a pain when switching branches. Changes approved.

PR #55, opened 2017-12-07
Random init

bhourahine:
Branch to initialize charge and spin with random values if required


doc/dftb+/manual/dftbp.tex
```diff
@@ -1186,9 +1185,16 @@ \section{Hamiltonian}
...
+Net charges on atoms not appearing in any \is{AtomCharge}
+specification are set to be zero.
 
+Finally, the charges on all atoms can be set using the \kw{Random} keyword:
+\begin{verbatim}
+  InitialCharges = { # Specifying charge for various species
+      Random = Yes
+  }
+\end{verbatim}
+These will then be supplied by the code's internal random number generator.
```
aradi: I think, that does not fit to the logic. For me
```
InitialCharges {
  RandomCharges {}
}
```
would be more natural. Also giving us the chance to parametrize it further (e.g. specify the range of the atoms, the range of the randomness).

doc/dftb+/manual/dftbp.tex
```diff
@@ -1461,6 +1467,19 @@ \subsubsection{Colinear\{\}}
...
+\invparskip
+\begin{verbatim}
+  SpinPolarisation = Colinear {
+    UnpairedElectrons = 0.0
+    InitialSpins = {
+      Random = Yes
+    }
+  }
+\end{verbatim}
+
```
aradi: The same as above for `InitialCharges`.

prog/dftb+/lib_math/ranlux.F90
```diff
@@ -268,6 +269,23 @@ subroutine Ranlux_getRandom2DArray(self, r2Darray)
...
+
+    !> Array containing the random numbers on exit.
+    real(dp), intent(out) :: r3Darray(:,:,:)
+
+    real(dp), allocatable :: rvec(:)
+
+    allocate(rvec(product(shape(r3Darray))))
+    call getRandomVector_local(rvec, self%iseeds, self%icarry, self%in24, &
+        &self%i24, self%j24, self%next, self%nskip, self%twom24, self%twom12)
+    r3Darray = reshape(rvec,shape(r3Darray))
```
aradi: `r3dArray(:,:,:) = ...` Although there is no possibility for automatic reallocation on the LHS as `r3dArray` is not allocatable, we should always indicate the array indices on the LHS if no reallocation is desired.

prog/dftb+/prg_dftb/initprogram.F90
```diff
@@ -852,7 +852,7 @@ subroutine initProgramVariables(input, env)
     type(OVelocityVerlet), allocatable :: pVelocityVerlet
     type(OTempProfile), pointer :: pTempProfile
 
-    type(ORanlux), allocatable :: randomInit, randomThermostat
+    type(ORanlux), allocatable :: randomInit, randomThermostat, randomCharges, randomSpins
```
aradi: I am somewhat undecided, whether we really need two more additional random generators, or whether this could be handled directly by `ṙandomInit` or by one common new random generator for both cases.

prog/dftb+/prg_dftb/initprogram.F90
```diff
@@ -1943,36 +1947,55 @@ subroutine initProgramVariables(input, env)
...
+              do ii = 1, nAtom
+                do jj = 1, 3
+                  qInput(1:orb%nOrbAtom(ii),ii,jj+1) = qInput(1:orb%nOrbAtom(ii),ii,1) *&
+                      & input%ctrl%initialSpins(jj,ii) / sum(qInput(1:orb%nOrbAtom(ii),ii,1))
+                end do
+              end do
+            end if
           end if
         end select
+
```
aradi: At some point, we will have to package initprogram also in subroutines, as it became a real mess. Maybe it would worth to package the charge/spin initialisation as a separate routine?

prog/dftb+/prg_dftb/initprogram.F90
```diff
@@ -2680,16 +2702,33 @@ subroutine createRandomGenerators(env, seed, randomInit, randomThermostat)
...
+    initRandom0 = (seed ==0)
 
     call init(randGenPool, env, seed, oldCompat=.true.)
 
+    if (initRandom0) then
+      write(stdOut, "(A,':',T30,I14)") "Chosen random seed:", seed
+    else
+      write(stdOut, "(A,':',T30,I14)") "Specified random seed:", seed
+    end if
+
```
aradi: Yes! That we definitely always need to print out.

prog/dftb+/prg_dftb/initprogram.F90
bhourahine: Probably not needed. Depends on whether there is ever anything which non-deterministically pulls out random values from the generator but at the same time needs a repeatable random sequence for something else.
About the only thing I can think of that might do this isn't likely to be implemented (coupling from the past in MC).

bhourahine:
Converted to issue #353

PR #58, opened 2017-12-09
Fe2 spin orbit test too small for > 2 processors

bhourahine:


aradi:
I can run that test on 4 processors without any problems with most compilers and ScaLAPACK implementations. The only combination which hangs is intel17 with corresponding MKL. So, probably, we should investigate it more in detail, it maybe a compiler/MKL problem.

aradi:
Looking more in detail: I can get various examples frozen with my local openmpi/intel combination using 4 processes, some of them even before any MKL-routines are invoked. (I can even get spinorbit/Fe2 frozen with 2 processes, if I keep starting it again after it has finished)...

aradi:
What do you think? Should I close this? It is probably more an issue as a PR, as we don't know, why sometimes that test can be run without problems and why sometimes it hangs.

bhourahine:
Agreed, this should now move over to become an issue (ifort17/MKL 4 processors).

PR #60, opened 2017-12-12
Lbfgs

bhourahine:
Limited memory BFGS routines contributed by Frank Stuckenberg


prog/dftb+/lib_geoopt/lbfgs.F90
```diff
@@ -0,0 +1,886 @@
...
+    real(dp), intent(in) :: x0(:)
+
+    !> New direction
+    real(dp), intent(in) :: d0(:)
+
+    real(dp), intent(in) :: firstStep
+
+    @:ASSERT(size(x0) == this%nElem)
+    @:ASSERT(size(d0) == this%nElem)
+    #!@:ASSERT(all(abs(d0) > 0.0_dp))
```
aradi: I had to disable this ASSERT in order to get the tests working. Shouldn't this be `any()` instead of all?

PR #61, opened 2017-12-12
Tolerance for Ewald summation set by parser

bhourahine:


prog/dftb+/lib_dftb/scc.F90
```diff
@@ -75,6 +71,10 @@ module scc
 
     !> if > 0 -> manual setting for alpha
     real(dp) :: ewaldAlpha = 0.0_dp
+
+    !> Ewald tollerance
+    real(dp) :: tolEwald = 0.0_dp
```
aradi: Let's set here a reasonable default (1e-9?).

bhourahine: Agreed, will change

PR #63, opened 2017-12-16
Serial complex SVD routines added

bhourahine:
Added wrappers for SVD decomposition of complex matrices using LAPACK. Will be required by future branches.


bhourahine:
Changes made and merged

PR #64, opened 2017-12-16
Exc fix

bhourahine:
Fix for forces being always calculated. Improved excited state openMP parallelism and posibility for Mulliken charges and similar for multiple excited states (not forces).

### DIFF
--- a/prog/dftb+/lib_timedep/linrespcommon.F90
+++ b/prog/dftb+/lib_timedep/linrespcommon.F90
@@ -103,17 +103,21 @@ contains
     eExc = wij(1)
     iStart = 1
     nxov_r = 0
+    ! Check, to single precision tolerance, for degeneracies when selecting bright transitions
     do ii = 2, nxov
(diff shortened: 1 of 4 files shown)


prog/dftb+/lib_timedep/linrespcommon.F90
```diff
@@ -172,11 +177,13 @@ subroutine getNorb_r(nxov, win, getij, homo, no, nv)
     no = 0
     nv = 0
 
+    !$OMP PARALLEL DO DEFAULT(SHARED) PRIVATE(ia,ii,jj) SCHEDULE(RUNTIME) REDUCTION(+:nv,no)
```
aradi: General remark: The iteration variable for which the `$OMP PARALLEL DO` is defined is private by default, so it does not need to be listed.
Side note: I am wondering, whether we should switch to use `DEFAULT(NONE)` at some point to reduce chances of having a shared variable due to forgetting of list it as private...

PR #65, opened 2017-12-18
Replaced parallel DM/EDM routines with pGEMM

bhourahine:
Improved parallel scaling for density and energy-density matrices.


bhourahine:
Timings show this isn't worth incorporating.

PR #67, opened 2017-12-20
Release name

aradi:
Automatic git tag based release name determination during build (but can be disabled if needed).

Note: in order to reliably determine last release, the tagged release must be reachable from the current development version, therefore the release branch has been merged into this branch. This does not introduce any code changes, only includes the corresponding merge commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ make.arch
 make.config
 __pycache__
 .idea
+RELEASE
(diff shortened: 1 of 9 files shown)


bhourahine:
All fine to merge. But on a user experience note, the appended
release += '* [commit %s]' % commit
message in update_release could possibly be used to make it clearer this is a version after the tagged release number (i.e. an unofficial).

aradi:
Any suggestions for saying it in a compact way? What about `Release 17.1* [unofficial release, commit: abcdefg]`?

PR #69, opened 2017-12-22
Real SVD added

bhourahine:
Also changed ASSERT choices to allow for larger matrices passed into SVD
routines.

### DIFF
--- a/prog/dftb+/lib_math/lapackroutines.F90
+++ b/prog/dftb+/lib_math/lapackroutines.F90
@@ -1183,19 +1183,13 @@ contains
     lda = size(A,dim=1)
     ldu = size(U,dim=1)
     ldvt = size(Vt,dim=1)
-    @:ASSERT(all(shape(u) >= (/m,mn/)))
-    @:ASSERT(all(shape(vt) >= (/mn,n/)))
+    @:ASSERT(all(shape(u) == (/m,mn/)))
+    @:ASSERT(all(shape(vt) == (/mn,n/)))
     @:ASSERT(size(sigma) == mn)
 
-   lwork = max(1,3*min(m,n)+max(m,n),5*min(m,n))
+    lwork = max(1,3*min(m,n)+max(m,n),5*min(m,n))
 
     allocate(work(lwork))
-
-    u = 0.0_rsp
-    vt = 0.0_rsp
-    sigma = 0.0_rsp
-    work = 0.0_rsp
-    info = 0
 
     ! get only the minimum(m,n) singular vectors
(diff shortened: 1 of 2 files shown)


prog/dftb+/lib_math/lapackroutines.F90
```diff
@@ -1155,6 +1157,111 @@ subroutine larnv_dblecplx(iDist,iSeed,x)
...
+   lwork = max(1,3*min(m,n)+max(m,n),5*min(m,n))
+
+    allocate(work(lwork))
+
+    u = 0.0_rsp
+    vt = 0.0_rsp
+    sigma = 0.0_rsp
+    work = 0.0_rsp
+    info = 0
+
```
aradi: Any reason to fill up those arrays with zeros? They are all intent(out) arguments for the following `sgesvd` call

prog/dftb+/lib_math/lapackroutines.F90
```diff
@@ -1155,6 +1157,111 @@ subroutine larnv_dblecplx(iDist,iSeed,x)
...
+    m = size(A,dim=1)
+    n = size(A,dim=2)
+    mn = min(m,n)
+    lda = size(A,dim=1)
+    ldu = size(U,dim=1)
+    ldvt = size(Vt,dim=1)
+    @:ASSERT(all(shape(u) >= (/m,mn/)))
+    @:ASSERT(all(shape(vt) >= (/mn,n/)))
+    @:ASSERT(size(sigma) == mn)
```
aradi: Why `>=` instead of `==`? According to LAPACK no elements beyond `[m, mn]` or `[mn, m]` are set within the call, so it would be logical to ask for exactly that array size.

bhourahine: Not if you have jobu/jobvt = 'A' instead of 'S'.

prog/dftb+/lib_math/lapackroutines.F90
bhourahine: If unreferenced, would they be 0 for all compilers? If that is always safe, then it can be removed.

aradi: No, if unreferenced their content would be arbitrary. But if they have the right size (see above) the F77-LAPACK routine should set all of them to some value, shouldn't it?

bhourahine: Both u and vt might have unreferenced elements inside LAPACK if respective jobs are not 'A'. Sigma and work are harmless though, so will check in a change there.

bhourahine:
Just realised your comments about asserts and initialisations are correct since this wraps the 'S' case.

PR #81, opened 2018-01-11
H0 sbuild parallel

bhourahine:
MPI parallelised H0 and S build and added extra timers


prog/dftb+/lib_dftb/nonscc.F90
```diff
@@ -303,6 +322,9 @@ subroutine buildDiatomicBlocks(out, skCont, coords, nNeighbors, &
     end do
     !$OMP  END PARALLEL DO
 
+    ! Note
+    ! Do not assemble chunks back together, as its done in routine calling this one
+
```
aradi: Routine should rather receive the boundaries as dummy argument. That is more explicit and causes less confusion. (I've changed it accordingly).

PR #84, opened 2018-01-18
Get slakos

aradi:
Slakos downloaded from github. Script `get_opt_externals` more verbose and robust.

### DIFF
--- a/make.config
+++ b/make.config
@@ -21,11 +21,11 @@ WITH_SOCKETS := 1
 # Whether the DFTD3 library (dispersion) should be linked.
 # Note: Due to license incompatibilities, distribution of a binary compiled with
 # this component is not permitted. You can use it only for your personal research.
-WITH_DFTD3 := 1
+WITH_DFTD3 := 0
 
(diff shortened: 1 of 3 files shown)


make.config
```diff
@@ -19,6 +19,8 @@ WITH_MPI := 0
 WITH_SOCKETS := 1
 
 # Whether the DFTD3 library (dispersion) should be linked.
+# Note: Due to license incompatibilities, distribution of a binary compiled with
```
bhourahine: Would having a Y/N user prompt at this stage be too annoying for people?

make.config
```diff
@@ -19,6 +19,8 @@ WITH_MPI := 0
...
+# this component is not permitted. You can use it only for your personal research.
 WITH_DFTD3 := 1
```
bhourahine: Agree that 'as shipped' this should be = 0

aradi:
I've tried to make the choices leading to a build with license incompatible components more prominent. By default, no such component would be involved any more.

PR #87, opened 2018-01-22
Printing random seed always happens when initialised

bhourahine:
Instead of only in an MD calculation. Also indicates whether the
value is chosen (seed 0) or specified by the user.

### DIFF
--- a/prog/dftb+/prg_dftb/initprogram.F90
+++ b/prog/dftb+/prg_dftb/initprogram.F90
@@ -2390,7 +2390,6 @@ contains
           & then
         write(stdOut, "(A,':',T30,E14.6)") "Temperature", tempAtom
       end if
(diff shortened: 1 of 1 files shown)


prog/dftb+/prg_dftb/initprogram.F90
```diff
@@ -2700,9 +2699,18 @@ subroutine createRandomGenerators(env, seed, randomInit, randomThermostat)
...
+    initRandom0 = (seed ==0)
 
     call init(randGenPool, env, seed, oldCompat=.true.)
 
+    if (initRandom0) then
+      write(stdOut, "(A,':',T30,I14)") "Chosen random seed", seed
+    else
+      write(stdOut, "(A,':',T30,I14)") "Specified random seed", seed
+    end if
+
```
aradi: I have moved them were all other print statements are grouped.

PR #100, opened 2018-02-11
Serialise iterative operations due to the PGI drift behavior

bhourahine:
Branch to calculate iterative results on the master node and broadcast.


aradi:
We should discuss about this and decide about a general synchronisation strategy...

bhourahine:
Converted to issue #352 to be addressed more generally

PR #102, opened 2018-02-12
Use dynamic neighbor lists for Ewald-summation

aradi:


aradi:
Just realised, that rthe esulting binary crashes, if compiled with PGI (all other compilers seem to be fine). So, do not merge yet.

aradi:
With latest fixes it seems to work fine.

aradi:
I force overwrote this branch. It should be now compatible with master.

prog/dftb+/lib_dftb/coulomb.F90
```diff
@@ -56,6 +57,10 @@ module coulomb
   !> Maximal argument value of erf, after which it is constant
   real(dp), parameter :: erfArgLimit = 10.0_dp
 
+  !> Chunk size to use when obtaining neighbors dynamically via an iterator
+  integer, parameter :: iterChunkSize = 1000
```
bhourahine: Any reason for that particular value? Not a change request, but could this be tuned for performance?

prog/dftb+/lib_dftb/coulomb.F90
```diff
@@ -351,25 +346,33 @@ subroutine invRPeriodicBlacs(grid, coord, nNeighborEwald, iNeighbor, img2CentCel
...
+    integer :: neighImages(iterChunkSize)
+    integer :: iAt1, iAt2, iAt2f, iNeigh, nNeigh, jj, ii, iLoc, jLoc
     logical :: tLocal
     
     @:ASSERT(volume > 0.0_dp)
 
     invRMat(:,:) = 0.0_dp
+    pNeighList => neighList
 
     ! Real space part of the Ewald sum.
```
bhourahine: Should probably add an OpenMP loop

prog/dftb+/lib_type/dynneighlist.F90
```diff
@@ -0,0 +1,244 @@
...
+  implicit none
+  private
+
+  public :: TDynNeighList, TDynNeighList_init
+  public :: TNeighIterator, TNeighIterator_init
+
+
+  !> Dynamic neighbor list
+  type :: TDynNeighList
+    private
```
bhourahine: Even though these are private, for future new developers, would be nice if the variables were commented.

prog/dftb+/lib_type/latpointiter.F90
```diff
@@ -0,0 +1,235 @@
...
+  use assert
+  implicit none
+  private
+
+  public :: TLatPointIter, TLatPointIter_init
+
+
+  !> Lattice point iterator
+  type :: TLatPointIter
+    private
```
bhourahine: Same comment about comments on variables

prog/dftb+/lib_type/latpointiter.F90
```diff
@@ -0,0 +1,235 @@
...
+      if (tFinished) then
+        exit
+      end if
+    end do
+    latticePoints = tmpLatPoints(:, 1:iLatPoint - 1)
+
+  end subroutine TLatPointIter_getAllPoints
+
+
+  !> Helper function to increase a tuple of 3 indices by one
```
bhourahine: Is this not cycling through permitted values of the tuple within the ranges, instead of just increasing?

prog/dftb+/lib_dftb/coulomb.F90
aradi: Is not really a conscious choice. Just something I thought would avoid repeated calls for the iterator for small and middle sized non-periodic systems and not to waste too much memory...
We could make it tunable by the user, for sure, though all related automatic allocations would have to be turned probably into explicit allocations.

prog/dftb+/lib_type/dynneighlist.F90
aradi: Absolutely correct, I did that in both files now.

prog/dftb+/lib_type/latpointiter.F90
aradi: Yes, it does, sorry for the misleading naming. I've changed it to something (hopefully) more self-explaining.

aradi:
Yes, we could use it to OMP-parallelise the generation of the static neighbor list. Some care must be taken, though, if reallocation is needed due to increasing number of neighbors. That's definitely worth of a PR after 18.1 release.

As for spelling, I am using a somewhat following strategy: For variables/function names etc. I always use AE as intrinsic Fortran names/keywords are also in AE. So reading the code, everything is consistent.
In comments, though, I tend to BE, although also not being always very consistent. (We should probably discuss it and add it to the develguide, whatever we decide on)

PR #105, opened 2018-02-18
Formatting change for eigenvalues/fillings

bhourahine:
Both detailed.out and band.dat with an index for the state, and
detailed.out reordered to put fillings next to eigenvalues.

### DIFF
--- a/prog/dftb+/prg_dftb/mainio.F90
+++ b/prog/dftb+/prg_dftb/mainio.F90
@@ -2193,7 +2193,8 @@ contains
       do iK = 1, size(eigen, dim=2)
         write(fd, *) 'KPT ', iK, ' SPIN ', iSpin, ' KWEIGHT ', kWeight(iK)
         do iEgy = 1, size(eigen, dim=1)
-          write(fd, "(2f12.5)") Hartree__eV * eigen(iEgy, iK, iSpin), filling(iEgy, iK, iSpin)
+          write(fd, "(I6,2f12.5)") iEgy, Hartree__eV * eigen(iEgy, iK, iSpin),&
+              & filling(iEgy, iK, iSpin)
         end do
         write(fd,*)
       end do
@@ -2373,8 +2374,7 @@ contains
     integer :: iAt, iSpin, iEgy, iK, iSp, iSh, iOrb, kk
     logical :: tSpin
 
-    character(*), parameter :: formatEigen = "(F14.8)"
-    character(*), parameter :: formatFilling = "(F12.5)"
+    character(*), parameter :: formatEigen = "(F14.8, F9.5)"
     character(lc) :: strTmp
 
     nAtom = size(q0, dim=2)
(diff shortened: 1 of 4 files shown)


prog/dftb+/prg_dftb/mainio.F90
```diff
@@ -2193,7 +2193,8 @@ subroutine writeBandOut(fd, fileName, eigen, filling, kWeight)
       do iK = 1, size(eigen, dim=2)
         write(fd, *) 'KPT ', iK, ' SPIN ', iSpin, ' KWEIGHT ', kWeight(iK)
         do iEgy = 1, size(eigen, dim=1)
-          write(fd, "(2f12.5)") Hartree__eV * eigen(iEgy, iK, iSpin), filling(iEgy, iK, iSpin)
+          write(fd, "(I6,2f12.5)") iEgy, Hartree__eV * eigen(iEgy, iK, iSpin),&
+              & filling(iEgy, iK, iSpin)
```
aradi: Note: the tool `dp_bands` must be adapted to be able to cope with this.

prog/dftb+/prg_dftb/mainio.F90
```diff
@@ -2475,19 +2475,50 @@ subroutine writeDetailedOut1(fd, fileName, tAppendDetailedOut, iDistribFn, nGeoS
...
+            if (iK + kk > nKPoint) then
+              exit
+            end if
+            write(fd, formatEigen, advance = 'no') Hartree__eV * eigen(iEgy, iK+kk, iSpin),&
+                & filling(iEgy, iK+kk, iSpin)
+          end do
+          write(fd, formatEigen)
+        end do
+        write(fd, *)
       end do
```
aradi: I know, that the write routines are not granular enough yet in general, but let's package the eigenvalue printing into a separate private routine, which is then called by `writeDetailedOut1`. Latter is already much beyond the optimal 50-100 statement lines per routine.

prog/dftb+/prg_dftb/mainio.F90
```diff
@@ -2373,8 +2374,7 @@ subroutine writeDetailedOut1(fd, fileName, tAppendDetailedOut, iDistribFn, nGeoS
     integer :: iAt, iSpin, iEgy, iK, iSp, iSh, iOrb, kk
     logical :: tSpin
 
-    character(*), parameter :: formatEigen = "(F14.8)"
-    character(*), parameter :: formatFilling = "(F12.5)"
+    character(*), parameter :: formatEigen = "(F14.8, F9.5)"
```
aradi: What about constraining the number of digits to yield meV precision only? I find too many digits (beyond DFTB precision) disturbing when analysing results.

prog/dftb+/prg_dftb/mainio.F90
aradi: It contradicts our style guide quite a few times on the usage of spaces 😉 (around operators, after comma, but not for optional arguments as in `advance=no`.

bhourahine: OK, will push something.

prog/dftb+/prg_dftb/mainio.F90
bhourahine: Will do.
But on the other hand, for finite difference derivatives, 3 decimal places is not enough. So in the longer term, a user flag to set the precision level might be useful.

prog/dftb+/prg_dftb/mainio.F90
bhourahine: OK, fixed.

prog/dftb+/prg_dftb/mainio.F90
bhourahine: Probably should make this backward compatible to the old format. Can you have a look at the regex in  dptools/bandout.py for this one?

aradi:
@bhourahine I have pushed a fix (parallel compilation did not work), fixed `dp_bands` and `dp_dos` and constrained the output to 80 characters in `detailed.out`. Pls. have a quick look, if you don't find anything else, I am fine with it to be merged.

PR #121, opened 2018-03-08
Corrected specification of PipekMezey input in manual

timostrunk:
The specification of the PipekMezey input was incorrect in the manual

    PipekMezey = Yes
leads to an error
 
    -> Node contains superfluous free text: 'Yes'

The correct format is:

    PipekMezey = {
    }

for defaults or

    PipekMezey = {
        Tollerance = 1.0E-4
        MaxIterations = 100
    }

This is changed in the Pull Request.  The spelling error in Tollerance is on purpose, because it's also written like this in the dftb+ executable.


bhourahine:
Many thanks for the PR. We'll look into correcting 'tolerances' the next time the parser version bumps up.

PR #125, opened 2018-03-19
H5 correction

Honza-R:
Implementation of the H5 correction for hydrogen bonds, and the associated small modification of the D3 dispersion allowing complete DFTB3-D3H5 calculations. Comes with tests and description of the input in the manual.


doc/dftb+/manual/dftbp.tex
```diff
@@ -2471,6 +2484,62 @@ \subsection{DFTB3}
 
 \end{verbatim}
 
+\subsection{DFTB3-D3H5}
+\label{sec:dftbp.D3H5}
+
+DFTB3-D3H5 \cite{rezac-jctc-13-2017} is a variant of DFTB3 with corrections for
+non-covalent interactions (dispersion an hydrogen bonds).
```
bhourahine: and hydrogen bonds

prog/dftb+/lib_dftb/dispdftd3.F90
```diff
@@ -261,4 +278,50 @@ function getRCutoff(this) result(cutoff)
...
+    !> Instance of D3
+    class(DispDftD3), intent(inout) :: this
+    !> Current coordinates
+    real(dp), intent(in) :: coords(:,:)
+
+    integer :: i, j, c
+    real(dp) :: r, repE, dEdR, dCdR
+
+    ! Parameters (with conversion to a.u.)
+    real(dp) :: k = 0.30_dp * 0.001593601371772_dp
```
bhourahine: Should use the constants.F90 parameters where possible, and add any missing conversion factors there.

prog/dftb+/lib_dftb/scc.F90
```diff
@@ -89,7 +96,7 @@ module scc
     !> Nr. of atoms
     integer :: nAtom
 
-    !> Nr. of species
+    !> Nr. of rpecies
```
bhourahine: Typo?

prog/dftb+/lib_dftb/h5correction.F90
```diff
@@ -0,0 +1,243 @@
...
+  end subroutine H5Corr_init
+
+  !> Print all the parameters for debugging
+  subroutine printH5Setup(this)
+    ! Arguments
+    class(H5Corr), intent(in) :: this
+    ! Local variables
+    integer :: iSp1
+
+     write(37,*) "H5 setup:"
```
bhourahine: Specific number units may cause problems, please use the newunit feature of Fortran 2008 (or set an id in the same way the main code opens files).

prog/dftb+/prg_dftb/parser.F90
```diff
@@ -2939,6 +2976,9 @@ subroutine readDispDFTD3(node, input)
         & modifier=buffer, child=child)
     call convertByMul(char(buffer), lengthUnits, child, input%cutoffCN)
     call getChildValue(node, "threebody", input%threebody, default=.false.)
+    ! D3H5 - additional H-H repulsion
+    call getChildValue(node, "hhrepulsion", input%hhrepulsion, default=.false.)
```
bhourahine: Does it make sense to set this without the H5 correction being also enabled?

prog/dftb+/prg_dftb/parser.F90
```diff
@@ -1421,6 +1421,43 @@ subroutine readDFTBHam(node, ctrl, geo, slako)
...
+        ! Switch the correction on
+        ctrl%h5SwitchedOn = .true.
+
+        ! H5 should not be used with X-H damping
+        if (ctrl%tDampH .and. ctrl%h5SwitchedOn) then
+          call error("H5 correction is not compatible with X-H damping")
+        end if
+
+        ! Read global parameters
+        ! Defaults are -1.0 to identify thet the parametrers were not read
```
bhourahine: that the parameters

prog/dftb+/lib_dftb/scc.F90
```diff
@@ -418,6 +431,16 @@ subroutine Scc_initialize(this, env, inp)
     this%tDampedShort(:) = inp%tDampedShort(:)
     this%dampExp = inp%dampExp
 
+    ! H5 correction
+    ! Input data are saved in the SCC object
+    this%useH5 = inp%useH5
```
bhourahine: Could instead use allocated(inp%h5Correction) to signal this.

prog/dftb+/lib_dftb/h5correction.F90
```diff
@@ -0,0 +1,243 @@
...
+    ! Get parameters for current pair of species
+    call this%getParams(iSp1, iSp2, applyCorrection, h5Scaling, sumVDW)
+
+    ! If applicable to the current pair, modify the gamma
+    if (applyCorrection) then
+      ! Gaussian calculation
+      fwhm = this%wScale * sumVDW
+      r0 = this%rScale * sumVDW
+      c = fwhm / 2.35482_dp
+      gauss = exp(-1.0_dp * ((rab*0.5291772083_dp)-r0)**2 / 2.0_dp / c**2) * h5Scaling
```
bhourahine: Is this internally in atomic or other unit system? Also constants from the module to be consistent with the rest of the code.

prog/dftb+/lib_dftb/h5correction.F90
```diff
@@ -0,0 +1,243 @@
...
+      ! Derivative calculation
+      dgauss = -1.0 * (0.5291772083 * ( 0.5291772083 * rab - r0 ) ) / c**2 * gauss
```
bhourahine: Values from parameters in constants.F90

prog/dftb+/lib_dftb/h5correction.F90
```diff
@@ -0,0 +1,243 @@
...
+
+    ! Get parameters for current pair of species
+    call this%getParams(iSp1, iSp2, applyCorrection, h5Scaling, sumVDW)
+
+    ! If applicable to the current pair, modify the gamma
+    if (applyCorrection) then
+      ! Gaussian calculation
+      fwhm = this%wScale * sumVDW
+      r0 = this%rScale * sumVDW
+      c = fwhm / 2.35482_dp
```
bhourahine: Where is this value coming from, also can fwhm be in atomic units internally.

prog/dftb+/lib_dftb/h5correction.F90
```diff
@@ -0,0 +1,243 @@
...
+      iSpHeavy = iSp1
+      spNameHeavy = spName1
+    else if (spName2 .ne. "H" .and. spName1 == "H") then
+      iSpHeavy = iSp2
+      spNameHeavy = spName2
+    else
+      return
+    end if
+   
+    ! For each species the correction is applied to,
```
bhourahine: What is the fall back for other species outside this set?

prog/dftb+/lib_dftb/h5correction.F90
bhourahine: newunit mechanism please.

prog/dftb+/lib_dftb/dispdftd3.F90
```diff
@@ -261,4 +278,50 @@ function getRCutoff(this) result(cutoff)
...
+      if (this%izp(i) == 1) then
+        do j = i + 1, this%nAtom
+          if (this%izp(j) == 1) then
+            ! Distance in a.u.
+            r = sqrt((coords(1,i) - coords(1,j))**2 + (coords(2,i) - coords(2,j))**2 &
+                & + (coords(3,i) - coords(3,j))**2)
+            ! Repulsion energy in a.u.
+            repE = repE + k * (1.0_dp - 1.0_dp/(1.0_dp + exp(-e*(r/r0 - 1.0_dp))))
+            ! Derivative
+            dEdR = -k * (1.0 / (1.0 + exp(-e*(r/r0-1.0)))**2 * e/r0 * exp(-e*(r/r0-1.0)))
```
bhourahine: 1.0_dp

prog/dftb+/lib_dftb/dispdftd3.F90
```diff
@@ -261,4 +278,50 @@ function getRCutoff(this) result(cutoff)
...
+    class(DispDftD3), intent(inout) :: this
+    !> Current coordinates
+    real(dp), intent(in) :: coords(:,:)
+
+    integer :: i, j, c
+    real(dp) :: r, repE, dEdR, dCdR
+
+    ! Parameters (with conversion to a.u.)
+    real(dp) :: k = 0.30_dp * 0.001593601371772_dp
+    real(dp) :: e = 14.31_dp
```
bhourahine: Atomic units internally please.

test/prog/dftb+/tests
```diff
@@ -160,3 +160,8 @@ geoopt/Vsi+O_lbfgs                     #? MPI_PROCS <= 4
 geoopt/Vsi+O                           #? MPI_PROCS <= 4
 scc/SiC64+V_dynforce_groups            #? MPI_PROCS <= 8 and MPI_PROCS % 4 == 0 
 scc/SiC64+V_dynforce                   #? MPI_PROCS <= 4 and MPI_PROCS % 2 == 0
+
+h-bonds/H5_defaults                    #? MPI_PROCS <= 1
+h-bonds/H5_ONS_forces                  #? MPI_PROCS <= 1
+h-bonds/H5_ONS_forces_eq               #? MPI_PROCS <= 1
+h-bonds/D3H5_water_dimer               #? MPI_PROCS <= 1
```
bhourahine: #? WITH_DFTD3 and MPI_PROCS <= 1
For this specific example

prog/dftb+/lib_dftb/h5correction.F90
```diff
@@ -0,0 +1,243 @@
...
+    this%speciesName = typeNames
+
+    ! Save H5 parameters
+    this%rScale = r
+    this%wScale = w
+    this%elementPara = elePara
+
+    ! Initialize the parameters
+    ! If value of the parameter is -1.0, it was not read from the input,
+    ! and default value will be used
```
bhourahine: Could the default universal values be over-ridden in the input?

prog/dftb+/lib_dftb/h5correction.F90
```diff
@@ -0,0 +1,243 @@
...
+
+    ! Get parameters for current pair of species
+    call this%getParams(iSp1, iSp2, applyCorrection, h5Scaling, sumVDW)
+
+    ! If applicable to the current pair, modify the gamma
+    if (applyCorrection) then
+      ! Gaussian calculation
+      fwhm = this%wScale * sumVDW
+      r0 = this%rScale * sumVDW
+      c = fwhm / 2.35482_dp
```
bhourahine: Where does the scaling factor come from? Not obvious how this is related to the s_r values in the paper (in the text that is 0.75, in table 1 its 0.714).

prog/dftb+/lib_dftb/h5correction.F90
Honza-R: Full width at half maximum (FWHM) is used to provide a physical meaning to the parameter, and this is then converted to the "c" parameter, dividing it by 2(2 ln 2)^0.5. I'm adding a comment to the code to explain this constant.

prog/dftb+/lib_dftb/h5correction.F90
Honza-R: It returns applyCorrection=.false. unless there exist parameters for the pair. I'm adding comment on that.

prog/dftb+/lib_dftb/dispdftd3.F90
Honza-R: I've switched to conversion factors from constants.F90 everywhere.

prog/dftb+/lib_dftb/scc.F90
Honza-R: Fixed.

prog/dftb+/prg_dftb/parser.F90
Honza-R: Yes, it's a modification of the dispersion that can be used without H5, when correction of H-bonds is not needed or a different one is used.

prog/dftb+/lib_dftb/h5correction.F90
Honza-R: Changed to a named file with unit number obtained using newunit.

prog/dftb+/prg_dftb/parser.F90
Honza-R: Fixed.

prog/dftb+/lib_dftb/dispdftd3.F90
Honza-R: I'm converting the units in the code, rather than providing converted values as constants, so that the value of the parameter in the code matches what is published in the paper.
Changed to using constants.F90 conversion factor.

prog/dftb+/lib_dftb/scc.F90
Honza-R: I've switched to allocated(inp%h5Correction) in the input data module as suggested. However, I still keep a separate flag for it in the scc module for performance reason, as it is evaluated repeatedly.

prog/dftb+/lib_dftb/h5correction.F90
Honza-R: I've switched to constants from constants.F90 everywhere.

prog/dftb+/lib_dftb/h5correction.F90
Honza-R: Changed to using constants.F90.

aradi:
:+1:  Very nice code, thanks a lot for all the work which went in it so far! 

I've recalculated the h-bond tests as many of them used the old inaccurate non-scc forces. I've also made a few minor style changes. They are still a few subtleties I'd like to have a second look at before merging, though.

aradi:
Sorry for the long delay. As for the code, I am fine with it. I'd suggest, however, some changes in the parser keywords. What about the following structure:
```
 HBondCorrection = H5 {
        RScaling = 0.714
        WScaling = 0.25
        H5Scaling {
          O = 0.06
        }
    }
```
Also, what I miss is a more detailed entry in the manual, especially the manual should contain an a  direct correspondence of the parser keywords to the symbols (s_r, s_w, k_XY) in the paper. I can eventually do it myself, if we agree on the keywords.

bhourahine:
The proposed syntax looks reasonable, so sure.

Honza-R:
I agree, it looks more clear now.

Honza-R:
Hi, thanks for extending the manual. I found one issue in the formula for the "w" parameter, there's a conversion factor that gives the scaled sum of radii meaning of full-width at half-maximum of the gaussian. (I'm sorry, but it's a thing that somehow dropped from the paper, I should write errata for that.) I added it to the formula along with a brief explanation. 

I also realized I should add one more thing to the manual to make the description complete. In our paper, we did not apply the correction to azide nitrogens because already uncorrected DFTB3 overestimates their interactions. I added a note to the manual on this, with a tip that this can be achieved by defining a new atom type in DFTB+ when there are different types of nitrogens in the system.

I'd appreciate if Ben can have look at the wording of my edits. But for the contents, I think everything correct and complete now.

PR #133, opened 2018-04-01
Removed fdEigvec in favour of newunit

bhourahine:
First of the file units replaced with the newunit mechanism.


bhourahine:
All units not using the finalize mechanism are now converted to use the newunit language feature.

PR #135, opened 2018-04-05
Cleaned up DFTB+U routines and allow spin free calculations

bhourahine:
Same results for spin free test case as spin polarized with S=0
and all spin constants 0.


aradi:
I am fine with it. I just made it more compact, see PR to your branch.

PR #146, opened 2018-05-24
Time propagation

fbonafe:
This is a implementation of real time propagation of the density matrix for frozen ions (no nuclear dynamics) under external time dependent perturbations (Delta kick, CW laser and pulse with different envelopes).


aradi:
Replaced by #329

PR #152, opened 2018-06-15
Autotest binary path

bhourahine:
Enabled the pre/pos run scripts to run dftb+ if needed. Autotest examples using this mechanism to generate files for restarting calculations that are actually tested.


bhourahine:
Changed to look for a testrun.sh script, but otherwise run DFTB+.

PR #153, opened 2018-06-22
Parser bump

bhourahine:
Change to parser 6, fixing 'tollerance' in the Pipek-Mezey localization and unifying syntax for the two hydrogen bond correction schemes currently available.

Ready to merge (unless we want to also add setting a default value for the H-X damping correction exponent at this parser revision).

### DIFF
--- a/prog/dftb+/prg_dftb/oldcompat.F90
+++ b/prog/dftb+/prg_dftb/oldcompat.F90
@@ -53,6 +53,9 @@ contains
       case (4)
         call convert_4_5(root)
         version = 5
+      case (5)
+        call convert_5_6(root)
+        version = 6
       end select
     end do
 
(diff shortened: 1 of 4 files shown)


prog/dftb+/prg_dftb/oldcompat.F90
```diff
@@ -324,12 +327,72 @@ subroutine convert_4_5(root)
...
+      call detailedWarning(ch1, "Keyword converted to 'SparseTollerances'.")
+      call setNodeName(ch1, "SparseTolerances")
+    end if
+
+    call getDescendant(root, "Analysis/Localise/PipekMezey/Tollerance", ch1)
+    if (associated(ch1)) then
+      call detailedWarning(ch1, "Keyword converted to 'Tolerance'.")
+      call setNodeName(ch1, "Tolerance")
+    end if
+
```
aradi: Isn't this block just a repetition of the first block in the subroutine?

prog/dftb+/prg_dftb/oldcompat.F90
```diff
@@ -324,12 +327,72 @@ subroutine convert_4_5(root)
...
+      call destroyNode(ch2)
+
+      ! clean out any HBondCorrection entry
+      call getDescendant(root, "Hamiltonian/DFTB/HBondCorrection", ch2, parent=par)
+      if (associated(ch2)) then
+        call detailedError(ch2, "HBondCorrection already present.")
+      end if
+
+      call getDescendant(root, "Hamiltonian/DFTB", ch2, parent=par)
+      if (associated(ch2)) then
```
aradi: I think the ``if`` is superfluos. ``Hamiltonian/DFTB`` must exist when this point is reached since ``Hamiltonian/DFTB/DampXH`` was checked to exist above.

prog/dftb+/prg_dftb/oldcompat.F90
bhourahine: Yes, removed it.

prog/dftb+/prg_dftb/oldcompat.F90
bhourahine: Yes, removed it.

PR #155, opened 2018-06-26
Option to disable excited state forces

bhourahine:
Relavant for MD in the ground state with excitation calculation
for cases with spin/fractional occupation numbers.

Also md.out now accumulates all of the calculated excitation
energies.


doc/dftb+/manual/dftbp.tex
```diff
@@ -3079,6 +3080,13 @@ \subsection{Casida}
   \item[\is{TestArnoldi}] If set to \is{Yes}, the file TEST\_ARPACK.DAT is created, which gives data
     on the quality of the resulting eigenstates.
 
+  \item[\is{ExcitedStateForces}] If ground state forces are enabled, either due
+    to a geometry driver (p.~\pref{sec:dftbp.Driver}) or requested in
+    \is{Analysis} (p.~\pref{sec:dftbp.Analysis}), this option includes the
+    contributions from the excited state of interest (if set to \is{Yes}). This
+    option can be used, for example, to calculate excitation energies during a
+    ground state molecular dynamics simulation.
+
```
aradi: I find the example confusing, as users would ask: Why do I need this option to calculate excited states energies during MD? Why not: By setting this option to No, .... (Not setting the option explicitely to No during the molecular dynamics may prevent the calculation of the excited state spectrum for those cases, where the excited states can not be calcualted)

Or something similar...

test/prog/dftb+/timedep/OCH2-md/dftb_in.hsd
```diff
@@ -0,0 +1,63 @@
...
+        StateOfInterest =  0
+        Symmetry = Both
+        WriteMulliken = No
+        ExcitedStateForces = No
+    }
+}
+
+ParserOptions {
+    ParserVersion = 5
+}
```
aradi: Let's use current parser version

prog/dftb+/lib_timedep/linrespgrad.F90
```diff
@@ -523,6 +526,16 @@ subroutine LinRespGrad_old(tSpin, natom, iAtomStart, grndEigVecs, grndEigVal, sc
...
+      if (allocated(allOmega)) then
+        if (size(allOmega) /= size(symmetries) * nExc) then
+          deallocate(allOmega)
+        end if
+      end if
+      if (.not. allocated(allOmega)) then
+        allocate(allOmega(size(symmetries) * nExc))
+      end if
+      allOmega(1+(iSym-1)*nExc:iSym*nExc) = sqrt(eval)
+
```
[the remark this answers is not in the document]
bhourahine: Not with current code structure. If substantially re-written, the number of excited states could in principle vary (but that suggests nExc was set to be too small in the first case).

doc/dftb+/manual/dftbp.tex
bhourahine: Not sure I follow the replacement text either. Have re-written the text, so please review .

test/prog/dftb+/timedep/OCH2-md/dftb_in.hsd
bhourahine: Updated

aradi:
I pushed a suggestion for the manual. If you are happy with it (feel free to correct it :wink:), the branch can be merged.

PR #156, opened 2018-06-28
Change Intel mpi compiler to mpifort

bhourahine:
Due to the 2017 change in the Intel tools wrapper naming
convention.

See for [example](http://www.hpc.cineca.it/center_news/important-use-intel-mpi-wrappers-mpif90-mpicc-mpicxx).

### DIFF
--- a/sys/make.x86_64-linux-intel
+++ b/sys/make.x86_64-linux-intel
@@ -9,7 +9,7 @@ ifeq ($(strip $(WITH_MPI)),1)
 ############################################################################
 
 # Compilers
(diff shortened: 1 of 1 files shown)


aradi:
I think, this is not compiler but MPI-framework related. MPICH 1.5 (compatible with MKL 2016) does not have it yet, only mpif90.

bhourahine:
Added the comment we discussed above the change.

aradi:
I've changed all mpi-wrapper names to mpifort and merged.

PR #157, opened 2018-06-28
Option to skip read in charge checks

bhourahine:
Option to switch off testing total charge / magnetization against current value if read from disc.


bhourahine:
Just realised this is useful for transport restarts, so would be useful to merge prior to merging the trunk onto that branch.

PR #164, opened 2018-08-10
SK truncation

bhourahine:
Input options to artificially shorten the SK tables to a set distance.


prog/dftb+/prg_dftb/parser.F90
```diff
@@ -1333,8 +1333,36 @@ subroutine readDFTBHam(node, ctrl, geo, slako)
     else
       skInterMeth = skEqGridNew
     end if
-    call readSKFiles(skFiles, geo%nSpecies, slako, slako%orb, &
-        &angShells, ctrl%tOrbResolved, skInterMeth, repPoly)
+    call getChild(node, "TruncateSKRange", child, requested=.false.)
+    if (associated(child)) then
+      call warning("Artificially truncating the SK table, this is normally a bad idea!")
+      ! Artificially truncate the SK table
+      call getChildValue(child, "SKMaxDistance", truncationCutOff, -1.0_dp, modifier=modifier,&
```
aradi: IMO, we should not have a default here. The block `TruncateSKRange` is already optional, if it has been specified, the user should provide a max distance. We could then also get rid of the misuse of `-1.0` for indicating a special choice. :wink:

prog/dftb+/prg_dftb/parser.F90
```diff
@@ -1333,8 +1333,36 @@ subroutine readDFTBHam(node, ctrl, geo, slako)
...
+          & child=field)
+      call convertByMul(char(modifier), lengthUnits, field, truncationCutOff)
+      call getChildValue(child, "HardCutOff", tFound, .true.)
+      if (tFound) then
```
aradi: Could you take a logical variable with more descripting name, rather than reuse an existing one. (Maybe putting the entire `TruncateSKRange` parsing into a separate subroutine would be the cleanest...

test/prog/dftb+/tests
```diff
@@ -90,6 +90,13 @@ scc/C60                                #? MPI_PROCS <= 4
 analysis/C60_ESP                       #? MPI_PROCS <= 4
 spin/H2O                               #? MPI_PROCS <= 1
 
+#! This block of tests are for code features capable of producing
+#! unphysical results
```
aradi: As we order the tests according timing, this block is not warranted to stay intact, so probably we should omit the comment. (The directory name dangerous should be scaring enough already...)

test/prog/dftb+/tests
```diff
@@ -90,6 +90,13 @@ scc/C60                                #? MPI_PROCS <= 4
...
+dangerous/C_HardN_defaultSKinterp      #? MPI_PROCS <= 4
+dangerous/C_HardY_defaultSKinterp      #? MPI_PROCS <= 4
+dangerous/C_HardN_oldSKinterp          #? MPI_PROCS <= 4
+dangerous/C_HardY_oldSKinterp          #? MPI_PROCS <= 4
+
```
aradi: Maybe,  `C_SkCutoff_hard_default`, `C_SkCutoff_soft_default` would be more descriptive (the main feature of the tests is the sk-cutoff).

test/prog/dftb+/tests
bhourahine: Removed, but all the test would take about the same time.

test/prog/dftb+/tests
bhourahine: Renamed

prog/dftb+/prg_dftb/parser.F90
bhourahine: Now packaged as a subroutine without a default for SKMaxDistance

PR #173, opened 2018-09-13
Use the $(PREFIX) variable in makefiles instead of hardcoding the path

yurivict:


yurivict:
You are right, I've updated the PR. ```PREFIX``` is among the variables that are commonly set for GNU make. Particularly, the complete GNU toolchain, and cmake-based toolchains always run GNU make with ```PREFIX``` set to the location where dependent packages are installed, and where this package is going to be installed.

aradi:
See PR #176. It includes your changes plus some additional ones to avoid redefinition of the `PREFIX` variable several times.

PR #174, opened 2018-09-13
Add missing -c flag to the xmlf90 makefile

yurivict:


external/xmlf90/make.dpbuild
```diff
@@ -31,7 +31,7 @@ clean:
 	rm -f *.o _mod_* *.mod
 
 %.o: %.f90
-	$(FXX) $(FXXOPT) -o $@ -c $<
+	$(FXX) -c $(FXXOPT) -o $@ -c $<
```
aradi: But `-c` is already in the command line. What's the problem?

yurivict:
Sorry, my bad. I was under impression that it lacks ```-c```, but apparently it isn't.

PR #176, opened 2018-09-18
Use prefix var in makefiles

aradi:
Extends the external request by rearranging the makefiles as discussed already earlier. Also, it fixes the wrong license information / notes about optional external components with GPL-license

### DIFF
--- a/make.config
+++ b/make.config
@@ -8,9 +8,6 @@ BUILDDIR := $(ROOT)/_build
 # Installation directory
 INSTALLDIR := $(ROOT)/_install
 
-# Directory where packages are installed in the system
-PREFIX := /usr/local
-
 ################################################################################
 # Configuration of optional components
 ################################################################################
@@ -50,24 +47,6 @@ TEST_OMP_THREADS := 1
 # Architecture dependent settings
 ################################################################################
 
-# Whether DFTD3 should be compiled during the build process. If set to 1 (yes),
-# you will have to download the dftd3 library before starting the build using
-# the utils/get_opt_externals tool.
-# (Only active, if WITH_DFTB3 was set to 1 above.)
-COMPILE_DFTD3 := 1
-
-# Set the compile time include and the link time library options for
-# dftd3-lib. Ignored if WITH_DFTD3 has been disabled or COMPILE_DFTD3 enabled.
-DFTD3_INCS := -I$(PREFIX)/include/dftd3-lib
-DFTD3_LIBS := -L$(PREFIX)/lib -ldftd3
-
-# Link time library options for linking ARPACK. Ignored if WITH_ARPACK was
-# disabled.
-ARPACK_LIBS := -larpack
-
-# Whether ARPACK depends on the external LAPACK and BLAS libraries
-ARPACK_NEEDS_LAPACK := 0
-
-# Include further architecture dependent settings from make.arch (make sure to
-# adapt those in make.arch for your system)
+# Include architecture dependent settings from make.arch (make sure to adapt
+# those in make.arch for your system)
 include $(ROOT)/make.arch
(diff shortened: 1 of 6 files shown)


make.config
```diff
@@ -23,8 +23,9 @@ WITH_SOCKETS := 1
 WITH_ARPACK := 1
 
 # Whether the DFTD3 library (dispersion) should be linked.
-# NOTE: Due to license incompatibility, distribution of a DFTB+ binary built
-# with this component is not permitted. Use it only for your personal research.
+# NOTE: Due to the license of the DFTD3 library, the linked code must be
```
bhourahine: Suggest 'combined code' instead of 'linked code'

sys/make.x86_64-linux-gnu
```diff
@@ -3,6 +3,10 @@
 # gfortran/gcc 5.4 - 7.1
 ############################################################################
 
+# Directory where standard packages are installed in the system
+PREFIX := /usr/local
```
bhourahine: Not sure if this should be /usr/ instead if its a 'standard' install ( or even /opt/ ).

utils/get_opt_externals
```diff
@@ -22,26 +22,28 @@ Downloads optional external components.
 
 Possible choices for OPTEXT:
 
-  DISTRIB (default) -- download only components which are LGPL-compatible,
-      can be distributed with the DFTB+ source code and do not restrict the
+  UNRESTRICTED (default) -- download only components which are less or equally
+      restrictive than the LGPL license of DFTB+ and do not restrict the
```
bhourahine: "restrictive as the LGPL"

utils/get_opt_externals
```diff
@@ -22,26 +22,28 @@ Downloads optional external components.
...
       distribution of the compiled code beyond its LGPL-license.
 
+  RESTRICTED -- download optional components with stricter license as LGPL
+      (typically GPL). The distribution of the compiled code will be restricted
```
bhourahine: (typically GPL2 without the "or later" permission)

utils/get_opt_externals
```diff
@@ -22,26 +22,28 @@ Downloads optional external components.
...
+
   ALL -- download all optional external components.
 
-      NOTE: this will also download components with LGPL-incompatible
-      licenses. Those components can not be distributed with the DFTB+ source
-      code and the distribution of the resulting binary or library is not
-      permitted if those components were part of the build. Using them for
-      personal research is permitted, though.
+      NOTE: this will also download components which may restrict the license of
+      the compiled code beyond the LGPL license.
```
bhourahine: LGPL3

make.config
aradi: Done.

sys/make.x86_64-linux-gnu
aradi: I think `/usr` is the correct choice if things were installed as packages. I changed this in all 4 makefiles.

utils/get_opt_externals
aradi: Thanks!

utils/get_opt_externals
aradi: I made it now more explicit at several places.

aradi:
Can you have a quick look on the changed wording, please, before I merge? Thx!

utils/get_opt_externals
```diff
@@ -22,26 +22,28 @@ Downloads optional external components.
 
 Possible choices for OPTEXT:
 
-  DISTRIB (default) -- download only components which are LGPL-compatible,
-      can be distributed with the DFTB+ source code and do not restrict the
+  UNRESTRICTED (default) -- download only components which are less or equally
+      restrictive as the LGPLv3 license of DFTB+ and do not restrict the
```
bhourahine: than the LGPLv3

utils/get_opt_externals
```diff
@@ -22,26 +22,28 @@ Downloads optional external components.
...
       distribution of the compiled code beyond its LGPL-license.
 
+  RESTRICTED -- download optional components with stricter license as the LGPLv3
```
bhourahine: stricter license than the

PR #178, opened 2018-10-09
GPU

quitocam:


aradi:
Thanks a lot for the PR! I would need a week or two before I can have a detailed look at it (and especially test it myself).

But just a question, which came to my mind glancing over the code: Do we really need to use nvcc as compiler? (As this seems to be hard-coded in the makefile.) I thought, since we use Magma calls onyl, our build system does not necessarily have to rely on NVIDIA tools any more. (Anyway, that makefile would have to be cleaned up anyway as it contains hard-wired paths on your system :wink: )

quitocam:
Sure! I removed the use of the nvcc compiler. One important thing is that MAGMA relies a lot on CUDA. Although there are not any calls to CUDA and everything is handle by MAGMA, we are still required to use some CUDA includes. Sadly, we do not have an ATI card to test the code with clMAGMA.

aradi:
That's is absolutely the plan. The only bottleneck currently is that my daily testing framework does not have access to any GPU machines. Talking about testing: @irles would it be possible, that you test the code as it is in this pull request? It would be really good to have 3rd party testing before any official merge. :wink:

irles:
Absolutely!  You probably mean a single test e.g. on Summit.

Stephan

aradi:
Exactly. :smile:

lib_device/functions.cpp
```diff
@@ -0,0 +1,21 @@
+#include <magma.h>
+
+// count GPUs function
+
+
+extern "C"  void  MagmaGpuNum_Available(int* max_ngpus)
+{
+    magma_device_t devices[117]; // 117 is large enough to fit all device IDs
```
bhourahine: Where is 117 coming from? Also as the value repeats in the getdevices call below, probably should be a variable to keep synchronized if any changes happen.

prog/dftb+/lib_math/eigensolver.F90
```diff
@@ -2042,4 +2054,368 @@ subroutine dblecmplx_zhbgv(ab, bb, w, uplo, z)
...
+
+    !> Padding for shared memory to avoid bank conflicts, 32-bit mode
+    nwarp = 32      
+    ldm   = nwarp* ((n+nwarp-1)/nwarp)
+    allocate(H(ldm*ldm), S(ldm*ldm),stat=info) 
+    if (info.ne.0) call  error('CPU alloc error in real_magma_ssygvd. Exiting')
+    
+    !> Convert nxn matrices to 1D arrays
+    i0=1
+    do i=1,n
```
bhourahine: Fortran intrinsic reshape probably faster, might be simpler code too.

prog/dftb+/lib_math/eigensolver.F90
```diff
@@ -2042,4 +2054,368 @@ subroutine dblecmplx_zhbgv(ab, bb, w, uplo, z)
...
+    liwork = int(iwork(1))
+    deallocate(work) ;  allocate(work(lwork))
+    deallocate(iwork) ; allocate(iwork(liwork))
+    
+    !> MAGMA Diagonalization
+    call magmaf_ssygvd_m(ngpus,iitype,jobz,uplo,n,H,ldm,S,ldm,w,work,lwork,iwork,liwork,info)
+
+    i0=1
+    do i=1,n 
+      call scopy(n,H(i0),1,a(1,i),1)
```
bhourahine: reshape going the other way?

prog/dftb+/prg_dftb/initprogram.F90
```diff
@@ -78,8 +78,18 @@ module initprogram
...
   use formatout
+#:if WITH_GPU
+  use iso_c_binding, only :  c_int
+  use device_info
+#:endif
+
   implicit none
 
+#:if WITH_GPU
+  integer (c_int) :: ngpus
```
bhourahine: The current magma library fortran wrappers for the actual calls seem to be integer not c_int

prog/dftb+/prg_dftb/eigenvects.F90
```diff
@@ -16,6 +16,10 @@ module eigenvects
   use message
 #:if WITH_SCALAPACK
   use scalapackfx
+#:endif
+#:if WITH_GPU
+  use initprogram, only: ngpus
```
bhourahine: Should be passed in, not as a global module variable

quitocam:
Working on it already!

bhourahine:
You might want to have a look at (or merge from)
https://github.com/bhourahine/dftbplus/tree/GPUupdate
which addresses some of the comments (particularly the lib_devices location) as well as being up to date with the changes on the master branch.

bhourahine:
Incidentally, do have you any ideas about getting the density matrix building via the ?dsyrk and ?herk calls onto the GPU? For CPU only calculations these typically take about 5-10% of the total time, with the rest being the  eigensolver.

quitocam:
Although we did a full profile of the code, I have not thought about it, but we can have a look into it to see if we can come up with something useful.

quitocam:
We have include virtually all given recommendations and have update it to the latest development code. Test examples are in the work. We did not include the rehash to optimize the code for the GPU.

aradi:
The delay is at our side. As other solvers are also entering the code (#175), we had to redesign slightly the solver interfaces to be able to handle it. Once those are on master (should happen next week), we would start to merge this branch as well.

prog/dftb+/lib_math/eigensolver.F90
scienception: I didn't change this because although is a nxn matrix conversion to a 1-dimensional one, we usually pad the number of rows to a multiple of 32 on the GPU.

bhourahine:
All changes included via merge of #279 routines.

PR #203, opened 2019-01-06
Onsites general

bhourahine:
Ground and excited state onsite corrections.
Note, as with orbital potentials, XLBOMD is not compatible. Excited state does not include onsite dipole allowed contributions (as is missing from the ground state formalism at the moment).


bhourahine:
See commit f8c2643f for a slight speed up

thomas-niehaus:
Much better!

PR #208, opened 2019-01-19
Transition charge caching

bhourahine:
Option to store occupied-> virtual transition charges


aradi:
I get failing tests due to array boundary problems. I am investigating, will pose a PR if I found the bug.

bhourahine:
Just corrected the array bound issue.

prog/dftb+/lib_timedep/linrespgrad.F90
```diff
@@ -1172,7 +1172,7 @@ subroutine getZVectorEqRHS(xpy, xmy, win, iAtomStart, homo, nocc, nmatup, getij,
     integer :: nxov, nxoo, nxvv
     integer :: i, j, a, b, ia, ib, ij, ab, ja
     real(dp) :: tmp1, tmp2
-    logical :: updwn
+    logical :: updwn = .true.
```
aradi: But that makes ``updwnd`` a saved variable! Is that intentional?

bhourahine:
Those routines are not spin-polarization adapted, so its always within the same spin channel and could even be set as a parameter.  But once (if) they are generalised this would need to be revisited.
The is a secondary question of the name of the variable vs its logical choice, at the moment updown=.true. is for same-spin transitions, and .false. is intra-spin, which looks a little weird.

PR #216, opened 2019-02-19
Updated submodule libnegf

bhourahine:
Brings the submodule up to date with recent bitbucket check ins.


aradi:
I did that already during fixing buildbot issues.

PR #217, opened 2019-02-21
Fix for some OMP loops in q_ij cache

bhourahine:


aradi:
Are you sure ``!!$OMP`` works as OMP-directive? (I was only aware of `!$OMP` so far.

bhourahine:
Yes, these were commented out, but are now correct but still commented out

PR #221, opened 2019-02-27
Fixed case where natural orbitals not returned

bhourahine:
Fixes analysis case where there isn't workspace for natural orbitals passed in from the main code.


aradi:
Has been already fixed exactly the same way in a personal branch of mine, however, the program than crashes at an other point. I am still investigating it...

PR #222, opened 2019-02-28
Removed redundant solveZVectorEq init for rhs

bhourahine:
The initial rhs2 set-up is over-written by a unit vector, so
can be removed.

### DIFF
--- a/prog/dftb+/lib_timedep/linrespgrad.F90
+++ b/prog/dftb+/lib_timedep/linrespgrad.F90
@@ -1379,25 +1379,11 @@ contains
     integer :: ia, i, a, k
     real(dp) :: rhs2(size(rhs)),rkm1(size(rhs)),pkm1(size(rhs)),apk(size(rhs))
     real(dp) :: qTmp(nAtom), rs, alphakm1, tmp1, tmp2, bkm1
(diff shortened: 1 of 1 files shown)


bhourahine:
Do you have something to hand as an alternative for the routine?
Timing the C66O10N4H44 molecule example did not show anything significant (actually slightly slower by .5 seconds if using the removed rhs code). 
For large systems,  as windowing and forces are not implemented together, is this solver a significant time cost compared to the eigensolver?

thomas-niehaus:
The routine takes around 50 % CPU time in the computation of the TD-LC-DFTB force evaluation. Have not tested TD-DFTB. I have no replacement for the moment, but since this is just solution of a large SLE, I thought something should be available. Don't you do similar things in the Sternheimer method?

thomas-niehaus:
Ok. But isn't 
rhs2 = 1.0_dp / sqrt(real(nxov,dp))
overwriting the preconditioning?

bhourahine:
The initial guess didn't seem to matter much (for TD-DFTB, the time to set up the guess was more significant in the test). This isn't currently preconditioned CG anyway (will try that out for this code).

bhourahine:
Now superseded by #223 using preconditioned CG

PR #226, opened 2019-03-08
Preprocessing to catch unallocated variable

bhourahine:
PreprocessorOpts isn't allocated unless compiled with MPI, but was being tested in the serial code.


bhourahine:
In the parser
if (withMpi .and. .not. associated(node)) then
      call setChild(root, "Parallel", node)
end if
leads to it not being allocated for serial compilation.

PR #229, opened 2019-03-18
Simple API to use DFTB+ as energy/force calculator

aradi:


bhourahine:
The INSTALL.rst probably should be updated to mention that a C compiler is needed only for building the API or for socket compilations.

api/mm/make.build
```diff
@@ -0,0 +1,59 @@
+#------------------------------------------------------------------------------#
+#  DFTB+: general package for performing fast atomistic simulations            #
+#  Copyright (C) 2017  DFTB+ developers group                                  #
+#                                                                              #
+#  See the LICENSE file for terms of usage and distribution.                   #
+#------------------------------------------------------------------------------#
+
+TARGETS = libdftb+.a
```
bhourahine: Is the library always intended to be static?

api/mm/make.build
```diff
@@ -0,0 +1,59 @@
...
+
+# Include common makefile
+SRCDIRS := $(ROOT)/api/mm
+SRCDIRS += $(wildcard $(ROOT)/prog/dftb+/lib_*)
+SRCDIRS += $(ROOT)/prog/dftb+/include
+INCDIRS := $(ROOT)/prog/dftb+/include
+include $(ROOT)/prog/dftb+/make.common
+
+libdftb+.a: $(dftbplus.o)
+	ar r $@ $(filter-out _%,$^)
```
bhourahine: archive tool choice (along with its options) should probably be set by the sys make files

bhourahine:
Disabling the C compiler (setting CC = ) and running
make test
runs the Fortran API tests successfully and tries, but fails, the C cases. Is is worth trying to detect if CC is defined and skipping the C cases if not?

test/prog/dftb+/bin/autotest2
```diff
@@ -196,12 +196,12 @@ case "${STAGE_SEP}$stages" in
 esac
 
 
-# Check if executable is there:
-if [ ! -x "${app_cmd}" -a -n "${STAGE_RUN}" ]; then
-  echo "${SCRIPTNAME}: Error: Executable '${app_cmd}' does not exist" 1>&2
-  echo "Type '${SCRIPTNAME} -h' for help" 1>&2
-  exit 1
-fi
+## Check if executable is there:
```
bhourahine: Any point in saving this and having a flag to turn the check on/off for use with the program instead of the API?

test/api/mm/tests
```diff
@@ -0,0 +1,8 @@
+extcharges
```
bhourahine: Would be good to have 
a) a header
b) C equivalents to all of the tests (presumably people may use these as templates, especially if point to in the documentation).

utils/srcmanip/change_module_name
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python
+################################################################################
+#
+# Replaces the name of a given module with a new one in all the files passed as
```
bhourahine: Replaces (in place) the name

api/mm/API_VERSION
```diff
@@ -0,0 +1 @@
+0.2.0
```
bhourahine: Version should be mentioned in changelog and documentation. Also, is there a mechanism for the regression tests to make reference to the API version choice?

doc/dftb+/manual/api.tex
```diff
@@ -0,0 +1,35 @@
...
+here.
+
+You can test the library functionality by issuing
+\begin{verbatim}
+make test_api
+\end{verbatim}
+
+
+\section{General guidelines}
+
```
bhourahine: Is it worth commenting on the choice of atomic units for internal variables?

api/mm/make.build
aradi: No, but with the current build system it's easier this way (shared library would need additional compiler options.)

api/mm/make.build
aradi: I agree. However, currently all external components have hardcoded "ar" command lines. Again, a later change to CMake should resolve this issue, as CMake would set the correct command line automatically.

bhourahine:
> The INSTALL.rst probably should be updated to mention that a C compiler is needed only for building the API or for socket compilations.

Done in 2f89d8bb

api/mm/make.build
bhourahine: OK, deffer to cmake

api/mm/make.build
bhourahine: OK defer to cmake

bhourahine:
> Disabling the C compiler (setting CC = ) and running
> make test
> runs the Fortran API tests successfully and tries, but fails, the C cases. Is is worth trying to detect if CC is defined and skipping the C cases if not?

OK, done in 2f89d8bb and be9b0cda

bhourahine:
> Looks good. Should the test examples now have #? not WITH_MPI

Done in 2f89d8b and be9b0cd

doc/dftb+/manual/api.tex
bhourahine: Done in 2f89d8bb

api/mm/API_VERSION
bhourahine: Done in 2f89d8bb

utils/srcmanip/change_module_name
bhourahine: Done in 2f89d8bb

test/prog/dftb+/bin/autotest2
bhourahine: Done in 2f89d8bb

PR #231, opened 2019-03-19
CustomReferenceOccupations

pecchia:
Added option CustomReferenceOccupations
Added test as test/scc/SiH_custom-ref
Testsuite run
Added manual description


prog/dftb+/prg_dftb/initprogram.F90
```diff
@@ -2052,17 +2062,17 @@ subroutine initProgramVariables(input, env)
 
     tWriteChrgAscii = input%ctrl%tWriteChrgAscii
 
-    if (tReadChrg) then
+    !if (tReadChrg) then
```
bhourahine: Any reason for commenting out these lines instead of removing? I think I see why they are redundant.

doc/dftb+/manual/dftbp.tex
```diff
@@ -1260,6 +1261,33 @@ \section{Hamiltonian}
...
+    }
+    ReferenceOccupation{
+      Atoms={31:60}
+      p=1.99
+    }
+  }
+  \end{verbatim}
+  
+  The example above sets a filling population of +0.01e or -0.01e in the p shell of the corresponding 
+  atom indeces. When the states are filled up, the electron excess or depletion results in a shift of
```
bhourahine: indeces -> indices

make.config
```diff
@@ -13,7 +13,7 @@ INSTALLDIR := $(ROOT)/_install
 ################################################################################
 
 # Whether DFTB+ should support MPI-parallelism
-WITH_MPI := 0
+WITH_MPI := 1
```
bhourahine: Please revert to 0

make.config
```diff
@@ -26,11 +26,11 @@ WITH_SOCKETS := 0
 
 # Whether the ARPACK library (needed by TD-DFTB) should be linked with DFTB+
 # Only affects serial build (MPI-version is built without ARPACK/TD-DFTB).
-WITH_ARPACK := 0
+WITH_ARPACK := 1
```
bhourahine: please revert to 0

make.config
```diff
@@ -26,11 +26,11 @@ WITH_SOCKETS := 0
...
 
 # Whether transport via libNEGF should be included.
 # Only affects parallel build (serial version is built without libNEGF/transport)
-WITH_TRANSPORT := 0
+WITH_TRANSPORT := 1
```
bhourahine: please revert to 0

make.config
```diff
@@ -46,7 +46,7 @@ WITH_DFTD3 := 0
 # 0 -- production code
 # 1 -- internal runtime checks and reduced compiler optimization
 # 2 -- level 1 checks plus extra runtime compiler checks
-DEBUG := 0
+DEBUG := 0
```
bhourahine: Remove white space at line end.

prog/dftb+/lib_dftb/sccinit.F90
```diff
@@ -131,6 +132,47 @@ subroutine initQFromShellChrg(qq, qShell, species, orb)
 
   end subroutine initQFromShellChrg
 
+  !> Initialise charge vector from user-defined custom charges.
```
bhourahine: user-defined custom reference atomic charges.

prog/dftb+/prg_dftb/parser.F90
```diff
@@ -5142,6 +5145,64 @@ end subroutine finalizeNegf
 #:endif
 
 
+  !> This subroutine overrides the neutral (reference) atom electronic occupation 
+  subroutine readCustomReferenceOcc(root, orb, referenceOcc, geo, iAtInRegion, &
+      & customOcc)
+    type(fnode), pointer, intent(in) :: root
```
bhourahine: Please comment the interface variables

prog/dftb+/prg_dftb/parser.F90
```diff
@@ -5142,6 +5145,64 @@ end subroutine finalizeNegf
...
+      end if
+      atomOverriden(iAtInRegion(iCustomOcc)%data) = .true.
+      iSpecies = geo%species(iAtInRegion(iCustomOcc)%data(1))
+      if (any(geo%species(iAtInRegion(iCustomOcc)%data) /= iSpecies)) then
+        call detailedError(child, "All atoms in a ReferenceOccupation&
+            & declaration must have the same type.")
+      end if
+      do iShell = 1, orb%nShell(iSpecies)
+        write(shellname, "(A)") orbitalNames(orb%angShell(iShell, iSpecies) + 1)
+          call getChildValue(node, shellname, customOcc(iShell, iCustomOcc), &
```
bhourahine: Not compatible with the extended basis set structure in the code (chance of multiple s shells for example).

make.config
pecchia: Yes. The make.config slipped into the commit. I have to revert to original. I think it should go into a modified gitignore file.

prog/dftb+/prg_dftb/parser.F90
pecchia: Then I don't know the logic you have in mind to treat extra shells. The orbital names:

orbitalNames(orb%angShell(iShell, iSpecies) + 1)

shouldn't be different for different shells ? So for an extra s they can have a name like s2 (or s* d*, as I suggested). Otherwise how should I differentiate ? The way it was done in the svn code was also wrong in my opinion.

pecchia: Ok, all the point is to create a unique 'label' for shellname that can be searched for in the input.
I can write a subroutine that create that string with the following convention:
First non-repeated shells with normal names 's','p',... then as soon as an OrbitalName repeats it is assigned a name like s2, s3, d2, ...  
This is just for the string to match. So user can use normal 's' 'p' 'd' names in usual calculations. 

Side Question: How do you set MaxAngularMomentum in case of sps* ?

pecchia:
I notice a failure of test="analysis/graphene_localise" unrelated to this PR.
I added a simple test with an H2 molecule with 2 s-shells. The second is the same as the first. 
H and S look fine but there is a problem in the factorization of S, though it is not singular

pecchia:
Fixed S that was not positive definite. 
Last issue: Balint did not like the long name CustomReferenceCharges. Can be changed to CustomRefCharges.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -5112,6 +5113,111 @@ end subroutine finalizeNegf
...
+        case(1)
+          sindx = ""
+        case(2:9)
+          write(sindx,'(I1)') ind(ii)   
+        case(10:)
+          call error("Exceeded maximum number of shells with the same angular momentum (9)")     
+        end select 
+        shellnames(ii) = trim(adjustl(shellnames(ii)))//sindx
+      end if
+      print*,shellnames(ii)
```
bhourahine: Should this print be here? Probably initprogram should be printing this information as part of the custom occupation block, also StdOut unit instead of *.

PR #237, opened 2019-04-16
Transport tool

pecchia:
Transport tool setupgeom with documentation in doc (not a dedicated test yet). 
- Fixed compilation both serial and parallel negf including poisson 
- Few fixes in poisson (including check of atoms in box)


bhourahine:
Changes all present in #263 so retiring this PR.

PR #252, opened 2019-05-07
Import

thomas-niehaus:
Modified https://github.com/aradi/dftbplus/tree/range-sep to incorporate spin-polarized LC-DFTB forces. Added test for this.


bhourahine:
New features adsorbed by  https://github.com/aradi/dftbplus/tree/range-sep

PR #254, opened 2019-05-10
Disable ELSI solvers with multiple threads

aradi:


doc/dftb+/manual/dftbp.tex
```diff
@@ -1771,6 +1771,12 @@ \subsection{Eigensolver}
 the additional \kw{ELPA}, \kw{OMM}, \kw{PEXSI} and \kw{NTPoly} solvers also
 become available.
 
+Note: The ELSI-solvers are not tested (or supposed to work) with multiple
```
bhourahine: Remove (or supposed to work)

bhourahine:
One comment only (see in place), but otherwise ready to merge.

doc/dftb+/manual/dftbp.tex
aradi: Done. :wink:

PR #255, opened 2019-05-10
Make autotest work even if time command not available

aradi:


test/prog/dftb+/bin/autotest2
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
bhourahine: Possible not a good idea to change, debian is in the process of depreciating bash for dash, with /bin/sh being a symlink to the latter now, so bash will probably disappear from future install defaults.

aradi: Good point, I made it sh compatible again, see dc73e0c

PR #256, opened 2019-05-12
Print hostname on standard output

aradi:


prog/dftb+/lib_utils/osutils.F90
```diff
@@ -0,0 +1,47 @@
...
+
+contains
+
+  !> Returns the host name.
+  function getHostName() result(hostName)
+
+    !> Name of the host, or the empty string, if could not be determined.
+    character(:), allocatable :: hostName
+
+    integer(c_size_t), parameter :: maxHostNameLength = 256
```
bhourahine: Looking at the man page HOST_NAME_MAX is probably better for the length

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -2550,6 +2551,7 @@ subroutine initProgramVariables(input, env)
       end do
     end if
 
+    write(stdOut, "('Host name:',T30,A)") getHostName()
```
bhourahine: Does this need a trim statement?

bhourahine:
Also, possibly not very cross platform:
https://stackoverflow.com/questions/27914311/get-computer-name-and-logged-user-name
so might be worth either generalising, or allowing a preprocessor macro to disable.

PR #258, opened 2019-05-15
Halogen X correction

bhourahine:
Honza has checked against the paper data and is happy.


bhourahine:
Bump

aradi:
I made a PR to this branch with some minor suggestions, please have a look at it, and merge it if you agree (should work with fast-forward).

Additionally, I'd suggest to rename the keyword to something like "HalogenXCorr" or so, instead of "DFTB3X", but that is open for discussion.

bhourahine:
PR absorbed.
Regarding the proposed name change, the correction only works with DFTB3 (and is unlikely to be extended), so not sure which is clearer but will change over.

PR #259, opened 2019-05-15
Fix deallocation bug

aradi:
All allocatables in initprogram are explicitely deallocated, so
that DFTB+ can be reinitialised after its finalisation again.


test/api/mm/testers/test_fileinitc.c
```diff
@@ -40,37 +40,40 @@ int main()
...
+    dftbp_set_coords_and_lattice_vecs(&calculator, coords, latvecs);
+
+    /* evaluate energy */
+    dftbp_get_energy(&calculator, &mermin_energy);
+    printf("Expected Mermin-energy: %15.10f\n", -2.5933460731);
+
+    /* and gradients */
+    gradients = (double *) calloc(3 * NR_OF_ATOMS, sizeof(double));
+    dftbp_get_gradients(&calculator, gradients);
+    printf("Expected gradient of atom 1: %15.10f %15.10f %15.10f\n", -0.010321385989,
```
bhourahine: Probably better to test with different cases in the iteration of the loop, in case of any internal state surviving the dftb_final() call.

prog/dftb+/include/common.fypp
```diff
@@ -78,4 +78,12 @@
 #{if rank == 0}##{else}#(${":" + ",:" * (rank - 1)}$)#{endif}#
 #:enddef
 
+
+#! Macro for automated deallocations
+#:def DEALLOCATE(*args)
```
bhourahine: The macro name is a little misleading, how about SAFEALLOC or REALLOCATE?

ExpHP:
Appears to be working well enough for my purposes!  Note that my usage almost entirely consists of very similar structures (it is used in relaxation and phonon computations), so I can't attest to how how well it works when the instances are given very diverse structures. I second Ben's suggestion about the test cases!

prog/dftb+/include/common.fypp
aradi: The macro is a simple conditional deallocation and does not contain any allocation. But I have renamed it in 671f0dd to `SAFE_DEALLOC` to express the intention more precisely.

test/api/mm/testers/test_fileinitc.c
aradi: Good point. I have did that now in 978f8 and the code still seems to work without changes.

bhourahine:
OK, good to go, will merge shortly.

PR #262, opened 2019-05-18
Fixed conjugation bug

bhourahine:
Phase was conjugated incorrectly for one triangle of replica atoms
outside central cell in packing from CSR to block matrices.

### DIFF
--- a/prog/dftb+/lib_dftbplus/initprogram.F90
+++ b/prog/dftb+/lib_dftbplus/initprogram.F90
@@ -1710,6 +1710,14 @@ contains
     end if
     call TElectronicSolver_init(electronicSolver, input%ctrl%solver%iSolver, nBufferedCholesky)
 
+    if (any(electronicSolver%iSolver ==&
+        & [electronicSolverTypes%elpa, electronicSolverTypes%omm, electronicSolverTypes%pexsi,&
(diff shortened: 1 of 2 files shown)


prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -1710,6 +1710,14 @@ subroutine initProgramVariables(input, env)
     end if
     call TElectronicSolver_init(electronicSolver, input%ctrl%solver%iSolver, nBufferedCholesky)
 
+    if (any(electronicSolver%iSolver ==&
+        & [electronicSolverTypes%elpa, electronicSolverTypes%omm, electronicSolverTypes%pexsi,&
+        & electronicSolverTypes%ntpoly]) .and.&
+        & input%ctrl%parallelOpts%nGroup /= nIndepHam * nKPoint) then
+
```
aradi: Isn't the first part of the condition equivalent to `electronicSolver%isElsiSolver`? That would be more compact and future safe.

bhourahine: Not quite, as the ELPA solver doesn't fall into this category (but shouldn't get this far as its not used in sparse mode).

PR #263, opened 2019-05-22
Bug fixes

pecchia:
Fixes of local currents with PBC. 
Relevant to the rest of the code: I modified periodic.F90 in order to obtain a symmetric neighbour list, quite useful for a simple coding of loops for local currents.


prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -1813,8 +1816,8 @@ subroutine readDFTBHam(node, ctrl, geo, slako)
       ! fixEf also avoids checks of total charge in initQFromFile
       ctrl%tFixEf = .true.
       if (geo%tPeriodic .and. greendens%doLocalCurr) then
```
bhourahine: Remove the if block as contents are commented out

prog/dftb+/lib_dftb/periodic.F90
```diff
@@ -512,12 +528,12 @@ function getNrOfNeighbours(neigh, cutoff, iAtom) result(nNeighbourSK)
     integer, intent(in) :: iAtom
 
     !> Nr. of neighbours for the specified atom.
-    integer :: nNeighbourSK
+    integer :: nNeighbour
```
bhourahine: Is this SK cut-off neighbours? Might be better to be consistent with naming elsewhere.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -3158,7 +3161,7 @@ subroutine readDispSlaKirk(node, geo, input)
       nNeighs(:) = 0
       do iAt1 = 1, geo%nAtom
         iSp1 = geo%species(iAt1)
-        do iNeigh = 1, neighs%nNeighbourSK(iAt1)
+        do iNeigh = 1, neighs%nNeighbour(iAt1)
```
bhourahine: nNeighbourSK again?

pecchia:
* Resolved conflict in parser.F90
* I now added that setupgeom write contact Ids into transport.hsd. This makes things more clear.

prog/setupgeom/parser_setup.F90
```diff
@@ -0,0 +1,641 @@
...
+        call detailedError(pTmp, "Contact id '" // trim(contacts(ii)%name) &
+            &//  "' already in use")
+      end if
+
+      call getChildValue(pNode, "PLShiftTolerance", contactLayerTol, 1e-5_dp, modifier=modif,&
+          & child=field)
+      call convertByMul(char(modif), lengthUnits, field, contactLayerTol)
+
+      if (task .eq. "setupgeometry") then
+        call getChildValue(pNode, "NumPLsDefined", nPLs(ii), 2)
```
bhourahine: Remove the ",2" default, so that this line becomes obligatory in the input file (instead of assuming 2 PLs, require the user to think about this).

PR #266, opened 2019-05-22
Dense ELSI LS

bhourahine:
Fixes a few issues for spin orbit with ELSI, the main being to disable for the moment the two solvers that give incorrect energy and forces with spin orbit.

### DIFF
--- a/prog/dftb+/lib_elecsolvers/elsisolver.F90
+++ b/prog/dftb+/lib_elecsolvers/elsisolver.F90
@@ -841,6 +841,18 @@ contains
       call mpifx_allreduceip(env%mpi%globalComm, iRhoPrim, MPI_SUM)
     end if
 
+    if (tSpinOrbit) then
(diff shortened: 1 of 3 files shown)


prog/dftb+/lib_elecsolvers/elsisolver.F90
```diff
@@ -345,7 +345,7 @@ subroutine TElsiSolver_init(this, inp, env, nBasisFn, nEl, iDistribFn, nSpin, iS
       this%spinDegeneracy = 0.0_dp
     end if
 
-    this%nSpin = nSpin
+    this%nSpin = min(nSpin, 2) ! copes with 2 component spin
```
aradi: Why min?

bhourahine: min(x,2) as occupation in ELSI is 2 for nSpin = 1 and 1 for nSpin = 2 (don't think it knows about nSpin = 4, but partly tests for > 1 in some places).

PR #289, opened 2019-06-20
Fix problems on PDOS regions in transport

pecchia:
Now there is always a default Region with all device atoms
I also added a check that atoms defined are within the device
ISSUE:
Now selecting atoms by specie it can happen they are found outside the device region and the code stop.
Maybe we should add a range bound to convAtomRangeToInt. 
This is currently set 1:nAtom  where nAtom=size(geo%specie).
Once the array of indices is returned it is lengthy to remove indices outside the device.


bhourahine:
See #290

PR #296, opened 2019-07-05
Quick hsd comment

pecchia:
1) I have added the possibility to comment out a whole block with a simple "%". Example:
%kpointsAndWeights = klines{
    ...
    ...
}

2) In doPostParseJobs I have added the definition of a list "ignoreList".  Now it is possible to
call warnUnprocessedNodes(root, ignoreList, nodeList)
Warnings and errors occurs only for unprocessed nodes and parentNodes that are not in ignoreList.
As an experiment I set "MaxAngularMomentum" in this list so redundant species can be ignored.

### DIFF
--- a/prog/dftb+/lib_dftbplus/hsdhelpers.F90
+++ b/prog/dftb+/lib_dftbplus/hsdhelpers.F90
@@ -16,7 +16,8 @@ module dftbp_hsdhelpers
   use dftbp_hsdutils2
   use dftbp_inputdata_module
   use dftbp_message
-  use dftbp_xmlf90, only : fnode, destroyNode
+  use dftbp_xmlf90, only : fnode, createElement, fnodeList, append, &
+                     &      destroyNode, getLength
   implicit none
   private
 
@@ -72,12 +73,26 @@ contains
     !> Parser specific settings in the output
     type(TParserFlags), intent(in) :: parserFlags
 
-    type(fnode), pointer :: root
+    type(fnode), pointer :: root, node
+    type(fnodeList), pointer :: ignoreList, nodeList
 
     call getChild(hsdTree, rootTag, root)
 
+    ! Sets a list of nodes that will be ignored even if not processed
+    ! Logic: warnUnprocessed will check each node and its parent with ignoreList 
+    !        Hence setting "MaxAngularMomentum" will allow ignoring all 
+    !        its children elements  
+    nullify(ignoreList)
+    node => createElement("")
+    call setNodeName(node,"MaxAngularMomentum")
+    call append(ignoreList, node)
+
     ! Issue warning about unprocessed nodes
-    call warnUnprocessedNodes(root, parserFlags%tIgnoreUnprocessed)
+    call warnUnprocessedNodes(root, ignoreList, nodeList)
+
+    if (.not.parserFlags%tIgnoreUnprocessed .and. getLength(nodeList)>0) then
+      call errorUnprocessedNodes()
+    end if    
 
     ! Dump processed tree in HSD and XML format
     if (tIoProc .and. parserFlags%tWriteHSD) then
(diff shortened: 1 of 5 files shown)


prog/dftb+/lib_dftbplus/hsdhelpers.F90
```diff
@@ -16,7 +16,8 @@ module dftbp_hsdhelpers
   use dftbp_hsdutils2
   use dftbp_inputdata_module
   use dftbp_message
-  use dftbp_xmlf90, only : fnode, destroyNode
+  use dftbp_xmlf90, only : fnode, createElement, fnodeList, append, &
+                     &      destroyNode, getLength
```
aradi: Please indent continuation line with 4 spaces only.

prog/dftb+/lib_dftbplus/hsdhelpers.F90
```diff
@@ -72,12 +73,26 @@ subroutine doPostParseJobs(hsdTree, parserFlags)
...
 
     call getChild(hsdTree, rootTag, root)
 
+    ! Sets a list of nodes that will be ignored even if not processed
+    ! Logic: warnUnprocessed will check each node and its parent with ignoreList 
+    !        Hence setting "MaxAngularMomentum" will allow ignoring all 
+    !        its children elements  
+    nullify(ignoreList)
+    node => createElement("")
+    call setNodeName(node,"MaxAngularMomentum")
```
aradi: I think, the nodes, which should not trigger an unprocessed warning, should be marked during the parsing in `parser.F90` by setting a specific hsd-attribute (e.g. ``hsdIgnoreUnproc``), rather than here. Setting them here would mix the generic parsing functions and specific DFTB+ related parsing issues. This routine is used by several hsd-parser (e.g. modes, waveplot, etc.) where a ``MaxAngularMomentum`` node does not appear. Also nodes added here won't have the proper path information (what, if there are nodes with the same name on different levels, and only one of them should be ignored...).

prog/dftb+/lib_io/hsdparser.F90
```diff
@@ -616,6 +618,9 @@ function createChildNode(parentNode, childName, curLine, file) result(newChild)
           dummy => removeChild(parentNode, sameChild)
           call destroyNode(sameChild)
           tCreate = .true.
+        elseif (iType == 6) then
+          newChild => null()
+          return
         else
```
aradi: Case 6 should be handled separately. There is now reason to look up for other nodes with the same name, if you find a comment sign in front of a node. What about:
```
if (iType == 0) then
...
else if (iType == 6) then
  newChild => null()
  return
else
...
end if
```

prog/dftb+/lib_io/hsdutils2.F90
```diff
@@ -121,54 +122,80 @@ end subroutine getUnprocessedNodes
...
+          call getNodeName(childI, nameI)
+          ! check by name both node and parent (for block ignores)
+          if (char(name1) /= char(nameI) .and. char(name2) /= char(nameI)) then
+            call append(list, child)    
+            call append_to_string(msg, "(" // i2c(ii+1) // ")")
+            call appendPathAndLine(child, msg)
+            call append_to_string(msg, newline)
+          end if  
+        end do  
       end do
```
aradi: We should just simply check, whether ``child`` has a given attribute set, and only include it in the list, it the attribute ``hsdIgnoreUnproc`` is not present or set to ``0``.

prog/dftb+/lib_io/xmlutils.F90
```diff
@@ -221,6 +222,74 @@ recursive subroutine getTagsWithoutAttr_recursive(node, name, rootOnly, &
 
   end subroutine getTagsWithoutAttr_recursive
```
aradi: The routine is OK, but as far as I can see, it is not used anywhere. Could we wait adding it, until it is really needed?

PR #297, opened 2019-07-18
Python 3 version of the dependacy util

bhourahine:
Python3 version of the cr_makedep script


aradi:
All fine, but then let's drop cr_makedep2 completely. All developers should have python3 working, and users just building the code don't use cr_makedep at all.

PR #300, opened 2019-08-13
CMake build framework

aradi:
Contains a fully functional build framework using CMake. The old legacy build framework still works. GPU-build has not been tested yet.


aradi:
Made a few more updates. The CMake base build system is hopefully robust now and allows easy integration of the library into other CMake based projects.

ExpHP:
Some very preliminary comments:

* It only builds one library (no `libxmlf90`), but it looks like this library contains symbols that used to be defined in `libxmlf90`, so hopefully it works.
* I think the recommendation to the user should not be to modify `.cmake` files that are checked into VCS, but rather to supply `-D` flags.
* Overall, the idea of platform-specific files in `/sys` sounds counter to the goals of cmake, and I think anything in there needs pretty strong justification.
   * It looks like much of it is to select the right flags based on compiler. Are you aware of `CMAKE_COMPILER_IS_GNUC` and `CMAKE_LANG_COMPILER_ID`, etc.? See [Chuck Atkin's message here](https://cmake.org/pipermail/cmake/2016-May/063449.html) for what looks like a fairly sensible solution using these flags.
    * There's also stuff to find dependencies.  Does `find_package` or similar not suffice for these?

aradi:
Thanks a lot for looking at it.
>     * It only builds one library (no `libxmlf90`), but it looks like this library contains symbols that used to be defined in `libxmlf90`, so hopefully it works.

Yes, libxmlf90 (and some other libraries as well) are now part of libdftbplus.a, so no extra linking is needed.

>     * I think the recommendation to the user should not be to modify `.cmake` files that are checked into VCS, but rather to supply `-D` flags.

I find it annoying, to specify 8-9 `-D` options on the command line, and I assumed, simple users won't like it either. As long as they do not submit PRs, it does not really matter, whether a file they change is under version control. So, let's them do it the more comfortable way. (For developers, there is an extra note in `config.cmake`...)

>     * Overall, the idea of platform-specific files in `/sys` sounds counter to the goals of cmake, and I think anything in there needs pretty strong justification.

Yes, there you are absolutely right. However, my experience with CMake, especially on supercomputers with various custom environments was, that it almost always picks the wrong compiler/library combination. (For example, it would take by default `/usr/bin/f95` as Fortran compiler, despite the fact, that you have a recent `gfortran` in your `PATH`. And then, it happily combines it with an MPI it finds somewhere, which was compiled with a completely different compiler, etc.) Therefore, I find it safer when users set those settings explicitly to ensure consistency and ease debugging.

>       
>       * It looks like much of it is to select the right flags based on compiler. Are you aware of `CMAKE_COMPILER_IS_GNUC` and `CMAKE_LANG_COMPILER_ID`, etc.? See [Chuck Atkin's message here](https://cmake.org/pipermail/cmake/2016-May/063449.html) for what looks like a fairly sensible solution using these flags.

That's true. We could do it within the depth of our CMakeLists.txt files somewhere, indeed. But then, it would be absolutely intransparent to the user, which flags we have used. (Typical example: you have to use for the intel compiler either `-standard-semantics` or `-assume realloc_lhs`, depending with which flag the libraries (e.g. ELSI) you link, has been compiled. There is no simple way, to detect this automatically, so the user would have to take care about it, when building the code. Then, it is better IMHO to have all the flags transparently connected into one file. Also, not to mention compilers, I have no access to (e.g. xlf on IBM machines). The user would have to set up the flags himself, and then, it helps if you see, what flags are usually used.

>       * There's also stuff to find dependencies.  Does `find_package` or similar not suffice for these?

Yes, indeed. But for most of those packages there is not Find* module in CMake. And even the one, which is there are not necessarily useful. (E.g. FindLAPACK: How do you know, whether it found the one with threaded or with serial MKL? Depending whether you compile with or without MPI support, you may want to have a different one. Again, we could try to find it out afterward, but it is probably more transparent to set it up in an architecture dependant config file.)

The point is, clueless users on standard Linux machines would be able to compile the code without touching any settings, since the default is Linux/GNU with settings working on Ubuntu. For anybody else, who needs to change configurations, it is probably more convenient, if they can do that in a transparent way. (Despite, that it is somewhat against the usual CMake philosophy...)

ExpHP:
Thank you for considering my comments.

> For example, it would take by default /usr/bin/f95 as Fortran compiler, despite the fact, that you have a recent gfortran in your PATH. And then, it happily combines it with an MPI it finds somewhere, which was compiled with a completely different compiler, etc.

I am tempted to call this a system configuration issue, but I can understand the sentiment that many HPC environments may be poorly configured, and that an issue like this could be very time-consuming to debug for an inexperienced user.

If anything, it would be nice if the configuration could still be overriden by the user environment variables that are normally honored by cmake, such as `FC`. I am trying to use DTFB+ from Rust, and the standard practice in that ecosystem is to locate a system-installed package if available, or else try to automatically build the dependency from source (for easy deployment to CI testing and etc.).  In this setting, the end user has no opportunity to modify `cmake` files.

>  (Typical example: you have to use for the intel compiler either -standard-semantics or -assume realloc_lhs, depending with which flag the libraries (e.g. ELSI) you link, has been compiled. There is no simple way, to detect this automatically, so the user would have to take care about it, when building the code.

Interesting.  I know that, in the case of `pkg-config` for C code, such flags can be listed under `Cflags:`, where dependent code can gather them with `--cflags`.  I don't know enough about `cmake` to know for certain that there is anything similar, but I would imagine there must be.

Of course, it would require cooperation from ELSI to ensure that their installed `cmake` configuration communicates these flags.

aradi:
> If anything, it would be nice if the configuration could still be overriden by the user environment variables that are normally honored by cmake, such as `FC`. I am trying to use DTFB+ from Rust, and the standard practice in that ecosystem is to locate a system-installed package if available, or else try to automatically build the dependency from source (for easy deployment to CI testing and etc.). In this setting, the end user has no opportunity to modify `cmake` files during an automatic build.

I see your point, and for simple builds (libdftbplus.a without any optional external dependencies using one of the common compiler suites) it may even work for. I have to see, whether I can implement it by still keeping the following two goals:
* it should be ease to see (in advance), which compiler flags the build system will use
* it should be easy and straightforward to fine tune and override compiler, library and various flag choices for the experts.

aradi:
OK, the PR is really ready now. Compiler tool-chain selection happens via ``-DCMAKE_TOOLCHAIN_FILE`` option. It still not automatic, but the only more or less safe and controllable bet for the moment.

sys/gnu.cmake
```diff
@@ -0,0 +1,79 @@
...
+
+#
+# External libraries
+#
+
+# LAPACK and BLAS
+if(WITH_MPI)
+  set(LAPACK_LIBRARIES "lapack blas" CACHE STRING "LAPACK and BLAS libraries to link")
+else()
+  set(LAPACK_LIBRARIES "openblas" CACHE STRING "LAPACK and BLAS libraries to link")
```
bhourahine: probably should be "blas lapack", firstly as lapack won't always be picked up and secondly as openblas still has that extra thread(s) issue at runtime. Also as mkl is now available, would some comment about using that be a good idea?

aradi: IMO, if you link statically, blas should come after lapack in the linking order to enable the linker to resolve the blas-calls in lapack (or we would need a --stat-group --end-group construct.) Currently we always link lapack and let the linker throw it out, if not needed. (True for the make-build as well.)

Yes, probably we should mention the possibility of linking MKL, indeed. I'll add such a line.

bhourahine: Just changing it to just "blas" here doesn't pick up the lapack symbols for me with cmake 3.10.2, but agree with the point about "lapack blas" (wrong way around in comment)/

PR #301, opened 2019-08-21
Python3

bhourahine:
Python3 conversion on other scripts


ExpHP:
Just pointing out that it looks like not all scripts have actually been ported to python3-compatible syntax.

(e.g. [dftbplus/tools/misc/convolve.py](https://github.com/dftbplus/dftbplus/blob/4bafae122d9356e9ae441ed53b5695949e644556/tools/misc/convolve.py))

ExpHP:
Here's a couple things that can be said in the way of non-obvious errors (i.e. things that still parse, but have changed to do something different):

```sh
$ rg '(keys|values|items)\(\)' $(cat pyfiles)
utils/build/cr_makedep/cr_makedep2
164:    newDeps = deps + self._dependencies.keys()
165:    newDfns = dfns + self._definitions.keys()

test/prog/dftb+/bin/tagreader.py
168:    __validTypes = __strToValue.keys()
```

These three calls to `keys()` should be wrapped in a call to `list(...)`.  (`keys()` now returns something that is not a `list`)

```sh
$ rg 'unicode' $(cat pyfiles)
test/prog/dftb+/bin/uncommlines.py
29:            TypeError -- if strSeparator is not string or unicode instance.
33:                            or isinstance(strSeparator, unicode)):
```

`unicode` was renamed to `str`, so this will produce a NameError instead of the TypeError it intends to.

---

I also looked for instances of

* integer division
* calls to `map`, `filter`, or `zip` that should be wrapped in `list`
* usage of `str` that should be `bytes`

but didn't find anything.

PR #310, opened 2019-10-27
Fix block population bug

inseonglee:
1. Fix the bug that the block population is not updated during SCC iteration when using LC and onsite correction simultaneously.
Current modified routine uses the 'gemm' routine in dense form.

2. Add the dependency "_mod_dftbp_halogenx" to parser.F90 in make.deps.
For this, I do not know the compatibility between halogenx and sockets/transport.
So, I just add the module to all parts (regardless of sockets or transport) in make.deps.
You may check this and change according to the compatibility.


prog/dftb+/lib_dftb/populations.F90
```diff
@@ -451,5 +452,85 @@ subroutine getChargePerShell(qOrb, orb, species, chargePerShell, qRef)
   end subroutine getChargePerShell
```
bhourahine: Could you add a !> comment on the top of the routine?

prog/dftb+/lib_dftb/populations.F90
```diff
@@ -451,5 +452,85 @@ subroutine getChargePerShell(qOrb, orb, species, chargePerShell, qRef)
...
+        tmpD(ii,ii,iS) = tmpD(ii,ii,iS) - rhoSqr(ii,ii,iS)
+      end do
+    end do
+
+    qq(:,:,:,:) = 0.0_dp
+    do iS = 1, nSpin
+
+      tmp_block(:,:) = 0.0_dp
+      tmp_mat(:,:) = 0.0_dp
+      call gemm(tmp_mat,tmpS,tmpD(:,:,iS),1.0_dp,0.0_dp,'N','N')
```
bhourahine: You can rely on defaults for these optional arguments (alpha=1,beta=0,transA=N,transB=N). If changing an optional, its usually clearer to also set by value instead of position.

prog/dftb+/lib_dftb/populations.F90
```diff
@@ -451,5 +452,85 @@ subroutine getChargePerShell(qOrb, orb, species, chargePerShell, qRef)
...
+    nAOs = size(rhoSqr, dim=1)
+
+    ALLOCATE(tmpS(nAOs,nAOs))
+    ALLOCATE(tmpD(nAOs,nAOs,nSpin))
+    ALLOCATE(tmp_mat(nAOs,nAOs))
+    ALLOCATE(tmp_block(nAOs,nAOs))
+
+    ! ... symmetrize overlap and density matrices
+    tmpS(:,:) = 0.0_dp
+    tmpS(:,:) = overSqr(:,:) + transpose(overSqr(:,:))
```
bhourahine: On the right hand side of assignments, specifying (:,:) might lead to some compilers making a copy of this variable. You cannot avoid it for array slicing, but otherwise
tmpS(:,:) = overSqr + transpose(overSqr)
is better.

prog/dftb+/lib_dftb/populations.F90
```diff
@@ -451,5 +452,85 @@ subroutine getChargePerShell(qOrb, orb, species, chargePerShell, qRef)
...
+    real(dp), allocatable :: tmp_mat(:,:)
+    real(dp), allocatable :: tmp_block(:,:)
+
+    integer :: nAOs, nat, nSpin, ii, jj, kk, iat, iS
+
+    nSpin = size(rhoSqr, dim=3)
+    nat = size(iSquare, dim=1) - 1
+    nAOs = size(rhoSqr, dim=1)
+
+    ALLOCATE(tmpS(nAOs,nAOs))
```
bhourahine: You don't need to capitalize ALLOCATE and DEALLOCATE (we usually keep language keywords in lower case).

prog/dftb+/lib_dftb/populations.F90
```diff
@@ -451,5 +452,85 @@ subroutine getChargePerShell(qOrb, orb, species, chargePerShell, qRef)
...
+      tmpD(:,:,iS) = 0.0_dp
+      tmpD(:,:,iS) = rhoSqr(:,:,iS) + transpose(rhoSqr(:,:,iS))
+      do ii = 1, nAOs
+        tmpD(ii,ii,iS) = tmpD(ii,ii,iS) - rhoSqr(ii,ii,iS)
+      end do
+    end do
+
+    qq(:,:,:,:) = 0.0_dp
+    do iS = 1, nSpin
+
```
bhourahine: Instead of two GEMM operations over all of the matrices, this can also be done as a series of small GEMM operations for each atom block:
ii = iSquare(iat)
jj = iSquare(iat+1)
nOrb = jj - ii
qBlock(:nOrb,:nOrb,iAt,iSpin) = matmul(Ssqr(ii:jj-1,:,iSpin), rho(:,ii:jj-1,iSpin))

then

qBlock(:nOrb,:nOrb,iAt,iSpin) = 0.5_dp * (qBlock(:nOrb,:nOrb,iAt,iSpin)&
& + transpose(qBlock(:nOrb,:nOrb,iAt,iSpin)))

Without measuring it, not sure for how large a system this becomes faster ( O(N) operations, but less effective cache use) than GEMM over the whole matrix.

prog/dftb+/lib_dftb/populations.F90
```diff
@@ -451,5 +452,86 @@ subroutine getChargePerShell(qOrb, orb, species, chargePerShell, qRef)
...
+      tmp_block(:,:) = 0.0_dp
+      tmp_mat(:,:) = 0.0_dp
+      call gemm(tmp_mat,tmpS,tmpD(:,:,iS),alpha=1.0_dp,beta=0.0_dp,transA='N',transB='N')
+      tmp_block(:,:) = tmp_block(:,:) + 0.5_dp * tmp_mat(:,:)
+
+      tmp_mat(:,:) = 0.0_dp
+      call gemm(tmp_mat,tmpD(:,:,iS),tmpS,alpha=1.0_dp,beta=0.0_dp,transA='N',transB='N')
+      tmp_block(:,:) = tmp_block(:,:) + 0.5_dp * tmp_mat(:,:)
+
+      do iat = 1, nat
```
bhourahine: Should this be done in the population evaluation routine? This is the gross electron count not the charge on the orbitals.

inseonglee:
Should I write the change log in the feature branch?

bhourahine:
Yes, please add under fixed heading. Also add yourself to the AUTHORS.rst file.

prog/dftb+/lib_dftb/populations.F90
```diff
@@ -451,5 +451,64 @@ subroutine getChargePerShell(qOrb, orb, species, chargePerShell, qRef)
...
+    !> Mulliken block charges on output (mOrb, mOrb, nAtom, nSpin)
+    real(dp), intent(out) :: qq(:,:,:,:)
+
+    real(dp), allocatable :: tmpS(:,:)
+    real(dp), allocatable :: tmpD(:,:)
+
+    integer :: nAOs, nat, nSpin, ii, jj, kk, iat, iS, nOrb
+
+    nSpin = size(rhoSqr, dim=3)
+    nat = size(iSquare, dim=1) - 1
```
aradi: Use camelCase `nAt`

prog/dftb+/lib_dftb/populations.F90
```diff
@@ -451,5 +451,64 @@ subroutine getChargePerShell(qOrb, orb, species, chargePerShell, qRef)
...
+
+    do iS = 1, nSpin
+
+      ! ... symmetrize density matrix for spin channel
+      tmpD(:,:) = rhoSqr(:,:,iS) + transpose(rhoSqr(:,:,iS))
+      do ii = 1, nAOs
+        tmpD(ii,ii) = rhoSqr(ii,ii,iS)
+      end do
+
+      do iat = 1, nat
```
aradi: `iat` -> `iAt`, `nat` -> `nAt` (or even better `nAtom`)

prog/dftb+/lib_dftb/populations.F90
```diff
@@ -451,5 +451,64 @@ subroutine getChargePerShell(qOrb, orb, species, chargePerShell, qRef)
...
+        nOrb = jj - ii
+        qq(:nOrb,:nOrb,iAt,iS) = matmul(tmpS(ii:jj-1,:), tmpD(:,ii:jj-1))
+        qq(:nOrb,:nOrb,iAt,iS) = 0.5_dp * (qq(:nOrb,:nOrb,iAt,iS)&
+            & + transpose(qq(:nOrb,:nOrb,iAt,iS)))
+      end do
+
+    end do
+
+    deallocate(tmpS)
+    deallocate(tmpD)
```
aradi: Remove deallocations (allocatables are automatically deallocated if going out of scope)

prog/dftb+/lib_dftb/populations.F90
```diff
@@ -451,5 +451,64 @@ subroutine getChargePerShell(qOrb, orb, species, chargePerShell, qRef)
...
+    integer :: nAOs, nat, nSpin, ii, jj, kk, iat, iS, nOrb
+
+    nSpin = size(rhoSqr, dim=3)
+    nat = size(iSquare, dim=1) - 1
+    nAOs = size(rhoSqr, dim=1)
+
+    allocate(tmpS(nAOs,nAOs))
+    allocate(tmpD(nAOs,nAOs))
+
+    ! ... symmetrize overlap
```
aradi: Please remove the `...` from the comments, as their are neither informative nor improve readability

PR #313, opened 2019-11-02
Add matrix based long-range Hamiltonian

inseonglee:
Add new algorithm using direct matrix multiplication (gemm routine).
It shows much faster computational speed than other two algorithms.
Also, change 'tSpin' to 'nSpin' in rangeSep variable.
Currently, I do not add test cases for this method.

In addLrGradients routine of rangeseparated.F90 file, I wonder that the following sentence is essential parts or not.
"precomputing the lr-gamma derivatives"
There is no reason to print in standard output.

Please tell me if the wrong part exists.


prog/dftb+/lib_dftb/rangeseparated.F90
```diff
@@ -204,12 +208,12 @@ subroutine checkRequirements(this)
       class(RangeSepFunc), intent(inout) :: this
 
       ! Check for current restrictions
-      if (this%tSpin .and. this%rsAlg == rangeSepTypes%threshold) then
+      if (this%nSpin == 2 .and. this%rsAlg == rangeSepTypes%threshold) then
```
aradi: Ok, but only if comment above has been taken into account!

prog/dftb+/lib_dftb/rangeseparated.F90
```diff
@@ -177,15 +181,15 @@ subroutine initAndAllocate(this, nAtom, hubbu, species, screen, omega, rsAlg, tS
...
+      !> Is this spin restricted (1) or unrestricted (2)
+      integer, intent(in) :: nSpin
 
       this%tScreeningInited = .false.
       this%pScreeningThreshold = screen
       this%omega = omega
       this%lrEnergy = 0.0_dp
       this%rsAlg = rsAlg
-      this%tSpin = tSpin
+      this%nSpin = nSpin
```
aradi: Dangerous! `nSpin` can be also `4` (non-collinear spin), which the routines here can not handle. You can keep `this%nSpin`, but please set it up using `tSpin` (and not `nSpin) accordingly. `tSpin` is only `.true.` for collinear spin and `.false.` otherwise (no spin polarisation, non-collinear spin).

prog/dftb+/lib_dftb/rangeseparated.F90
```diff
@@ -693,16 +699,181 @@ subroutine updateHamiltonianBlock(descM, descN, pSma, pSbN, pPab)
...
+    real(dp), intent(inout) :: HH(:,:)
+
+    real(dp), allocatable :: Smat(:,:)
+    real(dp), allocatable :: Dmat(:,:)
+    real(dp), allocatable :: LRgammaAO(:,:)
+    real(dp), allocatable :: Hlr(:,:)
+
+    integer :: nAOs
+
+    nAOs = size(overlap,dim=1)
```
aradi: Again, please `nOrb` instead of `nAOs` to be consistent with other routines.

prog/dftb+/lib_dftb/rangeseparated.F90
```diff
@@ -693,16 +699,181 @@ subroutine updateHamiltonianBlock(descM, descN, pSma, pSbN, pPab)
...
+
+    allocate(Smat(nAOs,nAOs))
+    allocate(Dmat(nAOs,nAOs))
+    allocate(LRgammaAO(nAOs,nAOs))
+    allocate(Hlr(nAOs,nAOs))
+
+    call allocateAndInit(this, iSquare, orb, overlap, densSqr, HH, Smat, Dmat, LRgammaAO)
+    call evaluateHamiltonian(this, Smat, Dmat, LRgammaAO, Hlr)
+    HH(:,:) = HH(:,:) + Hlr(:,:)
```
aradi: Remove `(:,:)` from the right hand side! (Also on several places further down...)

prog/dftb+/lib_dftb/rangeseparated.F90
```diff
@@ -693,16 +699,181 @@ subroutine updateHamiltonianBlock(descM, descN, pSma, pSbN, pPab)
...
+
+      nAOs = size(overlap,dim=1)
+
+      allocate(Hmat(nAOs,nAOs))
+      allocate(tmp_mat(nAOs,nAOs))
+
+      Hlr(:,:) = 0.0_dp
+
+      tmp_mat(:,:) = 0.0_dp
+      call gemm(tmp_mat,Smat,Dmat,1.0_dp,0.0_dp,'N','N')
```
aradi: Omit defaults! `call gemm(tmp_mat, Smat, Dmat)`

prog/dftb+/lib_dftb/rangeseparated.F90
```diff
@@ -693,16 +699,181 @@ subroutine updateHamiltonianBlock(descM, descN, pSma, pSbN, pPab)
...
+      type(RangeSepFunc), intent(inout) :: this
+
+      real(dp), intent(in) :: Smat(:,:)
+      real(dp), intent(in) :: Dmat(:,:)
+      real(dp), intent(in) :: LRgammaAO(:,:)
+
+      real(dp), intent(out) :: Hlr(:,:)
+
+      real(dp), allocatable :: Hmat(:,:)
+      real(dp), allocatable :: tmp_mat(:,:)
```
aradi: Use `camelCase` variables names without `_`

prog/dftb+/lib_dftb/rangeseparated.F90
```diff
@@ -693,16 +699,181 @@ subroutine updateHamiltonianBlock(descM, descN, pSma, pSbN, pPab)
...
+      tmp_mat(:,:) = 0.0_dp
+      call gemm(tmp_mat,Smat,Dmat,1.0_dp,0.0_dp,'N','N')
+      Hmat(:,:) = 0.0_dp
+      call gemm(Hmat,tmp_mat,Smat,1.0_dp,0.0_dp,'N','N')
+      do mu = 1, nAOs
+        do nu = mu, nAOs
+          Hmat(mu,nu) = Hmat(mu,nu) * LRgammaAO(mu,nu)
+          if (mu /= nu) Hmat(nu,mu) = Hmat(mu,nu)
+        end do
+      end do
```
aradi: Why not replacing the loop with `Hmat(:,:) = Hmat * LRgammaAO`? Costs a few multiplications more, but we can spare the `if`. You could do it at all 4 occasions below.

prog/dftb+/lib_dftb/rangeseparated.F90
```diff
@@ -1178,7 +1349,7 @@ subroutine allocateAndInit(tmpOvr, tmpRho, gammaPrimeTmp, tmpderiv)
         call symmetrizeSquareMatrix(tmpRho(:,:,iSpin))
       enddo
       ! precompute the gamma derivatives
-      write(stdOut,'(a)') "precomputing the lr-gamma derivatives"
+!      write(stdOut,'(a)') "precomputing the lr-gamma derivatives"
```
aradi: Just simply remove the line.

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -2285,7 +2285,7 @@ subroutine initProgramVariables(input, env)
       call ensureRangeSeparatedReqs(tPeriodic, tReadChrg, input%ctrl%tShellResolved,&
           & input%ctrl%rangeSepInp)
       call getRangeSeparatedCutoff(input%ctrl%rangeSepInp%cutoffRed, cutOff)
-      call initRangeSeparated(nAtom, species0, speciesName, hubbU, input%ctrl%rangeSepInp, tSpin,&
+      call initRangeSeparated(nAtom, species0, speciesName, hubbU, input%ctrl%rangeSepInp, nSpin,&
```
aradi: As discussed above, probably `tSpin` should be passed here.

inseonglee:
Thanks the comments!
After I modify the codes according to your comments, I will show new commit.

Overall, the reason why I change the variable 'tSpin' to 'nSpin' is due to the SSR parts which will be implemented after this.
In SSR case, it is based on spin-restricted open-shell scheme.
The spin polarisation (in this case, nSpin should be 2) is used to compute each microstate (in SSR(2,2) case, there are 6 microstates),
but this spin-restricted scheme (in this case, nSpin should be 1) is used to check the convergence in SCC loop.

Due to this point, the spin polarisation block is not used for SSR calculation in the dftb_in.hsd (in this case, tSpin is false), but the spin constants block is used. When I calculate the LC parts, the nSpin should be two since the spin contribution exists from 3rd to 6th microstates.

Thus, I use the nSpin variable as a standard to decide the spin-resolved calculation.
I do not know whether this explanation is enough to you.

If there exists better solution for this, please tell me about this.

bhourahine:
Is the intention to use other values of nSpin than {1,2} for the SSR part? Or instead could a local variable be set inside the module as part of the derived type using a different naming convention?
The matrix multiply form should be suitable for spinor wavefunctions, so nSpin={1,2,4} should be retained if possible.

inseonglee:
I modify the codes according to the comments.
For the nSpin variable, I change to the original 'tSpin' variable and add 'tREKS' variable in rangeSep module. Currently, it is always 'false' and will be changed to the input variables after adding SSR parts.

prog/dftb+/lib_dftb/rangeseparated.F90
```diff
@@ -703,6 +719,180 @@ end subroutine updateHamiltonianBlock
...
+      allocate(get_iat(nOrb))
+
+      ! Symmetrize Hamiltonian, overlap, density matrices
+      call symmetrizeSquareMatrix(HH)
+      Smat(:,:) = overlap
+      call symmetrizeSquareMatrix(Smat)
+      Dmat(:,:) = densSqr
+      call symmetrizeSquareMatrix(Dmat)
+
+      do mu = 1, nOrb
```
bhourahine: The rest of the routine can be simplified a bit as the atom resolved gamma is stored symmetrically and you can save the extra indexing array as well:

      ! Get long-range gamma variable
      LRgammaAO(:,:) = 0.0_dp
      do iAt = 1, nAtom
        do jAt = 1, nAtom
          LRgammaAO(iSquare(jAt):iSquare(jAt+1)-1,iSquare(iAt):iSquare(iAt+1)-1) =&
              & this%lrGammaEval(jAt,iAt)
        end do
      end do

prog/dftb+/lib_dftb/rangeseparated.F90
```diff
@@ -703,6 +719,180 @@ end subroutine updateHamiltonianBlock
...
+
+      integer :: nOrb
+
+      nOrb = size(Smat,dim=1)
+
+      allocate(Hmat(nOrb,nOrb))
+      allocate(tmpMat(nOrb,nOrb))
+
+      Hlr(:,:) = 0.0_dp
+
```
bhourahine: You can get the gemm operation to make the copying and assembly of Hlr which simplifies the code a little (but doesn't save much in run time or reduction of variables):

      call gemm(tmpMat, Smat, Dmat)
      call gemm(Hlr, tmpMat, Smat)
      Hlr(:,:) = Hlr * LRGammaAO

      tmpMat(:,:) = LRgammaAO * tmpMat
      call gemm(Hlr, tmpMat, Smat, alpha = 1.0_dp, beta = 1.0_dp)

      Hmat(:,:) = Dmat * LRgammaAO
      call gemm(tmpMat, Smat, Hmat)
      call gemm(Hlr, tmpMat, Smat, alpha = 1.0_dp, beta = 1.0_dp)

      call gemm(tmpMat, Dmat, Smat)
      tmpMat(:,:) = tmpMat * LRgammaAO
      call gemm(Hlr, Smat, tmpMat, alpha = 1.0_dp, beta = 1.0_dp)

inseonglee:
Sorry for late check of your pull request to my feature branch.
I checked and modified the gemm and gamma calculations.
Also, I added the test files for this matrix based method.

prog/dftb+/lib_dftb/energies.F90
```diff
@@ -33,6 +33,9 @@ module dftbp_energies
     !> spin energy
     real(dp) :: Espin   = 0.0_dp
 
+    !> range-separation energy
+    real(dp) :: Efock   = 0.0_dp
+
```
inseonglee: Here, I only add the energy variable 'Efock' to express range-separation energy.
In this case, current code do not include any atom-resolved LC energy.
Is it possible to calculate atom-resolved this LC energy?

bhourahine: For your algorithm this should be possible, but may need some care so that it agrees with other choices. 

For the evaluateLrEnergyDirect routine later in the code, this energy can be resolved as the coulombic energies of the exchange charges. Around line 1400 :

energy = energy + tmp * this%lrGammaEval(iAt1,iAt2)

replaced with

egyTmp = tmp * this%lrGammaEval(iAt1,iAt2)
energy = energy + egyTmp
Eatom(iAt1) = Eatom(iAt1) + 0.5_dp * egyTmp
Eatom(iAt2) = Eatom(iAt2) + 0.5_dp * egyTmp

I _think_ this is equivalent to a partial summation over the (Dmat * Hlr) that you use, but this would need to be tested for equivalence:

do iAt = 1, nAtom
  iStart = iSquare(iAt)
  iEnd = iSquare(iAt+1)-1
  Eatom(iAt1) = Eatom(iAt1) + sum(Dmat(:, iStart:iEnd) * Hlr(:, iStart:iEnd))
end do

bhourahine: Just numerically checked this. There is a slight difference between the partial summation method and the exchange change energies (~1%), so they are not equivalent.

inseonglee: In current step, it seems to be hard to compute atom-resolved LC energies in this algorithm. How about printing the warning sentence in stdout when AtomResolvedEnergies is 'Yes'?

aradi: Yes, I think we should print a warning and switch off `AtomResolvedEnergies` in this case.

inseonglee: Did you have a plan to implement to compute atom resolved LC energies for NeighbourBased or Thresholded cases? If not, then I will switch off AtomResolvedEnergies when we use LC calculation.

aradi: I think switching it off would be the best solution for the moment. We can still re-enable it for LC once it had been implemented.

inseonglee: Then current commit already did it. When we use LC with atom resolved energies, tAtomicResolved set to false. After implementing it, I will remove this part. Could you inform me when this PR is merged to master branch? As long as this PR is merged, I will start to merge ssr part.

PR #314, opened 2019-11-05
Implementation of DFT-D4 in the dftb+ infrastructure

awvwgk:
- [x] add DFT-D4 as dispersion correction to `dftb+`
  - [x] add keywords in parser/input
  - [x] Parameter setup for D4 calculator
  - [x] coordination number and derivatives
  - [x] Dispersion energy calculation and derivatives
    - [x] two-body contribution 
    - [x] non-additive three-body contribution (ATM) 
  - [x] electronegativity equilibration charges (requires modified Ewald summation)
    - [x] molecular
    - [x] periodic
    - ~MPI parallel~
- [x] add detailed documentation on DFT-D4
- [x] add tests for DFT-D4
- [x] follow style guide


aradi:
A few remarks at this early stage already:

- We'll have to see, whether D4 can be hidden beyond the usual dispersion interface. As it has a charge-dependence, we may need to handle it independently. (E.g. the Tkatchenko-type many body dispersion which is also being integrated has its own interface as well, as it is not compatible with the usual dipersion API).

- We use the NAG-compiler for testing purposes. However, that compiler unfortunately did not implement submodules yet, so for the moment, let's use conventional modules only.

awvwgk:
All right, the `submodule` was anyway just temporary (to speed up compilation). I will remove it when I am done with the implementation.

D4(EEQ) can be hidden behind the usual dispersion interface, for GFN2-xTB the D4 needs a tighter integration to the SCC, but this is an issue for the GFN2-xTB implementation, not for DFT-D4.

awvwgk:
@aradi I hope you are fine with compiling the DFT-D4 parameter file into `dftb+`. It not as heavy as the `copyc6` from DFT-D3 but still 6k lines.

aradi:
Do we have any other possibilities? :wink:  I think we can live with it it, although one could think about a more dynamical solution, where the data is read from disk on demand. The only question were then, how to find the external data reliably during run (hard coded paths derived from the install prefix?)

awvwgk:
I tested my implementation against the reference implementation and the results match so far. I still need to have a closer look on the strain derivatives and the Ewald summation to make sure everything works as expected.

I will start with adding tests and detailed documentation soon and will be happy to address style issues.

aradi:
@awvwgk Sebastian, short question: are we supposed to look at it and consider it for merging or is this still considered to be work in progress?

awvwgk:
@aradi, from the implementation side I am done with this feature. I still need to figure out how and where to add tests and docs, but was occupied otherwise this week, so I couldn't really look into it yet.

awvwgk:
@aradi now also the docs and tests are done, at least I hope so. I am looking forward to hear your comments.

prog/dftb+/lib_dftb/dispdftd4.F90
```diff
@@ -0,0 +1,1101 @@
...
+    end if
+
+    ! handle CN and charge contributions to the gradient by matrix-vector operation
+    call dgemv('n', 3*nAtom, nAtom, 1.0_dp, dcndr, 3*nAtom, dEdcn, 1, 1.0_dp, &
+        &      gradients, 1)
+    call dgemv('n', 3*nAtom, nAtom, 1.0_dp, dqdr, 3*nAtom, dEdq, 1, 1.0_dp, &
+        &      gradients, 1)
+    ! handle CN and charge contributions to the stress tensor
+    call dgemv('n', 9, nAtom, 1.0_dp, dcndL, 9, dEdcn, 1, 1.0_dp, stress, 1)
+    call dgemv('n', 9, nAtom, 1.0_dp, dqdL, 9, dEdq, 1, 1.0_dp, stress, 1)
```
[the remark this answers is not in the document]
awvwgk: This will not work without a lot of reshapes and temporary arrays here.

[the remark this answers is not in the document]
awvwgk: Alternatively I can patch the `gemv()` wrapper.

test/prog/dftb+/tests
```diff
@@ -63,6 +63,8 @@ dispersion/2H2O                        #? MPI_PROCS <= 1
 dispersion/2H2O_uff                    #? MPI_PROCS <= 1
 dispersion/2H2O_dftd3_zero             #? WITH_DFTD3 and MPI_PROCS <= 1
 dispersion/2H2O_dftd3_bj               #? WITH_DFTD3 and MPI_PROCS <= 1
+dispersion/2H2O_dftd4                  #? MPI_PROCS <= 1
+dispersion/diamond_dftd4               #? MPI_PROCS <= 1
```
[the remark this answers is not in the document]
awvwgk: Just copied from above and I haven't tried the code with MPI.

prog/dftb+/lib_dftb/dftd4param.F90
```diff
@@ -0,0 +1,500 @@
...
+      cncount = 0
+      cncount(0) = 1
+      iZp1 = izp(iAt1)
+      if (calculator%atoms(iZp1) == 0) then
+        calculator%numberOfReferences(iZp1) = refn(iZp1)
+        do iRef1 = 1, calculator%numberOfReferences(iZp1)
+          iSec = refsys(iRef1,iZp1)
+          eta1 = calculator%gc * calculator%chemicalHardness(iSec)
+          zEff1 = calculator%effectiveNuclearCharge(iSec)
+          alpha = sscale(iSec)*secaiw(:,iSec) &
```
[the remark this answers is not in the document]
awvwgk: Yes, it is. This information has to be hardcoded somewhere, since the reference systems do not carry their sum formula around.

prog/dftb+/lib_dftb/dispdftd4.F90
```diff
@@ -0,0 +1,1101 @@
...
+        if (iAt1 /= iAt2f) then
+          dcndL(:, :, iAt2f) = dcndL(:, :, iAt2f) + stress
+        endif
+
+      enddo
+    enddo
+  end subroutine getCoordinationNumber
+
+  !> cutoff function for large coordination numbers
+  subroutine cutCoordinationNumber(nAtom, cn, dcndr, dcndL, cn_max)
```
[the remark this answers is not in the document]
awvwgk: This is new for PBCs: https://doi.org/10.26434/chemrxiv.10299428.v1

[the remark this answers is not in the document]
awvwgk: This is indeed new for PBCs: https://doi.org/10.26434/chemrxiv.10299428.v1

prog/dftb+/lib_dftb/dispdftd4.F90
```diff
@@ -0,0 +1,1101 @@
...
+        do iCount1 = 1, calc%countNumber(iref1, iSp1)
+          wf = iCount1 * calc%wf
+          gw = weightCN(wf, cn(iAt1), calc%referenceCN(iRef1, iSp1))
+          expw = expw + gw
+          expd = expd + 2*wf*(calc%referenceCN(iRef1, iSp1) - cn(iAt1)) * gw
+        enddo
+
+        gwk = expw * norm
+        if (gwk /= gwk) then
+          if (maxval(calc%referenceCN(:calc%numberOfReferences(iSp1), iSp1)) &
```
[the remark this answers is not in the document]
awvwgk: Done.

prog/dftb+/lib_dftb/dispdftd4.F90
```diff
@@ -0,0 +1,1101 @@
...
+    @:ASSERT(allocated(this%calculator))
+
+    if (this%tPeriodic) then
+      call DispersionEnergy(this%calculator, this%nAtom, coords, this%izp, &
+          & neigh, img2CentCell, this%recPoint, this%parEwald, &
+          & this%energies, this%gradients, stress=this%stress, volume=this%vol)
+    else
+      call DispersionEnergy(this%calculator, this%nAtom, coords, this%izp, &
+          & neigh, img2CentCell, this%recPoint, this%parEwald, &
+          & this%energies, this%gradients)
```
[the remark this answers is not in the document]
awvwgk: Made this optional and only pass it for the periodic case.

prog/dftb+/lib_dftb/dispdftd4.F90
```diff
@@ -0,0 +1,1101 @@
...
+    if (present(dcndr)) then
+      do iAt = 1, nAtom
+        dcndr(:, :, iAt) = dcndr(:, :, iAt) * dCutCN(cn(iAt), cnmax)
+      enddo
+    endif
+
+    do iAt = 1, nAtom
+      cn(iAt) = cutCN(cn(iAt), cnmax)
+    enddo
+
```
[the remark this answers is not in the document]
awvwgk: Than I will prefer array operations.

prog/dftb+/lib_dftb/dispdftd4.F90
```diff
@@ -0,0 +1,1101 @@
...
+    real(dp), parameter :: k5 = 19.08857_dp
+    real(dp), parameter :: k6 = 2*11.28174_dp**2
+
+    integer :: iAt1, iSp1, iAt2, iAt2f, iSp2, iNeigh
+    real(dp) :: r2, r1, rc, vec(3), countf, countd(3), stress(3, 3), dEN, ENscale
+
+    cn = 0.0_dp
+    dcndr = 0.0_dp
+    dcndL = 0.0_dp
+
```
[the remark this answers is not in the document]
awvwgk: Done

prog/dftb+/lib_dftb/dispdftd4.F90
```diff
@@ -0,0 +1,1101 @@
...
+    !> Names of species.
+    character(*), intent(in) :: speciesNames(:)
+
+    !> Lattice vectors, if the system is periodic.
+    real(dp), intent(in), optional :: latVecs(:, :)
+
+    real(dp) :: recVecs(3, 3), maxREwald, maxGEwald
+    integer :: iAt
+
+    @:ASSERT(.not. allocated(this%calculator))
```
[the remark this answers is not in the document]
awvwgk: Removed

prog/dftb+/lib_dftb/dispdftd4.F90
```diff
@@ -0,0 +1,1101 @@
...
+  end type DispDftD4
+
+
+  interface init
+    module procedure :: DispDftD4_init
+  end interface init
+
+  interface assignment(=)
+    module procedure :: SymbolToNumber
+  end interface assignment(=)
```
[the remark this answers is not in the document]
awvwgk: Okay.

prog/dftb+/lib_dftb/dftd4param.F90
```diff
@@ -0,0 +1,500 @@
...
+
+  #:include 'dftd4_references.fh'
+
+contains
+
+  !> charge scaling function
+  pure elemental function zetaScale(a, c, qref, qmod) result(zeta)
+    real(dp),intent(in) :: qmod, qref
+    real(dp),intent(in) :: a, c
+    real(dp) :: zeta
```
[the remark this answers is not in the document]
awvwgk: Renamed and documented.

prog/dftb+/lib_dftb/dftd4param.F90
```diff
@@ -0,0 +1,500 @@
...
+    &   9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,26,  & ! Rb-Xe
+    &   9,10,11,30,31,32,33,34,35,36,37,38,39,40,41,42,43,  & ! Cs-Lu
+    &  12,13,14,15,16,17,18,19,20,21,22,23,24,25,26, & ! Hf-Rn
+    !  just copy & paste from above
+    &   9,10,11,30,31,32,33,34,35,36,37,38,39,40,41,42,43,  & ! Fr-Lr
+    &  12,13,14,15,16,17,18,19,20,21,22,23,24,25,26] ! Rf-Og
+
+  !> covalent radii (taken from Pyykko and Atsumi, Chem. Eur. J. 15, 2009,
+  !  188-197), values for metals decreased by 10 %
+  real(dp), parameter :: CovalentRadiusD3(118) = [ &
```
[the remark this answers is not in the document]
awvwgk: Done.

prog/dftb+/lib_dftb/dftd4param.F90
```diff
@@ -0,0 +1,500 @@
...
+          calculator%referenceCN(iRef1,iZp1) = refcovcn(iRef1,iZp1)
+          cncount(iCN) = cncount(iCN) + 1
+          calculator%referenceAlpha(:,iRef1,iZp1) = max(0.0_dp, &
+            & ascale(iRef1,iZp1)*(alphaiw(:,iRef1,iZp1) &
+            &                     - hcount(iRef1,iZp1)*alpha))
+        enddo
+        do iRef1 = 1, calculator%numberOfReferences(iZp1)
+          iCN = cncount(nint(refcn(iRef1,iZp1)))
+          calculator%countNumber(iRef1,iZp1) = iCN*(iCN+1)/2
+        enddo
```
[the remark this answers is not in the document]
awvwgk: This is the count/number of gaussian functions used in the weighting procedure for the references to allow a higher resolution for nearby reference systems. It's Fig. 4 in the paper.

awvwgk:
@aradi Thanks for the review.

I carried the habit of zeroing at allocation along from previous projects, since I usually zero variables anyway. I can remove all `source` statements from the code. In case it should be kept, I could use a `fypp` macro to switch between zero and NaN initialization.

prog/dftb+/lib_dftb/dftd4param.F90
```diff
@@ -0,0 +1,500 @@
...
+    calculator%gam = gam
+    calculator%rad = rad
+    calculator%kcn = kcn
+
+    calculator%sqrtZr4r2 = sqrtZr4r2
+    calculator%covalentRadius = CovalentRadiusD3
+    calculator%electronegativity = paulingEN
+    calculator%ChemicalHardness = ChemicalHardness
+    calculator%EffectiveNuclearCharge = EffectiveNuclearCharge
+
```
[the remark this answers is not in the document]
awvwgk: They currently map via the atomic number but could also use the species type instead, a copy allows to change this behaviour in one place. Also, I have encountered a lot of problems with OpenMP and global (constant) arrays in the past, since they seem to be handled inconsistently between compilers and OpenMP versions.

prog/dftb+/lib_dftb/dispdftd4.F90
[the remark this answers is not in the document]
aradi: Yes, I'd favor patching `gemv()` to avoid precision dependent calls within the main code. Actually,  as far as I can see, the current `gemv()` wrapper allows for everything you need (you just pass those as optional arguments...). Or do I overlook something?

prog/dftb+/lib_dftb/dftd4param.F90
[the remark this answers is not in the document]
aradi: OK.

prog/dftb+/lib_dftb/dispdftd4.F90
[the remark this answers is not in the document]
awvwgk: dcndr/dcndL are rank 3 arrays.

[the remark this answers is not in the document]
aradi: You're right! Then don't touch the `gemv()` wrapper as you would have to duplicate them for rank-3 arrays. We could, however, either apply a reshape  or even better create a 2D-pointer:
```
  dcndL2(1:9, 1:nAtom) => dcndL(1:3, 1:3, 1:nAtom)
```
and pass that to the `gemv()` wrapper.

awvwgk:
I was waiting for the second half of the review, but didn't expect to get it patched in directly, so many thanks to you.

I will look through the code again next week and see what I can do to move this forward.

Should I add the D4 damping parameters already?

aradi:
Sorry, I thought in case of stylistic changes it is easier to push directly as to describe them. I should have, however, probably made a PR against your branch instead of pushing directly, so that it is less invasive. Anyway, for me this is almost ready to go, apart of the S9 parameter, which does not pop up in the equations in the manual, but have to be set somehow in the input.

As for the D4 damping parameters, yes, adding them to the manual would make sense, for sure.

awvwgk:
Not a problem at all, that's why I allowed “edits from maintainers” in the first place.

aradi:
Oh, I see S9 is already in the manual. However, I think we should describe it more, as it is not clear from the first glance, that it should be either 0.0 or 1.0 depending whether 3 body terms should be included. Or do any intermediate values make sense? If not, maybe we should make it a logical (`ThreeBodyTerms = Yes/No`) and set internally 0.0. or 1.0.

awvwgk:
Somebody might want to try 0.9 or 1.1 for the `s9`.
I already got a request to implement this option (in Orca, not DFTB+).

prog/dftb+/lib_dftb/dispdftd4.F90
```diff
@@ -0,0 +1,1101 @@
...
+        iSp2 = species(iAt2f)
+        r2 = neighDist2(iNeigh, iAt1)
+        r1 = sqrt(r2)
+
+        rc = covalentRadius(iSp1) + covalentRadius(iSp2)
+
+        if (tENScale) then
+          dEN = abs(electronegativity(iSp1) - electronegativity(iSp2))
+          dEN = k4 * exp(-(dEN + k5)**2 / k6)
+        else
```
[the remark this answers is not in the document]
awvwgk: This is probably a typo in the paper.

awvwgk:
So there are two comments left. Than we should be ready to go.

1. Anything I should do about testing with MPI?
2. Should I remove `calc%gc`? Having the possibility to scale the hardnesses from the input is not a bad thing, but should be used very cautiously. I would prefer to leave it in since it is consistent with the `dftd4` and `xtb` implementation.

aradi:
As for the MPI, I updated the test diamond test to be big enough to be tested with MPI up to 4 procs. That works OK. Of course, the DFT-D4 part itself is not MPI-parallelized. At some point, we should think about it, as for a 64-diamond cell the D4-calculation takes order of magnitude longer than the rest...

Leave the charge steepness as freely adjustable parameter.

awvwgk:
That's caused mainly by the ATM term, there are quite a lot triples. Usually it is better to disable it with SQM methods.

prog/dftb+/lib_dftb/encharges.F90
```diff
@@ -0,0 +1,679 @@
...
+      call getCoulombDerivsCluster(nAtom, coords, species, rad, qVec, aMatdr, aTrace)
+    end if
+    do iAt1 = 1, nAtom
+      aMatdr(:, :, iAt1) = aMatdr(:, :, iAt1) + dcndr(:, :, iAt1) * Xfac(iAt1)
+      aMatdL(:, :, iAt1) = aMatdL(:, :, iAt1) + dcndL(:, :, iAt1) * Xvec(iAt1)
+    end do
+
+    if (present(gradients)) then
+      call gemv(gradients, aMatdr, qVec, beta=1.0_dp)
+    end if
```
aradi: Why do we need `beta=1.0_dp` here? The array `gradients` is `intent(out)`, so its content at this point is not well defined.

prog/dftb+/lib_dftb/encharges.F90
```diff
@@ -0,0 +1,679 @@
...
+
+    if (present(stress)) then
+      call gemv(stress, aMatdL, qVec, beta=1.0_dp)
+    end if
```
aradi: Again, it is not clear, why `beta=1.0_dp` here.

prog/dftb+/lib_dftb/dispdftd4.F90
```diff
@@ -0,0 +1,1290 @@
...
+        & calculator%electronegativity, .false., cn, dcndr, dcndL)
+    call cutCoordinationNumber(nAtom, cn, dcndr, dcndL, cn_max=8.0_dp)
+
+    allocate(q(nAtom), dqdr(3, nAtom, nAtom), dqdL(3, 3, nAtom))
+
+    call getNrOfNeighboursForAll(nNeigh, neigh, calculator%cutoffEwald)
+
+    call getEEQCharges(nAtom, coords, species, nNeigh, neigh%iNeighbour, neigh%neighDist2,&
+        & img2CentCell, recPoint, parEwald0, vol, calculator%chi, calculator%kcn, calculator%gam,&
+        & calculator%rad, cn, dcndr, dcndL, qAtom=q, dqdr=dqdr, dqdL=dqdL)
```
aradi: What is the point of passing `dcndr` and `dcndL` here? The corresponding variables (`stress` and `gradients`) are in `getEEQCharges()` `intent(out)`, so their values should not be used. And the next `getCoordinationNumber()` call rewrites them again, without using them.

prog/dftb+/lib_dftb/encharges.F90
awvwgk: It is more useful with `beta=1.0_dp`, but than it should be `intent(inout)` instead of `intent(out)`.

prog/dftb+/lib_dftb/encharges.F90
awvwgk: Like above.

prog/dftb+/lib_dftb/dispdftd4.F90
awvwgk: They are needed to generate `dqdr` and `dqdL`. See here:

https://github.com/dftbplus/dftbplus/blob/5a3a45ce7929cb552bb24bbe3aefc51dbddf8066/prog/dftb%2B/lib_dftb/encharges.F90#L647-L650

awvwgk: This is always done, even if `gradients` and `stress` are not present.

aradi: Sorry, I mixed up the correspondence between dummy arguments and actual ones. You are right.

awvwgk:
Should we attempt to include MPI parallelisation here? It would mostly be `distributeRangeInChunks` for the atom loops. BLACS could be used for the EEQ part, but this gets somewhat redundant with the implementation in `dftbp_coulomb`.

aradi:
No, at least not in this pull request. I think, we are ready to go, I'd like to merge this to master this week.

awvwgk:
All right, thanks for your help with this PR.

doc/dftb+/manual/dftbp.tex
```diff
@@ -2343,6 +2343,9 @@ \subsection{Dispersion}
   code~\cite{grimme-jcp-132-154104,grimme-jcp-32-1456-1465} (see
   section~\ref{sec:dftbp.DftD3}). Modification hydrogen bond interaction
   strengths (see section \ref{sec:dftbp.hcorr}).
+\item \is{DftD4}: Dispersion is calculated using the D4
+  model~\cite{caldweyher-jcp-147-034112,caldeweyher-jcp-150-154122,} (see
```
bhourahine: Extra comma at end of citation

doc/dftb+/manual/dftbp.tex
```diff
@@ -2614,6 +2617,127 @@ \subsubsection{DftD3 optional settings}
...
+\end{equation*}
+
+The D4 model uses the Becke--Johnson damping function for two-body contributions:
+\begin{equation*}
+   E_{\text{disp}}^{(2)} = -\frac{1}{2} \sum_{A\neq B} \sum_{n=6,8,10} s_n
+   \frac{C_n^{AB}}{r_{AB}^n + f(R_0^{AB})}
+\end{equation*}
+with
+\begin{equation*}
+   f(R_0^{AB}) = a_1 R_0^{AB} + a_2 \text.
```
bhourahine: f(R_0^{AB}) = a_1 R_0^{AB} + a_2.
will render correctly.

doc/dftb+/manual/dftbp.tex
```diff
@@ -2614,6 +2617,127 @@ \subsubsection{DftD3 optional settings}
...
+the description of dispersion interactions, but is also more expensive.
+
+As the DFTB energy functional is largely determined by the underlying
+parameterisation (the Slater--Koster-files) and the chosen DFTB model (e.g.\
+non-scc, scc, 3rd order, etc.), there are no universal parameter choices which
+can be used with all settings, but some relevant choices for various
+parameterisation are given in Appendix \ref{app:dftd4const}.
+
+\subsubsection{DftD4 optional settings}
+Apart from the , you can also adjust
```
bhourahine: Seems to be a typo, what should be before the comma?

doc/dftb+/manual/manual.bib
```diff
@@ -412,6 +412,30 @@ @Article{grimme-jcp-32-1456-1465
  pages = {1456-1465}
 }
 
+@article{caldweyher-jcp-147-034112,
+author = {Eike Caldeweyher and Christoph Bannwarth and Stefan Grimme},
+title = {Extension of the D3 dispersion coefficient model},
+journal = {J. Chem. Phys.},
```
bhourahine: The is a defined string in titles.bib for this journal, so
journal = jcp,
works (and handles the spacing correctly on 'Chem. ').

doc/dftb+/manual/manual.bib
```diff
@@ -412,6 +412,30 @@ @Article{grimme-jcp-32-1456-1465
...
+number = {3},
+pages = {034112},
+year = {2017},
+doi = {10.1063/1.4993215}
+}
+
+@article{caldeweyher-jcp-150-154122,
+author = {Caldeweyher,Eike  and Ehlert,Sebastian  and Hansen,Andreas  and Neugebauer,Hagen  and Spicher,Sebastian  and Bannwarth,Christoph  and Grimme,Stefan },
+title = {A generally applicable atomic-charge dependent London dispersion correction},
+journal = {J. Chem. Phys.},
```
bhourahine: See comment above about journal.

prog/dftb+/lib_dftb/dftd4param.F90
```diff
@@ -0,0 +1,549 @@
...
+    real(dp) :: s9
+
+    !> scaling parameter for <r4>/<r2> expectation value based critical radii.
+    real(dp) :: a1
+
+    !> constant offset of critical radii.
+    real(dp) :: a2
+
+    !> exponent of for the zero-damping function used for non-addititive
+    !  triple dipole contributions.
```
bhourahine: !> triple dipole contributions.

prog/dftb+/lib_dftb/dftd4param.F90
```diff
@@ -0,0 +1,549 @@
...
+    & 0.58884173_dp, 0.68441115_dp, 0.56999999_dp, 0.56999999_dp, 0.56999999_dp, &
+    & 0.56999999_dp, 0.56999999_dp, 0.56999999_dp, 0.56999999_dp, 0.56999999_dp, &
+    & 0.56999999_dp, 0.56999999_dp, 0.56999999_dp, 0.56999999_dp, 0.56999999_dp, &
+    & 0.56999999_dp, 0.87936784_dp, 1.02761808_dp, 0.93297476_dp, 1.10172128_dp, &
+    & 0.97350071_dp, 1.16695666_dp, 1.23997927_dp, 1.18464453_dp, 1.14191734_dp, &
+    & 1.12334192_dp, 1.01485321_dp, 1.12950808_dp, 1.30804834_dp, 1.33689961_dp, &
+    & 1.27465977_dp]
+
+  !> Element-specific chemical hardnesses for the electronegativity equilibration
+  !  charges used in DFT-D4.
```
bhourahine: !>

prog/dftb+/lib_dftb/dftd4param.F90
```diff
@@ -0,0 +1,549 @@
...
+    real(dp), allocatable :: referenceCN(:, :)
+    real(dp), allocatable :: referenceCharge(:, :)
+    real(dp), allocatable :: referenceAlpha(:, :, :)
+    real(dp), allocatable :: referenceC6(:, :, :, :)
+
+  end type DftD4Calculator
+
+
+  !> Element-specific electronegativity for the electronegativity equilibration
+  !  charges used in DFT-D4.
```
bhourahine: !>

prog/dftb+/lib_dftb/dftd4param.F90
```diff
@@ -0,0 +1,549 @@
...
+    & 0.15348654_dp, 0.05054344_dp, 0.11000000_dp, 0.11000000_dp, 0.11000000_dp, &
+    & 0.11000000_dp, 0.11000000_dp, 0.11000000_dp, 0.11000000_dp, 0.11000000_dp, &
+    & 0.11000000_dp, 0.11000000_dp, 0.11000000_dp, 0.11000000_dp, 0.11000000_dp, &
+    & 0.11000000_dp,-0.02786741_dp, 0.01057858_dp,-0.03892226_dp,-0.04574364_dp, &
+    &-0.03874080_dp,-0.03782372_dp,-0.07046855_dp, 0.09546597_dp, 0.21953269_dp, &
+    & 0.02522348_dp, 0.15263050_dp, 0.08042611_dp, 0.01878626_dp, 0.08715453_dp, &
+    & 0.10500484_dp]
+
+  !> Element-specific CN scaling constant for the electronegativity equilibration
+  !  charges used in DFT-D4.
```
bhourahine: !>

prog/dftb+/lib_dftb/dftd4param.F90
```diff
@@ -0,0 +1,549 @@
...
+    &-0.06188828_dp,-0.06568203_dp,-0.11000000_dp,-0.11000000_dp,-0.11000000_dp, &
+    &-0.11000000_dp,-0.11000000_dp,-0.11000000_dp,-0.11000000_dp,-0.11000000_dp, &
+    &-0.11000000_dp,-0.11000000_dp,-0.11000000_dp,-0.11000000_dp,-0.11000000_dp, &
+    &-0.11000000_dp,-0.03585873_dp,-0.03132400_dp,-0.05902379_dp,-0.02827592_dp, &
+    &-0.07606260_dp,-0.02123839_dp, 0.03814822_dp, 0.02146834_dp, 0.01580538_dp, &
+    &-0.00894298_dp,-0.05864876_dp,-0.01817842_dp, 0.07721851_dp, 0.07936083_dp, &
+    & 0.05849285_dp]
+
+  !> Element-specific charge widths for the electronegativity equilibration
+  !  charges used in DFT-D4.
```
bhourahine: !>

prog/dftb+/lib_dftb/dftd4param.F90
```diff
@@ -0,0 +1,549 @@
...
+    & 1.59815118_dp, 1.76401732_dp, 1.63999999_dp, 1.63999999_dp, 1.63999999_dp, &
+    & 1.63999999_dp, 1.63999999_dp, 1.63999999_dp, 1.63999999_dp, 1.63999999_dp, &
+    & 1.63999999_dp, 1.63999999_dp, 1.63999999_dp, 1.63999999_dp, 1.63999999_dp, &
+    & 1.63999999_dp, 1.47055223_dp, 1.81127084_dp, 1.40189963_dp, 1.54015481_dp, &
+    & 1.33721475_dp, 1.57165422_dp, 1.04815857_dp, 1.78342098_dp, 2.79106396_dp, &
+    & 1.78160840_dp, 2.47588882_dp, 2.37670734_dp, 1.76613217_dp, 2.66172302_dp, &
+    & 2.82773085_dp]
+
+  !> Element-specific chemical hardnesses for the charge scaling function used
+  !  to extrapolate the C6 coefficients in DFT-D4.
```
bhourahine: !>

prog/dftb+/lib_dftb/dftd4param.F90
```diff
@@ -0,0 +1,549 @@
...
+    & 0.21240778_dp, 0.07263525_dp, 0.09422158_dp, 0.09920295_dp, 0.10418621_dp, &
+    & 0.14235633_dp, 0.16394294_dp, 0.18551941_dp, 0.22370139_dp, 0.00000000_dp, &
+    & 0.00000000_dp, 0.00000000_dp, 0.00000000_dp, 0.00000000_dp, 0.00000000_dp, &
+    & 0.00000000_dp, 0.00000000_dp, 0.00000000_dp, 0.00000000_dp, 0.00000000_dp, &
+    & 0.00000000_dp, 0.00000000_dp, 0.00000000_dp, 0.00000000_dp, 0.00000000_dp, &
+    & 0.00000000_dp, 0.00000000_dp, 0.00000000_dp, 0.00000000_dp, 0.00000000_dp, &
+    & 0.00000000_dp, 0.00000000_dp, 0.00000000_dp]
+
+  !> Effective nuclear charges from the def2-ECPs used for calculating the
+  !  reference polarizibilities for DFT-D4.
```
bhourahine: !>

prog/dftb+/lib_dftb/encharges.F90
```diff
@@ -0,0 +1,679 @@
...
+
+    logical :: tPeriodic
+    integer :: nDim
+    integer :: iAt1, iSp1, ii, jj
+    real(dp) :: tmp
+
+    tPeriodic = allocated(recPoint)
+
+    nDim = nAtom + 1
+    allocate(xVec(nDim), xFac(nAtom), aMatdr(3, nAtom), aMat(nDim, nDim), aInv(nDim, nDim), &
```
bhourahine: aMatdr has 3 indices, so this doesn't compile. Should this be 
aMatdr(3, nAtom, nAtom)
?

prog/dftb+/lib_dftb/encharges.F90
```diff
@@ -0,0 +1,679 @@
...
+        & qVec(nDim), aMatdL(3, 3, nDim), aTrace(3, nAtom), source=0.0_dp)
```
bhourahine: For some compilers, source can only appear of there is only one item may appearing in alloc-list.

prog/dftb+/lib_dftb/encharges.F90
```diff
@@ -0,0 +1,679 @@
...
+    allocate(ipiv(nDim), source=0)
```
bhourahine: Is this being used as an initialization method? As with other cases, should be left to compiler choice (ditto above).

prog/dftb+/lib_dftb/encharges.F90
awvwgk: What? This is definitely wrong. Was introduced in 790ceab4, but not by me...

awvwgk: Fixed now.

prog/dftb+/lib_dftb/encharges.F90
awvwgk: Removed.

prog/dftb+/lib_dftb/encharges.F90
awvwgk: Removed.

doc/dftb+/manual/dftbp.tex
awvwgk: Fixed

doc/dftb+/manual/dftbp.tex
awvwgk: Fixed

prog/dftb+/lib_dftb/dispdftd4.F90
```diff
@@ -0,0 +1,1287 @@
...
+    !> Upgraded stress
+    real(dp), intent(out), optional :: stress(:, :)
+
+    !> Volume, if system is periodic
+    real(dp), intent(in), optional :: volume
+
+    integer, allocatable :: nNeigh(:)
+
+    real(dp) :: sigma(3, 3)
+    real(dp), parameter :: charge = 0.0_dp  ! FIXME
```
awvwgk: Maybe not as finished as I thought. How do we get the molecular charge in here in a reliable way?

bhourahine: Do you need the whole molecule charges or atom charges? The later is simple (but gross or net charge if so), the former would require something like a topological analysis to define 'molecules'.

awvwgk: Exactly, the net charge of the entire system is needed to constraint the number of electrons for DFT-D4 correctly in the EEQ model.

bhourahine: Is 'entire system' all atoms, in which case this is the net charge. Or is it net charge for each distinct molecular fragment? We can discuss details tomorrow.

awvwgk: Yes, all atoms, I need the net charge. There are no molecular fragments involved here, since this would lead to a non-continuous PES.

aradi: The total net charge of the system stays constant during the calculation (unless we have open boundary conditions as in a transport calculation). So, probably we should pass it, when `DispDftD4` becomes initialized and make sure, that D4 is disallowed for transport. On the long term, we would maybe need and `updateCharges()` call for the dispersion interface...

awvwgk: If this `updateCharges()` routine is general enough to return a potential to the Hamiltonian, one might generalize the `DispIface` for self-consistent DFT-D4 and MBD with this approach.

For now setting it in the input is fine, we just have to guarantee that the net charge is set before we setup the dispersion model.

bhourahine:
Commit 0c10972f seems to have broken the dispersion/2H2O_dftd4 regression test.
In dispdftd4.F90 both dEdq and dEdcn should be zeroed after allocation as there is a loop accumulating onto their values.

awvwgk:
Ups, good we have tests for this. Thanks for the fix.

prog/dftb+/lib_dftb/dispdftd4.F90
```diff
@@ -0,0 +1,1289 @@
...
+        if (iAt1 /= iAt2f) then
+          stress = stress + sigma
+        else
+          stress = stress + 0.5_dp * sigma
+        end if
+
+        dEdcn(iAt1) = dEdcn(iAt1) - dc6dcn1 * dEr
+        dEdcn(iAt2f) = dEdcn(iAt2f) - dc6dcn2 * dEr
+        dEdq(iAt1) = dEdq(iAt1) - dc6dq1 * dEr
+        dEdq(iAt2f) = dEdq(iAt2f) - dc6dq2 * dEr
```
awvwgk: Oh, I overlooked this one! It should be guarded by `if (iAt1 /= iAt2f) then` otherwise the gradient is wrong.

awvwgk:
I checked the entire code again against the reference implementation. For molecular systems both the EEQ model and the D4 dispersion are up to some numerical noise identical. I found an issue I have to investigate further before this can be merged: For periodic systems the EEQ model does not produce the same charges as in the reference implementation.

In `dftd4` I evaluate the Coulomb-matrix in 5×5×5 Wigner-Seitz-cell in real and reciprocal space. But for this implementation I just took the implementation from `dftbp_coulomb` and used a static neighbourlist instead of the dynamic one. Did I miss something here?

awvwgk:
Finally, this ready from my side.

aradi:
@awvwgk Sorry, one last thing before merge: I checked in our script comparing forces/stress for a given configuration with those obtained by finite differences. The results are good for the molecular case (match up to 1e-9), but quite unsatisfying for the periodic one. As for the diamond it would take ages, I used a 8 atom GaAs cluster as example, see attached files. 
[dftb_in.hsd.txt](https://github.com/dftbplus/dftbplus/files/4135924/dftb_in.hsd.txt)
[geo.gen.template.txt](https://github.com/dftbplus/dftbplus/files/4135938/geo.gen.template.txt)
I get an agreement of 1e-4 in the forces and only 1e-2 in the lattice derivatives. You can try it yourself (use the "hyb" SK-set) with something like `DFTBROOT/utils/numcheck/calcderivs -c ~/ramdisk/build/dftbplus/gnu/prog/dftb+/dftb+`.

Could you have a look a this? Is this inaccuracy just due to the paramater choice in the input? Or is this something more fundamental?

awvwgk:
The only difference from this code to the one used in `xtb` and `dftd4` is the Ewald summation for the EEQ, so I guess I should start looking at `dftbp_encharges`.

awvwgk:
I usually have done the Ewald implementation using a 5×5×5 super cell of the Wigner–Seitz cell, so effectively a 7×7×7 super cell. But this shouldn't be an issue at all, should it?

awvwgk:
Indeed if I set the charges constant (by zeroing dqdr & dqdL) the difference between numerical and analytical is only 2e-9. In retrospect, it was a really stupid idea to use an uncharged system like diamond in the testing phase... :man_facepalming: at least it explains why I didn't noticed.

I will fix this ASAP.

aradi:
Great, thanks in advance! I'd suggest to add the GaAs example also as a test to the test suite as it should be more reliable in catching errors.

awvwgk:
It was _the_ sign error in the CN derivative which keeps haunting me. Never going to get this consistent over all implementations, but now it is consistent within DFTB+.

Otherwise, it is finally working as intended. So thanks a lot for testing and finding this error.

awvwgk:
I figured I will just OMP parallelize the whole code. I'm not sure what your policy is here, but I prefer to setup everything explicitly with `default(none)` since `default(shared)` usually leads to errors. Also you seem to prefer the ALL-CAPS version here, which makes it even more unpleasant to read and write :).

prog/dftb+/lib_dftb/encharges.F90
```diff
@@ -0,0 +1,695 @@
...
+
+    ! Step 1: contruct RHS for linear system
+    do iAt1 = 1, nAtom
+      iSp1 = species(iAt1)
+      tmp = kcn(iSp1) / (sqrt(cn(iAt1)) + small)
+      xVec(iAt1) = -chi(iSp1) + tmp*cn(iAt1)
+      xFac(iAt1) = -0.5_dp*tmp
+    end do
+    xVec(nDim) = charge
+    xFac(nDim) = 0.0_dp
```
aradi: Note: this writes beyond the boundaries of `xFac` (triggers run-time error with GNU compiler when compiled in debug mode).

awvwgk: Removed.

awvwgk:
Just rebased and cleaned up this PR.

aradi:
Thanks a lot. I have checked it again. Finite difference forces and analytic ones are now acceptably close to each other. Still they are much more noisy (3 orders of magnitude more than for example LJ-dispersion). Maybe we should discuss this shortly (next devel-meeting?). If you think, that this is not an implementation issue, I am happy to have this branch merged, provided @bhourahine is also OK with it.

awvwgk:
Let's discuss this at the next devel-meeting. I have an idea for the origin (the Ewald sum), but still no idea why it is in issue in this implementation but not in `dftd4` or `xtb`. Could be a simple oversight on my side or a more fundamental problem.

bhourahine:
I'm getting a compiler warning with gcc7 and 8:
prog/dftb+/lib_dftb/dispdftd4.f90:158:24-34:

     call symbolToNumber(this%izp, speciesNames(species0))
                        1         2
Warning: INTENT(OUT) actual argument at (1) might interfere with actual argument at (2).
might be harmless, but comes up with production flags (i.e. without any particular warnings enabled).

awvwgk:
Might be an issue with the `elemental subroutine`, I will just replace it with a loop than.

bhourahine:
There might be a bug in the force evaluation. I stripped the routine down to the 2 body term and for the GaAs case get numerical force checks that look like:
Difference between obtained and reference forces:
      -2.735891943767E-12      -3.096628671461E-12       1.171962786894E-12
      -9.917138724808E-13       5.909841409649E-12      -3.595198574610E-12
       2.553090588741E-12      -1.400644452559E-04       1.393986522622E-12
      -4.173208003125E-13      -4.542688707837E-12      -1.121172592429E-12
      -2.749054166612E-12       5.928909164836E-12      -1.468008318530E-12
      -8.329118915840E-14       1.400644336425E-04       5.904326723719E-12
       2.156784489503E-12       1.560250924770E-12       1.425989968468E-12
      -1.037322381854E-12      -2.134765617317E-14      -3.711894168228E-12

awvwgk:
@bhourahine, this is an artefact from the numerical differentiation. Just move the atom with the spurious gradient contribution a bit and check again, the artefact will be gone.

```patch
diff --git a/geo.gen.template b/geo.gen.template
index 79ea3e72..755002a0 100644
--- a/geo.gen.template
+++ b/geo.gen.template
@@ -2,7 +2,7 @@
  Ga As
  1 1 0.00 0.00 0.00
  2 2 0.17 0.25 0.25
- 3 1 0.50 0.62 0.00
+ 3 1 0.50 0.42 0.00
  4 2 0.75 0.75 0.32
  5 1 0.38 0.00 0.50
  6 2 0.75 0.10 0.75
```

aradi:
@bhourahine could you try it with a stepsize of 1e-3 (instead of the 1e-5) with the critical geometry? I made the experience, that the agreement between numerical and analytical forces degrade for D4 if the stepsize goes below 1e-3 Ansgström. I am still not sure, though, where it comes from. My guess would be more on numerical instability / noise in the D4 calculation as problems with the finite difference calculation.  (Without D4 I get agreement up to 1e-9 with a step size of 1e-5...).

bhourahine:
@aradi  I tried the larger step size as well, without a significant change in results.
@awvwgk I guess the question is on which side the difference comes from at the original geometry. We could try a more sophisticated numerical differentiator (higher order, extrapolation to limit, ...) is its not obvious.
One test I haven't done recently is to look what other dispersion model give for force errors.

awvwgk:
@bhourahine With the modified geometry we arrive at:

```
...
Difference between obtained and reference forces:
       6.639819041809E-10      -6.138754409316E-10      -2.032428021020E-09
       2.279917616678E-09      -7.210614050290E-10       7.227968154555E-10
       1.907800563361E-09      -1.178456929263E-09       1.849387968288E-09
       2.067615105170E-09       1.170149276086E-09       8.489017752633E-10
      -1.817979954023E-09       2.551010652718E-10       9.998537414679E-10
       9.454846679885E-10      -5.153777543621E-10       6.474279896918E-10
      -2.305635508904E-10       1.453821686301E-09       1.277009735803E-09
      -7.402150663249E-10       3.377009366590E-10       1.139094374381E-09
Max diff in any force component:
       2.279917616678E-09
...
Difference between obtained and reference lattice derivatives:
       2.672237854558E-06      -1.809075546172E-07       4.497044225212E-07
      -8.412911196160E-08       1.705759536695E-06      -5.068280745984E-07
       3.350999484108E-07      -6.422317633419E-07       3.701807014343E-06
Max diff in any lattice derivatives component:
       3.701807014343E-06
```

We should be able to compare to the `dftd4` standalone and the `xtb` implementation of D4, we can for molecular systems or completely neutral periodic systems (which did match perfectly), but we cannot for periodic systems with charges because the Coulomb matrix for the EEQ is not identical, which is due to the different setup of the Ewald summation (even with the same Ewald splitting parameter).

awvwgk:
We should discuss this at the devel-meeting in detail. There is still the option to use the same Wigner–Seitz *ansatz* like in `dftd4` or `xtb`, than we should be able to reproduce the exact same energy and gradient. But I'm not sure if this is really an issue, the geometry optimizations I tested so far did converge smoothly.

awvwgk:
I cleaned up the branch and rebased it against master.

bhourahine:
The fix for the old make system is around line 135 of prog/dftb+/make.common add in

# DFTD4
L_INCS += -I$(ROOT)/external/dftd4refs

If everyone is happy, we can then merge.

bhourahine:
prog&#47;dftb+&#47;make.common
even (markdown, humph)

PR #321, opened 2019-11-13
Transport uupload

bhourahine:
Uploaded contacts can now be optionally used for Fermi level values, contain +U data for various orbital potentials and be a binary file.


pecchia:
I went through the changes and I think they are fine. 
I was wandering if it make sense (in order to simplify the input), to have just one debug global option for formatted I/O of ALL files. I agree that a fine-grained control is more flexible but yet one more option to remember.

bhourahine:
Thanks. For the binary vs. ascii file control, I think that should be a separate pull request, so will merge this one now.

PR #324, opened 2019-11-15
Option to set masses in input for modes

bhourahine:
Matches functionality already in DFTB+

### DIFF
--- a/prog/modes/initmodes.F90
+++ b/prog/modes/initmodes.F90
@@ -30,7 +30,7 @@ module dftbp_initmodes
 
 
   !> program version
-  character(len=*), parameter :: version =  "0.02"
+  character(len=*), parameter :: version =  "0.03"
 
   !> root node name of the input tree
   character(len=*), parameter :: rootTag = "modes"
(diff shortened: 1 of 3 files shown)


prog/modes/initmodes.F90
```diff
@@ -326,4 +335,55 @@ subroutine readGeometry(geonode, geo)
...
+      call convertByMul(char(modifier), massUnits, child, rTmp)
+      do jj = 1, size(pTmpI1)
+        iAt = pTmpI1(jj)
+        if (masses(iAt) >= 0.0_dp) then
+          call detailedWarning(child3, "Previous setting for the mass  of atom" // i2c(iAt) //&
+              & " overwritten")
+        end if
+        masses(iAt) = rTmp
+      end do
+      deallocate(pTmpI1)
```
aradi: A minor node: You don't necessarily need to deallocate here, as it will be automatically deallocated on entry to `convAtomRangeToInt` since the dummy argument is `allocatable, intent(out)`.

PR #328, opened 2019-11-27
Fixes

pecchia:
1. I found an arror in closeing timer postGeoOpt
2. I added to make.common  -DWITH_OMP otherwise there are problems in initprogram.F90. 
I don't know exactly if make.common still works with cmake but for the moment I am still using the old makefiles


bhourahine:
You should really do this via a 'feature branch', not your master.

PR #330, opened 2019-12-01
REKS method within DFTB formalism

inseonglee:
- [x] REKS variables
- [x] parser and initialize
- [x] energy calculation
- [x] gradient calculation
- [x] relaxed density (QM/MM gradient)
- [x] stress calculation in only single-state REKS
- [x] add tests and manuals


inseonglee:
Currently, energy calculation parts are implemented in the code. Please see the codes and give me any comments about the contents or structures! I will implement the remaining parts within few days.

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -3907,6 +4108,7 @@ subroutine initArrays(env, electronicSolver, tForces, tExtChrg, tLinResp, tLinRe
         allocate(rhoSqrReal(sqrHamSize, sqrHamSize, nSpin))
       end if
     end if
+    ! TODO : chargePerShell is already allocated in 1501 line
     allocate(chargePerShell(orb%mShell, nAtom, nSpin))
```
inseonglee: Here, chargePerShell is allocated, but this array is already allocated in 1501 line. When I comment out this part, then the code shows seg fault error. Could you check this part why this line is needed?

inseonglee: I found the reason for this seg fault. Due to the 'out' intent characteristic, allocated(chargePerShell) changes to F. In InitArray routine, chargePerShell should be removed in input arguments as well as 4112 line.

inseonglee:
@aradi @bhourahine Here, I add the gradient, stress, QM/MM parts in REKS. In energy calculation part in main.F90, REKS should have own SCC loop (lpSCC_REKS) with original DFTB scc loop (lpSCC). In the part of REKS gradient and stress, the detailed process is shown in reksinterface.F90 file.

I wonder that which one seems to be nice, well-organized code structure for the energy calculation?
1st choice is current structure (lpSCC_REKS in main.F90), 2nd choice is to move the energy calculation loop to reksinterface.F90 so that the overall structure is similar with gradient calculation part.
Is this explanation enough to you?
Please, give me some advice about the code structure. 

Remaining works: manual, test files, rearrange overall code structure for concise, neat form, remove useless comments

bhourahine:
There is a similar issue to your energy loop in another extension to the code as a part of
https://github.com/dftbplus/dftbplus/pull/335
See
https://github.com/deshaym/dftbplus/blob/58ef58d418ec3b38c423d538e4107911dc2a8373/prog/dftb%2B/lib_dftbplus/main.F90#L166
for their determinant loop. I think we should make a plan for the best way to handle these types of multiple ground state loops in a general way (there will at some point also be a configuration interaction DFTB development with the same issues). But this may take some time due to current deadlines and the holiday closures.

bhourahine:
I also have a question regarding stress, the LC-DFTB implementation does not include periodic boundary conditions at the moment, but there are some comments in your code about Gamma point periodicity (I am still in the process of learning about your routines). Does this mean you have implemented a lattice sum for the exchange operation, or is this for isolated molecules in periodic boxes where there are not interactions across cell boundaries?

inseonglee:
@bhourahine As you said, the LC-DFTB does not support the periodicity. Thus, in (LC-)DFTB/SSR method, only DFTB/SSR support the periodicity. In this case DFTB/SSR can calculate the periodic condition as well as stress calculation or lattice optimization.

Therefore, the inconsistent matching error in routine "checkGammaPoint" can occur in only DFTB/SSR with periodic condition, not LC-DFTB/SSR.

inseonglee:
@aradi @bhourahine The implementation of REKS is complete! I finished to add manuals and test files for REKS calculations. Please see the changed things, and please give me any comments!

The remaining work is OpenMP works in reksgrad.F90 file. (I inserted TODO lines in this file.)
#338 issue is related to this thing.

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -978,6 +979,9 @@ module dftbp_initprogram
   !> Is this DFTB/SSR formalism
   logical :: tREKS
 
+  !> data type for REKS
+  type(TReksCalc), save :: reks
```
bhourahine: Can you make this allocatable (as is done for other variables like sccCalc)? The allocated status then shows the type of calculation (and also in later Fortran standards allows it to be combined with optional arguments for subroutines). Also, please remove the save attribute, as this will need to be removed to make the code thread safe for the API.

bhourahine:
With gfortran 7.4 I get segmentation faults for:
The following tests FAILED:
         15 - dftb+_reks/HBI+H2O_periodic_stress (Failed)
         16 - dftb+_reks/HBI+H2O_periodic_stress_disp (Failed)
         17 - dftb+_reks/PSB3_2SSR (Failed)
         18 - dftb+_reks/PSB3_2SSR_extchrg_periodic (Failed)
         19 - dftb+_reks/PSB3_2SSR_rangesep (Failed)
         20 - dftb+_reks/PSB3_2SSR_rangesep_extchrg (Failed)
         21 - dftb+_reks/PSB3_2SSR_rangesep_extchrg_blur (Failed)
         22 - dftb+_reks/PSB3_2SSR_rangesep_tuning (Failed)
         23 - dftb+_reks/PSB3_SA (Failed)

Most of these are segmentation faults at:
#3  0x55799db1d89e in __dftbp_reksinterface_MOD_getreksgradproperties
        at /home/ben/Src/dftb/dftbplus/build/prog/dftb+/lib_reks/reksinterface.f90:425
which is at
 else
          dipoleMoment(:) = 0.0_dp
 end if
which is probably not allocated in this case. The thing happens for other cases where the dipoleMoment variable is accessed when .not. tDipoleMoment, but occurring at
#3  0x55c573f8d34e in getstateinteraction
        at /home/ben/Src/dftb/dftbplus/build/prog/dftb+/lib_dftbplus/main.f90:6922
in subroutine getStateInteraction.

inseonglee:
@bhourahine Thanks the comments! As you said, the segmentation fault occurs due to the dipoleMoment variable when it is not allocated. I fix this problem and remove save attribute in reks variable.

For changing reks variable with allocatable attribute, I just add directly allocatable attribute and allocate(reks) in 2608 line. Is there some parts to modify the code for allocatable attribute?
I have never used allocatable attribute for type variables in fortran coding so far.
I passed the test calculations with this modification, please check this.

bhourahine:
The allocated use looks correct, but this now means that allocated(reks) can be used to test for this type of calculation, instead of also having to pass around tREKS. This would require allocatable attributes in the subroutine interfaces.

Also the  @:SAFE_DEALLOC(reks) macro in the initprogram.F90 will need to be applied. This is required for use with the API, as the derived types need to be cleaned up at the end of the calculation in case there is a later call to make a new calculation.

bhourahine:
I have a question regarding granularity of timings. You have separate timings for each iteration of several algorithms.
As you will have noticed, there are global timers in the code for larger sections of the calculation (which given the accuracy of timing short sections of code, give better averages per iteration). Hence is there a good reason not to use these aggregate timings? For example, do large changes in run time for individual iterations indicate a problem about to occur?

inseonglee:
@bhourahine There is one simple reason for me to add new timings using `system_clock`. I just want to add a local variable in some routines such as SCC and CP-REKS iterations. It is not essential part for other routines such as `getExtChrgGradients` in reksinterface.F90 file. (check elapsed time for OMP implementation)

In REKS calculation, only first SCC iteration takes more time than other iteractions due to the diagonalization of H0 matrix.

bhourahine:
OK, I think I understand. It might be a good idea if this is re-factored to use the environmental timing mechanisms (as this will be kept up to date with any future code changes as  well as robustly reporting in the event of MPI aborts for the parallel code). But these routines will need some expansion to do this, so I will discuss with Bálint and get back to you.
I have another question regarding the test for a closed shell system. At the moment this is tested with mod(int(n),2)==0, but DFTB+ can also have fractionally charged systems. So should that also be checked for explicitly? The test uses int, so can miss this for some values of n.

inseonglee:
@bhourahine For your second question, REKS can treat only closed shell system. (even number of electrons) It automatically determines the fractional occupations from the energy of microstates, thus we do not use the input for electron filling information. Currently, REKS(2,2) is only possible choice, so the meaning of `mod(int(dble(nEl(1))),2) /= 0` is just check for even number of electrons, not fractional occupations.

bhourahine:
OK, so I will then also add a test for non-integer electrons (as at the moment this can be set in the input).

prog/dftb+/lib_dftbplus/mainio.F90
```diff
@@ -3743,8 +3744,7 @@ end subroutine printSccInfo
 
 
   !> Prints info about scc convergence.
-  subroutine printReksSccInfo(iSccIter, Etotal, diffTotal, &
-      & sccErrorQ, t1, t2, FONs, tSSR22, tSSR44)
+  subroutine printReksSccInfo(iSccIter, Etotal, diffTotal, sccErrorQ, deltaT, reks)
```
inseonglee: Originally my purpose for this file and whole codes is that the the `reks` type variables is not used except interface routines. Thus, this variable is not added to lib_reks/*.F90 files except reksinterface.F90. Likewise I did not add the `reks` variable in this file (mainio.F90), but you did. Is this more nice/well-organized coding for DFTB+?

prog/dftb+/lib_dftbplus/mainio.F90
```diff
@@ -5156,44 +5148,17 @@ subroutine writeReksDetailedOut1(fd, nGeoSteps, iGeoStep, tMD, tDerivs, &
...
-    real(dp), intent(in) :: enLtot(:)
-
     !> Third order DFTB
     logical, intent(in) :: t3rd
 
     !> Whether to run a range separated calculation
     logical, intent(in) :: tRangeSep
 
+    !> data type for REKS
+    type(TReksCalc), intent(in) :: reks
```
inseonglee: Same for line 3747.

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -1136,11 +1133,15 @@ subroutine initProgramVariables(input, env)
...
-    ! nSpin should be two but some variables such as qOutput
-    ! should be treated in restricted scheme. Here nSpin
-    ! is one and changes to two in the latter part.
+    if (input%ctrl%reksIni%tREKS) then
+      ! REKS follows spin-restricted open-shell scheme so nSpin should be two but some variables
+      ! such as qOutput should be treated in restricted scheme. Here nSpin is one and changes to two
+      ! in the latter part.
+      allocate(reks)
+      if (tSpin) then
+        call error("REKS is not compatible with standard spin polarization")
```
inseonglee: This line is repeated check with 2586 line. If you check `tSpin` in this line, then I want to the sentence about spin constants like below (2586 line).

bhourahine: As this duplicates the test, I have removed the test here and simplified the logic as well as expanding the comments

prog/dftb+/lib_dftbplus/mainio.F90
bhourahine: We are in the process of changing to simplify the main code by passing derived types into subroutines, instead of components of a type. There are two reasons for this:
1) The higher level code becomes simpler as details of what the data type contains is only needed to be known inside the routines directly processing that type, not the higher level code.
2) It makes the calling arguments to routines shorter in the higher level code, but more importantly it is less likely to be broken by other developers (who are less familiar with the routines) if they change the code. This is since the compile time checking can just examine the derived type position in the argument list, which would be different from other variables the calling list. If the individual components are passed, there is a higher chance of requiring positional orderings of variables with the same signature to be kept correct.

The big disadvantage, as I'm sure you know, is that if parts of the derived type should be intent in and others intent out,  then the whole type needs to be intent (inout). We are in the process of working out a solution for part of this problem with use of protected and another mechanism (at least for large arrays) to allow read only contents from pointers (the current Fortran standard does not offer a mechanism to do this). But this is unlikely to be decided on before the next code release.

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -1132,20 +1132,15 @@ subroutine initProgramVariables(input, env)
     ! Basic variables
     tSccCalc = input%ctrl%tScc
     tDFTBU = input%ctrl%tDFTBU
-    tSpin = input%ctrl%tSpin
```
inseonglee: Why do you remove this line? Then, `tSpin` always set to false in the code even though we use spin polarisation in the input file. I think that the line `tSpin = input%ctrl%tSpin` should remain here.

prog/dftb+/lib_dftbplus/mainio.F90
```diff
@@ -45,6 +45,7 @@ module dftbp_mainio
   use dftbp_elstatpot, only : TElStatPotentials
   use dftbp_message
   use dftbp_rekscommon
+  use dftbp_reksvar, only : TReksCalc
```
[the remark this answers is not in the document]
bhourahine: For the next release, we will be changing to use cmake to build the code. There are two external projects which package DFTB+, their developers have said that they find it much easier to working with cmake/pkg-config than hand written make files.

prog/dftb+/lib_dftbplus/mainio.F90
inseonglee: Thanks for your answers. Then, should I change the variables used in the routines of lib_reks/*.F90 files for simplification?

prog/dftb+/lib_dftbplus/initprogram.F90
bhourahine: Yes, you are correct. Thank you for pointing out the mistake.

prog/dftb+/lib_dftbplus/mainio.F90
bhourahine: Yes, that would be very helpful, thank you.

inseonglee: Then, I will modify the parts related to the type `reks` variables in the whole codes. Please check the commits in the later.

bhourahine:
Thank you for moving to pass the derived type to the interfaces. Please let me know when you are ready for the branch to be updated to include the changes on the main repository (the major relevant change are likely to be #392 and #398 ).

inseonglee:
The modification is not finished yet. I have to check other routines such as gradient calculations.

I have one question. Should I change the expression `dble()` to `real(**,dp)` for other routines as you did?

bhourahine:
> I have one question. Should I change the expression `dble()` to `real(**,dp)` for other routines as you did?

Yes, that would be very useful (this gives a better control over the numerical model if everything is done consistently). 
Using the old *8 extension instead of the kind specification also prevents at least one compiler from building the code.

inseonglee:
@bhourahine Now the modification to REKS branch is finished. Please see the change carefully, and if there is wrong points or bugs, then I will fix them. In my cluster, it passes all tests related to reks calculation. So, I am ready for the branch to be updated.

inseonglee:
@bhourahine Currently, I am in the process of fixing some broken parts for REKS part occurring in the recent update. Here I have three questions for scc/coulomb routines.

1. Recently, you modified (separated?) scc/coulomb routines, and this affects the REKS part in these files. Originally, the routine `getPeriodicInfo` was included in scc.F90 since scc type variable has the periodic information. But this now moved to coulomb routine, so I want to know the criterion (I guess you want to keep only charge-related variables to SCC module). Am I correct?

2. When I implemented REKS, I added four routines to coulomb.F90 (`addInvRPrimeClusterMat`, `addInvRPrimeClusterAsymmTderiv`, and periodic versions). Here only `addInvRPrimeClusterAsymmTderiv` is included in interface `addInvRPrime` and `addInvRPrimeClusterMat` is not included in the interface (just procedure and public state). Is there a specific reason for this?

3. For the coulomb routine, three routines (`getPeriodicInfo`, `addInvRPrimeClusterMat`, and periodic one) has public property, and other routines such as `updateShifts` do not have. Thus, should the public property be removed for these routines?

aradi:
@inseonglee You can execute `ctest -R reks` to run only the tests containing "reks" in their name.

aradi:
I've tried to compile the REKS-branch with the NAG compiler and run the tests. There are several issues:
- `int(kind=8)` is not valid Fortran. The `kind` value should not be considered the byte size of the type (although some compilers use it that way.) The proper solution would be to use `kind=int64` using the `int64` from the intrinsic `iso_fortran_env_module`
- You are apparently accessing non-allocated allocatable arrays, for example in `initprogram.F90`. Unfortunately, the GCC does not report this error (not even in debug mode :disappointed: ). I'll try to have a look at it.

Additionally, I'd ask you to remove the individual timings from the routines. One should use either the central infrastructure (with cumulative timings) or not time at all. I absolutely understand, that during development you wish to see indidivual loop times, but it should be removed before it is merged to master.

bhourahine:
@inseonglee yes the Coulomb routines were separated from the SCC module, this was mostly required now due to the changes from the xTB developer (but is also needed in the longer term for several other extensions or to improve existing features).
The existing REKS test cases did not catch the broken parts you mention (I tested with gcc7.5 and intel18.2 compilers, plus there is now continuous integration testing on github enabled for gnu compilers). I see from [d5d565a2](https://github.com/dftbplus/dftbplus/pull/330/commits/d5d565a2c3cad9d67c39eac4a3623b1f36462975) that the changes so far are tests against other types of model, but are there further problems introduced by [a624c18c](https://github.com/dftbplus/dftbplus/pull/330/commits/a624c18cb95eb5a2a45dafcf199c9b36699e7721) that should have specific tests added?

Regarding the specific points:

1) The electrostatics will all be packaged into the Coulomb module, this removes the need for the scc module routines to have any information regarding things like boundary conditions (and allows the Coulomb routines to be used independently in other types of Hamiltonian without requiring interaction with the scc module). In the longer term the Coulomb module will also absorb the Poisson solver interface as well as any external potential couplings (solvent models, polarizable force-fields for QM/MM, ...).

2) As mentioned in the commit, there needs to be future refactoring for the changes introduced, but probably this should happen after your branch is merged into the main code (hopefully this can be done very soon). At the moment, these *AsymmT* routines are in a transitional state, but as to package them as procedure will also require changing the relevant REKS routines, I did not do that as  part of merging in the main repository.

3) Yes, these should eventually changed to private. Eventually it would be good to avoid having to extract the getPeriodicInfo data and instead make all of the lattice sum information private to the Coulomb module (for example, then methods like p3m or multipolar methods can be transparently introduced without modifying the REKS  code).

prog/dftb+/lib_dftbplus/main.F90
```diff
@@ -85,6 +85,12 @@ module dftbp_main
   use dftbp_initprogram, only : TRefExtPot
   use dftbp_qdepextpotproxy, only : TQDepExtPotProxy
   use dftbp_taggedoutput, only : TTaggedWriter
+  use dftbp_rekscommon
+  use dftbp_reksen
+  use dftbp_reksfon
+  use dftbp_reksinterface
+  use dftbp_reksproperty
+  use dftbp_reksvar
```
aradi: Please create a high level `dftbp_reks` module which exports all the names form those 6 modules. This would clutter the code less and make maintainance easier, if the internal structure of the REKS part changes.

inseonglee:
@aradi Thanks for the convenient command! For the individual timing routines, I will remove those after all developments (modifications) are done.

prog/dftb+/lib_dftbplus/main.F90
```diff
@@ -304,6 +310,10 @@ subroutine runDftbPlus(env)
 
     call env%globalTimer%stopTimer(globalTimers%postGeoOpt)
 
+    if (allocated(reks)) then
+      call REKS_destroy(reks)
```
aradi: This is unnecessary. Type `TReksCalc` only contains allocatable arrays and needs no extra finalization. This lines can be removed and also the implementation of the  `REKS_destroy()` routine.

prog/dftb+/lib_dftbplus/main.F90
```diff
@@ -1361,6 +1505,9 @@ subroutine reallocateSparseArrays(sparseSize, ham, over, H0, rhoPrim, iHam, iRho
       deallocate(ERhoPrim)
       allocate(ERhoPrim(sparseSize))
     end if
+    if (allocated(reks)) then
+      call REKS_reallocate(reks, sparseSize)
```
aradi: Please implement it as type bound procedure, so that the call looks as `call reks%reallocate(sparseSize)`

inseonglee:
@bhourahine Thanks for the answers. First you asked the further problems introduced by a624c18c, and I did not see all changes yet. Currently, I see the relavent files in lib_dftbplus and lib_dftb directories. Thus, I will check whether further problems about commit a624c18c exist or not.

For your answers (1 - 3), I understood 1 and 3 answers, but I am slightly confused about 2 answer. When we see the coulomb routines, AsymmTderiv routines are already included in the interface routine, but Mat routines are not included. But you said that you do not include these routines to main parts since AsymmTderiv routines are in a transitional state. (Here I think the main part corresponds to interface routine, `AddInvRPrime`). Thus, these explains seem to be opposite thing.

@bhourahine @aradi I will modify the REKS branch according to your comments ASAP. Thanks for the all comments!

prog/dftb+/lib_dftbplus/main.F90
inseonglee: @aradi I am trying to create a high level `dftbp_reks` module, but I have no idea for this. I think that the parts in the loop lpSCC_REKS should be moved to the `dftbp_reks` module to rearrange those modules (`dftbp_reksen`, `dftbp_reksfon`, etc). Is there any method to use a high level `dftbp_reks` without moving the SCC iteration loops?

aradi: For the moment, just create one `dftbp_reks` module (within the `lib_reks`) directory, which just imports those 6 modules and exports the public names:
```
module dftbp_reks
  import dftbp_rekscommon
  [..]
  implicit none
  private

  public :: [Add all public names here]

end module dftbp_reks
```
And then, in `main.F90` you would just only import that single module. (At some point later we may turn it then into a module with submodules).

prog/dftb+/lib_reks/reks.F90
```diff
@@ -0,0 +1,60 @@
...
+  use dftbp_reksfon
+  use dftbp_reksinterface
+  use dftbp_reksproperty
+  use dftbp_reksvar
+!  import dftbp_rekscommon
+!  import dftbp_reksen
+!  import dftbp_reksfon
+!  import dftbp_reksinterface
+!  import dftbp_reksproperty
+!  import dftbp_reksvar
```
inseonglee: @aradi I created a high level module `dftbp_reks`, but it shows an error during compilation when I use `import` statement. Simply, I changed to `use` statement, and it shows no error and it works well. But, is this OK for this module? Actually, I am not familiar with `import` statement, so there can exist some mistake in the code. Please see this commit 2eee49d.

bhourahine:
@inseonglee the import statement is for use within the [interface](https://www.ibm.com/support/knowledgecenter/SS2MB5_14.1.0/com.ibm.xlf141.bg.doc/language_ref/importstatement.html) only, but this is to limit the scope of things that are added via a use statement. I think the original comment probably meant use, but @aradi can confirm.

bhourahine:
> For your answers (1 - 3), I understood 1 and 3 answers, but I am slightly confused about 2 answer. When we see the coulomb routines, AsymmTderiv routines are already included in the interface routine, but Mat routines are not included. But you said that you do not include these routines to main parts since AsymmTderiv routines are in a transitional state. (Here I think the main part corresponds to interface routine, `AddInvRPrime`). Thus, these explains seem to be opposite thing.

The comment is more that addInvRPrimeClusterMat/addInvRPrimePeriodicMat are currently typebound, but AsymmTderiv are not. To make them type bound would eventually require some refactoring.
At the moment, once the high level dftbp_reks module is completed, we should merge this. Then the future changes should be made as later pull requests.

inseonglee:
@bhourahine Thanks for answer. For the coulomb routines, I will not modify the addInvRPrime routines. When you modify the routines about Mat or AsymmT routines, I will check them later.

inseonglee:
Additionally, I do not add `assert` to REKS subroutines. Is this essential part for code? If so, where should I add it?

aradi:
@inseonglee `@:ASSERT` macros can be used to check for the existence of a given condition (typically at the entry into a subroutine), which the Fortran compiler can not easily check (neither at compile, nor at run-time in debug mode). So, if you have anything like this, you may use it. You should not use it for trivial things e.g. checking allocation status, as compilers are quite good at detecting it nowadays.

aradi:
@inseonglee What is the status of the PR from your side? Is there anything you still want to change? I just want to make sure, that you are ready, before I look once more over it.

inseonglee:
@aradi From my side, I have to remove the timing for individual iteration. If this is done, then I am ready for this PR. For `assert` macros, I will check several given conditions, so that it can be merged with other PR.

aradi:
@inseonglee Great, thanks, than I'll now start looking over it. :smile:

inseonglee:
@bhourahine @aradi Now, I am ready for this PR. As I mentioned before, I will make other PR for `assert` macros, REKS grandient for 3rd order terms and so on.

p.s) This point is not related to REKS. In 639 and 728 lines of main.F90 file, the band.out and detailed.out is written for every SCC iteration. I think that it can give some IO overload (or burden) since dftb calculation is very fast. Thus, if possible, I want to move this routine into line 739 so that DFTB+ write band.out and detailed.out only when it converged. For the current REKS SCC case, these files are written when it converged.

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -95,6 +95,7 @@ module dftbp_initprogram
   use dftbp_qdepextpotproxy, only : TQDepExtPotProxy
   use dftbp_forcetypes, only : forceTypes
   use dftbp_elstattypes, only : elstatTypes
+  use dftbp_reksvar
```
aradi: `use dftbp_reks, only : .........`

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -1138,11 +1139,15 @@ subroutine initProgramVariables(input, env)
     tSccCalc = input%ctrl%tScc
     tDFTBU = input%ctrl%tDFTBU
     tSpin = input%ctrl%tSpin
-    tREKS = .false.
-    if (tSpin) then
+    nSpin = 1
+    if (input%ctrl%reksIni%tREKS) then
+      ! REKS follows spin-restricted open-shell scheme so nSpin should be two in the main code, but
+      ! some variables such as qOutput should be treated in a restricted scheme. Here nSpin is set
+      ! to one and changes to two later in the initialization.
```
aradi: I can not find the place, where `nSpin` changes to 2 in the initialization.

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -2590,11 +2609,86 @@ subroutine initProgramVariables(input, env)
...
-        & chargePerShell, occNatural, velocities, movedVelo, movedAccel, movedMass, dipoleMoment)
+        & tPrintExcitedEigvecs, tDipole, allocated(reks), orb, nAtom, nMovedAtom, nKPoint, nSpin,&
+        & nExtChrg, indMovedAtom, mass, denseDesc, rhoPrim, h0, iRhoPrim, excitedDerivs, ERhoPrim,&
+        & derivs, chrgForces, energy, potential, TS, E0, Eband, eigen, filling, coord0Fold,&
+        & newCoords, orbitalL, HSqrCplx, SSqrCplx, eigvecsCplx, HSqrReal, SSqrReal, eigvecsReal,&
+        & rhoSqrReal, occNatural, velocities, movedVelo, movedAccel, movedMass, dipoleMoment)
+
+
+    if (allocated(reks)) then
+      ! here, nSpin changes to 2 for REKS
```
aradi: Why? Where does `nSpin` occur here?

Also, would it be possible to factor this entire block out into a `checkReksConsistency()` subroutine, followed by a `call TReksCalc_init()`? Otherwise it makes the already quite unreadable initialization more unreadable...

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -3276,6 +3389,94 @@ subroutine initProgramVariables(input, env)
 
     end if
 
+    if (allocated(reks)) then
+
```
aradi: Please factor out the entire block into a subroutine (e.g. `printReksInfo()`). I know, that all other components are also not factored out into routines, but we badly need a modularization of the initialization, so let's not make the situation worse by dumping further tons of code in it. :wink:

doc/dftb+/manual/dftbp.tex
```diff
@@ -3715,6 +3718,153 @@ \subsection{Casida}
...
+
+In general, REKS calculation can be classified as single-state REKS, SA-REKS and SI-SA-REKS.
+In single-state REKS, only ground state is calculated and it can treat the state with multireference
+character. SA-REKS and SI-SA-REKS can calculate the vertical excitation energies. The difference
+is that the state-interaction term is considered in SI-SA-REKS so that more accurate states can
+be generated. The corresponding oscillator strengths as well as excited state geometry optimisation
+can be performed with these options, details of the resulting output files are given in appendix
+\ref{sec:reks_files}. DFTB/SSR is currently implemented for the LC-DFTB level of theory with external
+point charges or dispersion corrections. Third order SCC calculation can be considered in only
+energy evaluation. The gamma point periodic systems can be calculated using DFTB/SSR without
```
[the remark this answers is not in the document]
inseonglee: @aradi In current Implementation of REKS, LC-DFTB/SSR can be used with external point charges or dispersion correction for both energy and gradient calculation. But, LC-DFTB/SSR with 3rd order correction is possible with only energy calculation (At the moment, the REKS gradient for 3rd order correction term is not implemented in the current version, so I will add this part with other PR).

- Summary for several features (for non-periodic system)
Energy: SCC, LC, 3rder SCC, external point charges (for QM/MM calculation), dispersion
gradient: SCC, LC, external point charges, dispersion

doc/dftb+/manual/dftbp.tex
```diff
@@ -3715,6 +3718,153 @@ \subsection{Casida}
...
+    \kw{MemoryLevel}                        & i & & 2 & \\
+  \end{ptable}
+
+  \begin{description}
+
+  \item[\is{EnergyFunctional}] Specifies the minimized energy functional in DFTB/SSR.
+    The minimized energy functionals can be PPS (\is{EnergyFunctional} = 1) or (PPS+OSS)/2
+    (\is{EnergyFunctional} = 2) state in DFTB/SSR(2,2) case. The former considers only
+    single-state REKS and the latter considers SA-REKS or SI-SA-REKS. The inclusion of
+    state-interaction term is determined by \is{UseSSRstate} option.
```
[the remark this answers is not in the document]
inseonglee: @aradi For DFTB/SSR(2,2), there is only two choices for the `EnergyFunctional`. When the size of active space increases, the number of choice (and state) for the `EnergyFunctional` increases.
For example, the number of states used in SSR(4,4) is nine, so the length for the `EnergyFunctional` is too long if we includes all states. It may be `(PPS+OSS1+OSS2+ ... + DSPS)`. I think this is very complicated so I just use only number as choice. Is this explanation enough?

prog/dftb+/lib_dftb/coulomb.F90
```diff
@@ -2415,4 +2427,360 @@ end subroutine addNeighbourContribsStress
...
+    ind3 = size(self%rCellVec,dim=1)
+    ind4 = size(self%rCellVec,dim=2)
+
+    if (.not. allocated(gVec)) then
+      allocate(gVec(ind1,ind2))
+      allocate(rVec(ind3,ind4))
+    end if
+
+    gVec(:,:) = self%gLatPoint
+    rVec(:,:) = self%rCellVec
```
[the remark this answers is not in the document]
inseonglee: @aradi The variable `gVec` and `rVec` is only allocated in this routine, not `REKS_init`. Does this not show any errors even though the size of these variables changes for geometry optimization?

prog/dftb+/lib_dftb/coulomb.F90
```diff
@@ -180,6 +190,8 @@ module dftbp_coulomb
     module procedure addInvRPrimeClusterAsymm
     module procedure addInvRPrimePeriodic
     module procedure addInvRPrimePeriodicAsymm
+    module procedure addInvRPrimeClusterAsymmTderiv
+    module procedure addInvRPrimePeriodicAsymmTderiv
```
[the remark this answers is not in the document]
inseonglee: @aradi Here, T derivative means the derivative of Hamiltonian with respect to the coordinates. When we treat the external point charges, we have to consider the contribution of point charges with respect to the coordinates to calculate the REKS grandient. Thus, these routines are used to calculate the routine `getPc2ndTerms_` (line 5377) in reksgrad.F90 file.

The difference between `addInvRPrimeClusterAsymmTderiv` and `addInvRPrimeClusterAsymm` is shown in the line 1360 and 2632 of coulomb.F90 file. I think that these two routines can be combined with additional arguments to express REKS.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -6251,4 +6255,151 @@ subroutine parseRangeSeparated(node, input)
...
+    type(fnode), pointer, intent(in) :: node
+
+    !> Control structure to fill
+    type(TControl), intent(inout) :: ctrl
+
+    !> geometry of the system
+    type(TGeometry), intent(in) :: geo
+
+    !> Minimized energy functional; 1: PPS, 2: (PPS+OSS)/2
+    call getChildValue(node, "EnergyFunctional", ctrl%reksIni%Efunction, default=2)
```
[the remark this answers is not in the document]
inseonglee: @aradi Currently, I have no idea for the shorter and descriptive names for this. I think we have to some discussions?

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -6251,4 +6255,151 @@ subroutine parseRangeSeparated(node, input)
...
+    !> 2: calculate all possible states
+    call getChildValue(node, "EnergyLevel", ctrl%reksIni%Elevel, default=1)
+    !> Calculate SSR state (SI term is included)
+    !> 1: calculate SSR state, 0: calculate SA-REKS state
+    call getChildValue(node, "UseSSRstate", ctrl%reksIni%useSSR, default=1)
+
+    !> Target SSR state
+    call getChildValue(node, "TargetState", ctrl%reksIni%rstate, default=1)
+    !> Target microstate
+    call getChildValue(node, "TargetStateL", ctrl%reksIni%Lstate, default=0)
```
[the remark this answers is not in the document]
inseonglee: @aradi Here, L index indicate the micorstate. In SSR(2,2), there are 6 microstaes. First microstate is perfectly spin-paired configuration, second one is doubly excited configuration, third and fourth are open shell configuration, and finally fifth and sixth are triplet configuration.

So, when we want to calculate the gradient (or property) for certain microstate such as triplet configuration, then `TargetStateL` should be 5.

For overall coding and documentation, I followed the paper (JCTC, 2019, 15, 3021) for index. (e.g. T derivative, index L, etc.)

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -6251,4 +6255,151 @@ subroutine parseRangeSeparated(node, input)
...
+    call readSpinTuning(node, ctrl, geo%nSpecies)
+
+    !> Calculate transition dipole moments
+    call getChildValue(node, "TransitionDipole", ctrl%reksIni%tTDP, default=.false.)
+
+    !> Algorithms to calculate analytic gradients
+    !> 1: preconditioned conjugate gradient (PCG)
+    !> 2: conjugate gradient (CG)
+    !> 3: direct inverse-matrix multiplication
+    call getChildValue(node, "GradientLevel", ctrl%reksIni%Glevel, default=1)
```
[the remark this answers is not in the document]
inseonglee: @aradi As you said, that is "algorithm" to used in CP-REKS equation (calculation of gradient). The reason is simple, just I want to match the name (word) of the variable `EnergyLevel`, so this variable is `GradientLevel`. But, I am OK when you changed the name of this variable.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -6251,4 +6255,151 @@ subroutine parseRangeSeparated(node, input)
...
+      if (nAt /= nType) then
+        call detailedError(node, "Incorrect number of 'SpinTuning' block: " &
+            & // i2c(nAt) // " supplied, " &
+            & // i2c(nType) // " required.")
+      end if
+      allocate(tmpTuning(1,nAt))
+      call asArray(realBuffer, tmpTuning)
+      call destruct(realBuffer)
+      allocate(ctrl%reksIni%Tuning(nType))
+      ctrl%reksIni%Tuning(:) = tmpTuning(1,:)
```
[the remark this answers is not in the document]
inseonglee: @aradi For this routine, I just copied the reading part of initial velocities for MD case since this part is too hard for me about Buffer or other terms. If possible, could you modify this routine? For the variable `SpinTuning`, the exact size is just the `nType` and the order of atomic species follows in the order of gen input file.

prog/dftb+/lib_dftbplus/initprogram.F90
inseonglee: @aradi `nSpin` changes to 2 in the `REKS_init` routine. You can find this at line 654 of reksvar.F90 file.

prog/dftb+/lib_dftbplus/initprogram.F90
inseonglee: @aradi I just imitated the initialization of LinResp part for REKS case. I will try to factor this part.

inseonglee:
@aradi I answered your questions, and I will make tutorials for REKS according to your comments.

doc/dftb+/manual/dftbp.tex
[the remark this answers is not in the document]
inseonglee: Done with commit 7485acf.

doc/dftb+/manual/dftbp.tex
```diff
@@ -4281,6 +4431,22 @@ \subsection{XREST.DAT}
...
 Dipole moment of the specified excited state in units of Debye.
 
+\section{REKS results files}
+\label{sec:reks_files}
+Several files are produced during REKS calculations depending on the
+particular settings from section~\ref{sec:dftbp.REKS}.
+
+
+\subsection{tdp.dat}
+\index{tdp.dat.reks}
```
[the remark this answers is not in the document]
inseonglee: Done with commit 7485acf.

prog/dftb+/lib_dftb/externalcharges.F90
```diff
@@ -461,4 +464,18 @@ subroutine getElStatPotentialCluster(this, env, locations, V, epsSoften)
   end subroutine getElStatPotentialCluster
 
 
+  !> Copy Q * inverse R contribution for the point charges
+  subroutine copyInvRvec(this, QinvR)
```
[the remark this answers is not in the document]
inseonglee: Done with commit 315c266.

prog/dftb+/lib_dftbplus/inputdata.F90
```diff
@@ -29,6 +29,7 @@ module dftbp_inputdata
 #:endif
   use dftbp_pmlocalisation, only : TPipekMezeyInp
   use dftbp_elstatpot, only : TElStatPotentialsInp
+  use dftbp_reksvar, only : TReksIni
```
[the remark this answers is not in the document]
inseonglee: @aradi I have one problem for importing reks modules with `dftbp_reks`. There is no problem for inputdata.F90 and initprogram.F90, but circular dependency appears for mainio.F90 file. Since `dftbp_reksen`, `dftbp_reksinterface` and `dftbp_reksproperty` modules shows `use mainio`, thus the conflict between mainio and dftbp_reks occurs when I change `use dftbp_reksvar` in mainio.F90 to `dftbp_reks`.

prog/dftb+/lib_reks/reks.F90
```diff
@@ -0,0 +1,60 @@
...
+
+  !> dftbp_reksfon module used in main.F90
+  public :: optimizeFONs
+
+  !> dftbp_reksinterface modules used in main.F90
+  public :: getReksGradients, getReksGradProperties
+  public :: getReksStress
+
+  !> dftbp_rekspreoperty modules used in main.F90
+  public :: getUnrelaxedDMandTDP
```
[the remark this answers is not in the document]
inseonglee: done with commit 9da09c0.

prog/dftb+/lib_reks/reks.F90
```diff
@@ -0,0 +1,60 @@
...
+  public :: qmExpandL!, udExpandL
+
+  !> dftbp_reksen modules used in main.F90
+  public :: constructMicrostates, calcWeights
+  public :: activeOrbSwap, getFilling, calcSaReksEnergy
+  public :: getFockandDiag, guessNewEigvecs
+  public :: adjustEigenval, solveSecularEqn
+
+  !> dftbp_reksfon module used in main.F90
+  public :: optimizeFONs
```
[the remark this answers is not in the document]
inseonglee: done with commit 9da09c0.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -6251,4 +6255,151 @@ subroutine parseRangeSeparated(node, input)
...
+    !> Target SSR state
+    call getChildValue(node, "TargetState", ctrl%reksIni%rstate, default=1)
+    !> Target microstate
+    call getChildValue(node, "TargetStateL", ctrl%reksIni%Lstate, default=0)
+
+    !> Initial guess for eigenvectors in REKS
+    !> 1: diagonalize H0, 2: read external file, 'eigenvec.bin'
+    call getChildValue(node, "InitialGuess", ctrl%reksIni%guess, default=1)
+    !> Maximum iteration used in FON optimization
+    call getChildValue(node, "FONmaxIter", ctrl%reksIni%FONmaxIter, default=20)
```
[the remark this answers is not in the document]
inseonglee: done with commit 9da09c0.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -6251,4 +6255,151 @@ subroutine parseRangeSeparated(node, input)
...
+
+    !> Minimized energy functional; 1: PPS, 2: (PPS+OSS)/2
+    call getChildValue(node, "EnergyFunctional", ctrl%reksIni%Efunction, default=2)
+    !> Calculated energy states in SA-REKS
+    !> 1: calculate states included in 'EnergyFuncitonal'
+    !> 2: calculate all possible states
+    call getChildValue(node, "EnergyLevel", ctrl%reksIni%Elevel, default=1)
+    !> Calculate SSR state (SI term is included)
+    !> 1: calculate SSR state, 0: calculate SA-REKS state
+    call getChildValue(node, "UseSSRstate", ctrl%reksIni%useSSR, default=1)
```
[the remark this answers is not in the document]
inseonglee: done with commit 9da09c0.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -200,6 +200,10 @@ subroutine parseHsdTree(hsdTree, input, parserFlags)
         & allowEmptyValue=.true., dummyValue=.true.)
     call readExcited(child, input%ctrl)
 
+    call getChildValue(root, "REKS", dummy, "", child=child, list=.true., &
+        & allowEmptyValue=.true., dummyValue=.true.)
+    call readREKS(child, input%ctrl, input%geom)
```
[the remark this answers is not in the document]
inseonglee: done with commit 9da09c0.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -200,6 +200,10 @@ subroutine parseHsdTree(hsdTree, input, parserFlags)
         & allowEmptyValue=.true., dummyValue=.true.)
     call readExcited(child, input%ctrl)
 
+    call getChildValue(root, "REKS", dummy, "", child=child, list=.true., &
```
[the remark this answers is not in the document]
inseonglee: done with commit 9da09c0.

prog/dftb+/lib_dftbplus/initprogram.F90
inseonglee: Done with commit c2bdedc.

doc/dftb+/manual/dftbp.tex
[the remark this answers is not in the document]
inseonglee: @aradi @bhourahine For the naming of REKS input, I have one idea for this. General form is followed below.

`Energy = {
  Functional = {PPS OSS}
  Level = Same
}`

Here, `Functional` has the elements that we want to include into averaged functional in SA-REKS (This is same as `EnergyFunctional`). `Functional` can have following elements (`{PPS}`, `{PPS OSS}` or `{All}`) for SSR(2,2) case. The possible choice can be extended for SSR(4,4) case (`{PPS}`, `{PPS OSS1 OSS2}`, `{PPS DSPS}`, '{All}'). Thus, `Functional` determines which functional is included in the minimized functional in SA-REKS.

After that, we can decide the states to be included in SI-SA-REKS using `Level` variable. Here, there is two choices. One is same state used in SA-REKS and the other is all states. The role of variable `Level` is same as `EnergyLevel`. `Level` reads string (`Same` or `All`), not `1` or `2`.

For this case, I am not sure whether the parser routine can read only string as variable.
In addition, the variables such as `TargetState` related to energy calculation can be added into this `Energy` block.

I hope this explanation is clear to yours.

prog/dftb+/lib_dftbplus/parser.F90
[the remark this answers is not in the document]
inseonglee: I already commented upper conservation, please see it.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -6251,4 +6255,151 @@ subroutine parseRangeSeparated(node, input)
...
+    integer :: nAt, iAt
+    real(dp), allocatable :: tmpTuning(:,:)
+
+    call getChildValue(node, "SpinTuning", value1, "", child=child, &
+        & modifier=modifier, allowEmptyValue=.true.)
+    call getNodeName2(value1, buffer)
+    if (char(buffer) == "") then
+      ! no 'SpinTuning' block in REKS input
+      allocate(ctrl%reksIni%Tuning(nType))
+      do iAt = 1, nType
```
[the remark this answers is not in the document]
inseonglee: Done with commit 5c55634.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -6251,4 +6255,151 @@ subroutine parseRangeSeparated(node, input)
...
+    !> Tolerance used in calculation of gradient with PCG and CG
+    call getChildValue(node, "GradientTolerance", ctrl%reksIni%Glimit, default=1.0E-8_dp)
+
+    !> Calculate relaxed density of SSR or SA-REKS state
+    call getChildValue(node, "RelaxedDensity", ctrl%reksIni%tRD, default=.false.)
+    !> Calculate nonadiabatic coupling vectors
+    call getChildValue(node, "NonAdiabaticCoupling", ctrl%reksIni%tNAC, default=.false.)
+
+    !> Print level in standard output file
+    call getChildValue(node, "PrintLevel", ctrl%reksIni%Plevel, default=1)
```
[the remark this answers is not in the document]
inseonglee: Done with 5c55634 and 070684d.

prog/dftb+/lib_dftb/coulomb.F90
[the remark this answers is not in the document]
aradi: Note the difference between
```
gVec = self%gLatPoint
```
and
```
gVec(:,:) = self%gLatPoint
```
Former automatically allocates the RHS to match the LHS (Fortran 2003), while latter assumes that the RHS has been already allocated.

prog/dftb+/lib_dftb/coulomb.F90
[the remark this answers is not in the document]
aradi: OK, I see. I think we should merge the two routines.

prog/dftb+/lib_dftbplus/inputdata.F90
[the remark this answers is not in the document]
aradi: I see. OK, then leave it for the moment as it is, I'll have a look, whether we can fix it somehow.

prog/dftb+/lib_dftbplus/parser.F90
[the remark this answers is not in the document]
aradi: OK, but why not using `TargetSsrState` and `TargetMicroState` as tags? Then it would be more descriptive, wouldn't it? (And you could even spare the two comment lines in the code :wink: )

aradi:
@inseonglee Thanks for the answers, changes so far. The tutorial would be amazing! I'll now continue to review the rest.

prog/dftb+/lib_dftbplus/inputdata.F90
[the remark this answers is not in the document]
inseonglee: Done with commit c2bdedc except mainio-related modules.

prog/dftb+/lib_dftbplus/parser.F90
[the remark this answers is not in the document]
inseonglee: @aradi I will follow your suggestion if the exact form for `Energy` and `Gradient` input block is determined. For the inputs, I will modify all changes at one time. (as well for test files)

prog/dftb+/lib_dftb/coulomb.F90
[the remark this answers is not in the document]
inseonglee: Done with commit 9cb0276.

prog/dftb+/lib_dftbplus/initprogram.F90
inseonglee: Done with commit a8cb284 to factor out this part, please see this commit carefully.

prog/dftb+/lib_dftbplus/initprogram.F90
inseonglee: Done with commit a8cb284 to factor out this part, please see this commit carefully.

inseonglee:
@aradi At the moment, I finished to modify REKS-related routines with your comments. Please see all commits and changes. The only remaining works are coulomb and parser routines, which mean combination of AddInvRPrime and change of input for REKS, respectively.

The additional works are to add tutorials for REKS and to add some test files for geometry (or lattice) optimization with REKS. These will be added after all REKS changes are complete.

inseonglee:
@bhourahine @aradi I added explanation part for the REKS representation. 6889 line in main.F90 file, you asked why I multiply two times. I imitated this part from the `convertToUpDownRepr` routine of main.F90. In this routine, the hamiltonian is multipled by 2, so I multipled 2 times in my routine.

For the naming of representation (my_ud, my_qm), I will change the name with other PR (When I changed the parser input, this will be fixed.)

inseonglee:
@aradi @bhourahine When this PR is merged? If all changes are finished, then we will solve the conflicts.

bhourahine:
There are some regression tests that do not work with all of the compilers that are tested in Bremen. We are investigating and once this is resolved, should be able to merge.

bhourahine:
> There are some regression tests that do not work with all of the compilers that are tested in Bremen. We are investigating and once this is resolved, should be able to merge.

If you have access to the intel19 compiler you can replicate some of the problems by compiling in debug mode, instead of the default release mode. A large fraction of the problems are due to unallocated arrays passed into routines that do not have the allocatable attribute set.

prog/dftb+/lib_dftbplus/main.F90
```diff
@@ -522,6 +522,13 @@ subroutine processGeometry(env, iGeoStep, iLatGeoStep, tWriteRestart, tStopDrive
               & iGeoStep, tStopScc, eigvecsReal, reks)
         end if
 
+      #:if WITH_TRANSPORT
+        ! Overrides input charges with uploaded contact charges
+        if (tUpload) then
+          call overrideContactCharges(qInput, chargeUp, transpar, qBlockIn, blockUp)
+        end if
+      #:endif
+
```
inseonglee: Currently, REKS is not compatible with transport calculation, so this part have to be removed.

inseonglee:
@bhourahine My intel compiler version is 19.1.0.166. According to your comments, I tried to compile REKS code with debug option, but I cannot find any problems that you said (unallocated arrays). The only thing that I found is warnings to show unused variables in many routines, thus I modify this warning for REKS routines.

Instead, I found one catastrophic error for waveplot compilation. The error message is following.
'**Internal compiler error: internal abort** Please report this error along with the circumstances in which it occurred in a Software Problem Report.  Note: File and line given may not be explicit cause of this error.'

Please check this.

aradi:
@inseonglee I have started to rework your version, but had to stop due to my starting teaching duties. Basically, whenever you pass a not allocated array to subroutine, where the corresponding dummy argument has not the `optional` or the `allocatable` flag, you are violating the standard. If I build your code with `-DCMAKE_BUILD_TYPE=Debug`, I get a lot of run-time errors due to this, also with ifort 19.0.5.281. (May be an important note: I have merged master into your branch first to make it up to date with current master...) If interested, you find the code in my personal fork in the `reksWithLC` branch.

inseonglee:
@aradi I also found the error for unallocated arrays when I run REKS calculation. I will try to solve this problem.

inseonglee:
@aradi @bhourahine Can I know current step for solving the problems? The unallocated errors can be solved by adding allocatable attributes (Maybe you already know this). If this is not the solution of the unallocated errors, please tell me. For now, I am trying to adding the allocatable attribute to the variables.

bhourahine:
> @aradi @bhourahine Can I know current step for solving the problems? The unallocated errors can be solved by adding allocatable attributes (Maybe you already know this). If this is not the solution of the unallocated errors, please tell me. For now, I am trying to adding the allocatable attribute to the variables.

Yes, that is exactly the solution for this as it allows the compiler to realise that the variable may not be allocated at call time.

inseonglee:
@aradi @bhourahine Here, I solved the errors caused from many unallocated arrays, please see commit 6c93d2e carefully. The current version of REKS code passes the test with intel19 compiler, so please check this code with other compilers.

bhourahine:
I will run some tests after bringing it up to date with the current master branch (as there have been several bug fixes and merging of the electrostatics for Born solvent models in the last few days).

bhourahine:
There are several variables that are used uninitialized, which causes problems with one compiler. My attempt to fix some of them (see below) leads to the code failing various REKS test cases as the conjugate gradient does not succeed. I am not sure if the Fa and Fc should be set to zero, or intent inout (the calling routine suggests the second case).

diff --git a/prog/dftb+/lib_reks/reksen.F90 b/prog/dftb+/lib_reks/reksen.F90
index 2022f60f..c547190a 100644
--- a/prog/dftb+/lib_reks/reksen.F90
+++ b/prog/dftb+/lib_reks/reksen.F90
@@ -391,7 +391,7 @@ module dftbp_reksen
 
   !> Calculate filling of each microstate in REKS(2,2)
   subroutine getFillingL22_(Nc, fillingL)
-    
+
     !> Number of core orbitals
     integer, intent(in) :: Nc
 
@@ -402,6 +402,7 @@ module dftbp_reksen
 
     nSpin = size(fillingL,dim=2)
     Lmax = size(fillingL,dim=3)
+    fillingL(:,:,:) = 0.0_dp
 
     ! Filling of core orbitals
     do iL = 1, Lmax
@@ -830,7 +831,7 @@ module dftbp_reksen
   subroutine fockFcAO_(hamSqr, weight, Lpaired, iL, Fc)
 
     !> dense fock matrix for core orbitals
-    real(dp), intent(out) :: Fc(:,:)
+    real(dp), intent(inout) :: Fc(:,:)
 
     !> Dense Hamiltonian matrix for each microstate
     real(dp), intent(in) :: hamSqr(:,:)
@@ -865,7 +866,7 @@ module dftbp_reksen
       & Lpaired, iL, Fa)
 
     !> dense fock matrix for active orbitals
-    real(dp), intent(out) :: Fa(:,:,:)
+    real(dp), intent(inout) :: Fa(:,:,:)
 
     !> Dense Hamiltonian matrix for each microstate
     real(dp), intent(in) :: hamSqr(:,:)
diff --git a/prog/dftb+/lib_reks/reksgrad.F90 b/prog/dftb+/lib_reks/reksgrad.F90
index 99108c99..f3c34a23 100644
--- a/prog/dftb+/lib_reks/reksgrad.F90
+++ b/prog/dftb+/lib_reks/reksgrad.F90
@@ -4354,6 +4354,7 @@ module dftbp_reksgrad
     integer :: iL, Lmax, tmpL1, tmpL2
 
     Lmax = size(ZmatL,dim=3)
+    ZmatL(:,:,:) = 0.0_dp
 
     do iL = 1, Lmax
       if (iL <= Lpaired) then

inseonglee:
@bhourahine The variable Fc, Fa and ZmatL should not be set to zero since these variables are updated continuously within loop. I changed intent from out to inout for Fc, Fa and ZmatL, but I am not sure why the routine for calculating fillingL shows problem.

bhourahine:
The unoccupied states are not set to have a filling value, so if a compiler fails on uninitialized values it will stop there (or even worse, random memory content may be used).

inseonglee:
If there is further problems in the code, please let me know. I will solve them.

bhourahine:
Yes, there are other uninitialized variables being used. Can you set one of the compilers you have access to such that is will test for this? Some of these either have a test for variables directly, or can set undefined variables to NaN.

inseonglee:
@bhourahine I have only intel19 compilers so I do not test with other compilers. Is this right that fillingL is the example for uninitialized variables in the code? Originally, all allocatable variables is allocated in REKS_init routines, but in this case, I think that the allocated changes to false due to the intent out (not inout). If this is right, I will check the code such as fillingL.

bhourahine:
> @bhourahine I have only intel19 compilers so I do not test with other compilers. Is this right that fillingL is the example for uninitialized variables in the code? Originally, all allocatable variables is allocated in REKS_init routines, but in this case, I think that the allocated changes to false due to the intent out (not inout). If this is right, I will check the code such as fillingL.

It is more that there are some variables that are not initialized to contain a value before being used, either by not being set or by being intent(out) where the initialization is in the calling routine (so is lost on passing into the routine). I think most of the relevant variables are now covered by commit [a3b820ac](https://github.com/inseonglee/dftbplus/commit/a3b820ace98c487a3b9a5f167e97fdef9795cb17), can you please check whether this breaks anything?

These are only the cases that appear for the existing tests, so eventually we need to look at coverage tests to try and identify any others (this should be at some point after the merge).

prog/dftb+/lib_reks/reksproperty.F90
```diff
@@ -670,17 +670,18 @@ subroutine getUnrelaxedTDM22_(eigenvecs, FONs, Nc, nstates, rhoXdel)
     a = Nc + 1
     b = Nc + 2
 
+    rhoXdel(:,:,:) = 0.0_dp
```
inseonglee, line 673: Why is this additional zeroing needed? I think that modification from intent out to inout is enough in this routine.

prog/dftb+/lib_reks/reksinterface.F90
```diff
@@ -292,7 +292,9 @@ subroutine getReksGradients(env, denseDesc, sccCalc, rangeSep, dispersion, &
 
     end if
 
-    call setReksGradients_(derivs, self)
+    if (allocated(self%SSRgrad)) then
+      call setReksGradients_(derivs, self)
+    end if
```
inseonglee, line 297: self%SSRgrad is always allocated when we calculate the gradient. Why is the condition `allocated(self%SSTgrad)` added?

prog/dftb+/lib_reks/reksgrad.F90
```diff
@@ -2426,6 +2423,7 @@ subroutine getRmatSp(env, denseDesc, neighbourList, nNeighbourSK, &
         real(dp), allocatable :: tmpMat(:,:)
         integer :: iAt1, iAt2, nOrb, LmaxR, mu, nu, iL
 
+        RmatSpL(:,:) = 0.0_dp
```
inseonglee, line 2426: Why is this additional zeroing needed? I think that modification from intent out to inout is enough in this routine.

prog/dftb+/lib_reks/reksgrad.F90
```diff
@@ -2311,9 +2310,7 @@ subroutine RTshift(env, sccCalc, denseDesc, neighbourList, nNeighbourSK, &
     end if
 
     deriv1(:,:,:) = 0.0_dp
-    if (tNAC) then
-      deriv2(:,:,:) = 0.0_dp
-    end if
+    deriv2(:,:,:) = 0.0_dp
```
inseonglee, line 2313: Here, the size of `deriv2` is not too large, thus it do not increase memory allocation so much although we always allocate the `deriv2`. I think that it do not cause the problems when we always allocate the `deriv2`, but should we remove the if condition for this line?

inseonglee:
@bhourahine Can you show me the example or case that the code shows errors? With intel19 compiler, I cannot find any errors regardless of applying the commit a3b820a.

prog/dftb+/lib_reks/reksproperty.F90
bhourahine: The getUnrelaxedTDM22_ routine is intent(out) :: rhoXdel(:,:,:), so either changing to inout or initializing to zero inside works. What do you prefer?

inseonglee: Please change intent(out) to intent(inout).

prog/dftb+/lib_reks/reksinterface.F90
bhourahine: Its not always allocated, the HBI+H2O_periodic_stress case, for example, tries to access this variable without it being allocated with some compilers. Possibly its an intent() issue somewhere, but it returns the correct forces with if test in place.

prog/dftb+/lib_reks/reksgrad.F90
bhourahine: This was causing a problem for cases where tNAC was false but the various later changes of (out) to (inout) removed the problem, so the if could be returned. 
However it is now more consistent with the the other variables which initialise to zero at higher levels and, as you mention, is always allocated.

prog/dftb+/lib_reks/reksinterface.F90
bhourahine: The Thymine_single case is easier to test (much faster to run), and for the intel compiler with the debug setting, in main.F90 allocated(reks%SSRgrad) is false at all of the places I have tested it.

inseonglee: I modified this line, please see commit 6dc8ae2, the `setReksGradients_` routine moved to inner part of if condition (`Efunction /= 1`) so that it is not needed to access this routine for single-state REKS.

inseonglee:
@aradi @bhourahine At the moment, I am fine for the current code (6dc8ae2). Now I will try to compile and run the code with NAG compiler according to your comments, and I will fix the codes if there is more errors in the code.

inseonglee:
@aradi @bhourahine Now I finished to add these 3 commits, please check ca3b972 and run the test calculations with other compilers. If all calculations run successfully, it is fine that this branch will be merged.

bhourahine:
All checks pass successfully.

PR #332, opened 2019-12-03
Extension of transport to k-point distribution  

pecchia:
Most of the changes affect negf_int.F90   
Mainly  adding mpi_reduce over the ks communicator and debugging IO issues. 
negf_int is currently providing IO of k-resolved datafiles.


prog/dftb+/lib_extlibs/negf_int.F90
```diff
@@ -993,12 +994,13 @@ end subroutine negf_current
...
   !> Calculates density matrix with Green's functions
 #:if WITH_MPI
-  subroutine calcdensity_green(iSCCIter, mpicomm, groupKS, ham, over, iNeighbor, nNeighbor,&
+  subroutine calcdensity_green(iSCCIter, enecomm, kscomm, groupKS, ham, over, iNeighbor, nNeighbor,&
       & iAtomStart, iPair, img2CentCell, iCellVec, cellVec, orb, kPoints, kWeights, mu, rho, Eband,&
       & Ef, E0, TS)
 
     !> MPI communicator
-    type(mpifx_comm), intent(in) :: mpicomm
+    type(mpifx_comm), intent(in) :: enecomm
```
bhourahine: What does the enecomm name mean? It suggests a communicator between energies, but looking at the call is the intra-group communicator for a specific (k,s) matrix.

prog/dftb+/lib_extlibs/negf_int.F90
```diff
@@ -1360,24 +1366,32 @@ subroutine calc_current(groupKS, ham, over, iNeighbor, nNeighbor, iAtomStart, iP
...
     unitOfCurrent%name = "A"
 
+    ! groupKS is local, hence nKS il local
     nKS = size(groupKS, dim=2)
+#:if WITH_MPI
+    call mpifx_allreduce(kscomm, nKS, nTotKS, MPI_SUM)
+#:else
+    nTotKS = nKS
+#:endif
+    nS = nTotKS/size(kpoints, 2)
```
bhourahine: Is there any reason not to use nS = size(ham,dim=2); nTotKS = nS * size(kpoints, dim=2)? This would remove the need  for the reduce operation.

prog/dftb+/lib_extlibs/negf_int.F90
pecchia: Exactly, it is a communicator dealing with energy points distribution for a given k,s.
On this communicator I can make reduce operations across energy points, hence this name

prog/dftb+/lib_extlibs/negf_int.F90
pecchia: Yes. I think it is easier. Provided I add a check: nS=size(ham,2); if (nS>2)  nS=1

PR #333, opened 2019-12-07
ELPA via the csc interface

bhourahine:
Also re-enabled PEXSI with k-points as a warning instead of an
error.


bhourahine:
Review #340 first, as that effects a change here.

bhourahine:
#340 merged in, Relevant autotest unchanged (gapped system, so filling methods irrelevant). Ready to review.

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -4140,13 +4140,13 @@ subroutine ensureSolverCompatibility(iSolver, tSpin, kPoints, tForces, parallelO
 
     ! Temporary error test for PEXSI bug (July 2019)
     if (iSolver == electronicSolverTypes%pexsi .and. any(kPoints /= 0.0_dp)) then
-      call error("A temporary bug prevents correct evaluation with PEXSI at general k-points.&
+      call warning("A temporary PEXSI bug may prevent correct evaluation at general k-points.&
```
awvwgk: It there a bug tracker for PEXSI we could link in the manual?

bhourahine: Unfortunately there doesn't seem to be one at the moment. The ELSI [issues](https://git.elsi-interchange.org/elsi-devel/elsi-interface/issues) page is the nearest thing, but that is downstream from PEXSI.

PR #335, opened 2019-12-10
Delta dftb clean

deshaym:
Introduces optional spin-purified excited states via non-Aufbau occupation. Optional spin purification and a ground-state initial guess as well.
(We might have built this with an earlier version of the code, but I don't think there will be any issues.)


prog/dftb+/lib_dftb/etemp.F90
```diff
@@ -464,8 +514,10 @@ subroutine electronFill(Eband, filling, TS, E0, Ef, eigenvals, kT, distrib, kWei
 #:else
             filling(j, i, iSpin) = 1.0_dp / (1.0_dp + exp(x))
 #:endif
-            if (filling(j, i, iSpin) <= elecTol) then
+            if (tNonAufbau .and. j/=1 .and. ((filling(j, i, iSpin)+filling(j-1, i, iSpin))) <= elecTol) then
```
bhourahine: When j==1 in the loop, the j-1 term goes below the lower limit of the filling array (compiling with DEBUG := 2 in the make.config file finds this).

bhourahine: the slightly hacky fix is filling(j-1, i, iSpin) =>  filling(max(j-1,1), 1, iSpin)
It might be slightly clearer to split the test as 
if (tNonAufbau .and. j/=1) then
  if ( filling(j, i, iSpin)+filling(j-1, i, iSpin) <= elecTol) then

prog/dftb+/lib_dftb/densitymatrix.F90
```diff
@@ -71,20 +71,38 @@ subroutine fullDensityMatrix_real(dm, eigenvecs, filling)
...
     dm(:,:) = 0.0_dp
     do ii =  size(filling), 1, -1
       nLevels = ii
       if (abs(filling(ii)) >= epsilon(1.0_dp)) then
         exit
       end if
     end do
+
+    ! determine whether fillings are non-Aufbau
+    mixIndx=nLevels-1 !sets to HOMO-1
```
bhourahine: In 1 electron systems (H2+) this will cause problems below

prog/dftb+/lib_dftb/etemp.F90
deshaym: Either change works equally well, do you have a preference? I prefer the more compact hacky fix, but either way is fine by me!

prog/dftb+/lib_dftb/densitymatrix.F90
deshaym: I can wrap that in a statement  "if (nLevels/=1) then..." to exclude one-electron systems in the ground state. The issue is testing a zero index in the filling vector correct?

bhourahine: It might be worth working out a solution that could be used for 1 electron cases. I've made a PR against your branch that just uses the max() construct.

prog/dftb+/lib_dftb/etemp.F90
bhourahine: The max() form is OK for me (Bálint may disagree...), but it doesn't cover the 1 electron case. it might be worth testing for the 1 electron case in initprogram.F90 and stopping with an error if it doesn't fill correctly in several places.

prog/dftb+/lib_dftb/densitymatrix.F90
deshaym: If (H2+) were excited I think that nLevels==2, from what I can tell nLevels is determined by the first non-zero filling by reading the filling vector from the top. This little hack is only useful in saving the empty MO in the mixed state (the previous HOMO that has been excited from).

prog/dftb+/lib_dftb/etemp.F90
deshaym: You make a good point, we have assumed a closed shell ground state system with an even number of electrons, but we have not coded that assumption in yet. I will get right on that! I will enforce that restriction in initprogram.F90

prog/dftb+/lib_dftb/etemp.F90
```diff
@@ -464,7 +514,9 @@ subroutine electronFill(Eband, filling, TS, E0, Ef, eigenvals, kT, distrib, kWei
 #:else
             filling(j, i, iSpin) = 1.0_dp / (1.0_dp + exp(x))
 #:endif
-            if (filling(j, i, iSpin) <= elecTol) then
+            if (tNonAufbau .and. j/=1 .and. ((filling(max(j-1,1), i, iSpin)+filling(j-1, i, iSpin))) <= elecTol) then
```
bhourahine: This line might still have a bug, as the loop runs j=1,... and filling(j-1, ) is referenced by some compilers even if the j/=1 test fails. Also, most of the time max(j-1,1) = j-i so the same level is tested twice.
I think the line  should probably read
if (tNonAufbau .and. j/=1 .and. ((filling(max(j-1,1), i, iSpin)+filling(j, i, iSpin))) <= elecTol) then

prog/dftb+/lib_dftb/etemp.F90
```diff
@@ -449,6 +481,24 @@ subroutine electronFill(Eband, filling, TS, E0, Ef, eigenvals, kT, distrib, kWei
...
       E0(:) = (real(MPorder + 1,dp) * (Eband - TS) + Eband) / real(MPorder + 2, dp)
     else
+! TI-DFTB Non-Aufbau filling routine
+      if (tNonAufbau) then
+        allocate(tmpMtx(size(eigenvals, dim=1),kpts,size(eigenvals, dim=3)))
+        tmpMtx=eigenvals
+        do iSpin = 1, size(eigenvals, dim=3)
+          do i = 1, kpts
+            if (iDet == 1 .and. tSpinPurify .and. iS==1) then
+              eigenvals(int(nEl(iS)) + 1, i, iSpin)=eigenvals(int(nEl(iS)), i, iSpin)
```
bhourahine: Shouldn't eigenvals(int(nEl(iS)), i, iSpin) then be over-written with the original value of eigenvals(int(nEl(iS)) + 1, i, iSpin) ? Also, any reason for using int (this would behave like ceiling() as nEl >= 0, which might be clearer) or do you actually need nint() which would be robust to numerical noise in nEl?

prog/dftb+/lib_dftb/etemp.F90
```diff
@@ -449,6 +481,24 @@ subroutine electronFill(Eband, filling, TS, E0, Ef, eigenvals, kT, distrib, kWei
...
+            else if (iDet == 1 .and. tSpinPurify .and. iS==2) then
+              eigenvals(int(nEl(iS)), i, iSpin)=eigenvals(int(nEl(iS)) + 1, i, iSpin)
```
bhourahine: Same comment as above about replacing the old value

prog/dftb+/lib_dftbplus/main.F90
```diff
@@ -3142,12 +3236,13 @@ subroutine getFillingsAndBandEnergies(eigvals, nElectrons, nSpinBlocks, tempElec
...
         call electronFill(Eband(iS:iS), fillings(:,:,iS:iS), TS(iS:iS), E0(iS:iS), Ef(iS),&
-            & eigvals(:,:,iS:iS), tempElec, iDistribFn, kWeights)
+            & eigvals(:,:,iS:iS), tempElec, iDistribFn, kWeights, tNonAufbau, &
+            & tSpinPurify, iDet, nEl, iS)
       end do
     else if (nSpinHams == 2 .and. tSpinSharedEf) then
       ! Common Fermi level across two colinear spin channels
       call Efilling(Eband, Ef(1), TS, E0, fillings, eigvals, sum(nElecFill), tempElec, kWeights,&
-          & iDistribFn)
+          & iDistribFn, tNonAufbau, tSpinPurify, iDet, nEl, iS)
```
bhourahine: iS is not set in this call

deshaym: iS is needed by us in electronFill, which is called by this subroutine I thought

aradi (approved):
Good to go apart of the minor suggestions. Also, I've pushed two commits o remove macro constructs and simplify code a bit.

bhourahine:
@aradi somewhere after 54fca6a, dptools broke for the osx CI.

aradi:
I've checked. Even current main branch fails on GitHubs OSX image during NumPy import... So the issue is unrelated to the changes here.
```
   import numpy as np
  File "/Users/runner/hostedtoolcache/Python/3.9.0/x64/lib/python3.9/site-packages/numpy/__init__.py", line 286, in <module>
    raise RuntimeError(msg)
RuntimeError: Polyfit sanity test emitted a warning, most likely due to using a buggy Accelerate backend. If you compiled yourself, see site.cfg.example for information. Otherwise report this to the vendor that provided NumPy.
RankWarning: Polyfit may be poorly conditioned
```

PR #347, opened 2020-01-10
Add error handling for missing forces

vanderhe:
Prevent unexpected behavior in the case of communication via the ctypes interface, if the calculation of forces is not explicitly set in the input file.


bhourahine:
Thanks. Now merged.

PR #348, opened 2020-01-12
Particle-Particle Random Phase Approximation

adrielDom:
calculates the excitation energies of a N-electron system with singlet or triplet ground state by solving the pp-RPA equations for the (N-2)-electron system.


bhourahine:
Can you add at least one example input to the test system in test/prog/dftb+/ ? This will be needed for regression testing.

bhourahine:
Can you add a description of the keywords to the manual and also a table of the U values (as an appendix)?
Are there other changes that you want to make before this is reviewed?

adrielDom:
Working on the manual. 
No further changes will be made at the moment.

adrielDom:
Dear @bhourahine , I added the pp-RPA part to the manual.
Please proceed with review!

doc/dftb+/manual/dftbp.tex
```diff
@@ -3756,6 +3757,42 @@ \subsection{Casida}
...
     
   \end{description}
+  
+  
+\subsection{Pp-RPA}
+\label{sec:dftbp.pprpa}
+
+This tag contains the specifications for the calculation of excitation energies using the particle-particle random phase approximation (pp-RPA)~\cite{Yang2017}. This approach, unlike the time-dependent DFTB formalism, allows the computation of double and charge-transfer transitions. However, it has the limitation, that the computed excitations has to involve, at least partially, the highest occupied molecular orbital (HOMO) of the system. 
+
+For the computation of the excitation energies of a neutral $N$-electron system, one needs to set a DFTB ground state calculation for the two-electron deficient ($N-2$) system ( \kw{Charge} = 2.0 ). Please note that if \kw{Charge} is set to 0.0 (default value), the obtained transition energies will correspond to the negatively charged ($N+2$)-electron system.
```
bhourahine, line 3767: Any reason not to change the default charge behaviour in the parser/initialisation such that a requested charge of 0 gives internally the correct +2 reference system?

prog/dftb+/lib_timedep/pprpa.F90
```diff
@@ -0,0 +1,716 @@
...
+    end if
+
+    ! Overlap times wave function coefficients - most routines in DFTB+ use lower triangle (would
+    ! remove the need to symmetrize the overlap and ground state density matrix in the main code if
+    ! this could be used everywhere in these routines)
+    do iSpin = 1, nSpin
+      call symm(stimc(:,:,iSpin), "L", SSqr, grndEigVecs(:,:,iSpin))
+    end do
+
+    nocc = nint(rnel) / 2
```
bhourahine, line 206: Needs a test that this really is a closed shell system

bhourahine:
> At the moment the Tamm Dancoff test case is failing if the code is compiled in Debug mode (instead of the default Release mode, see config.cmake for the setting). Its not clear so far what is causing the problem, so any suggestions for fixing this would be very useful.

It looks like tightening the SCC tolerance makes this case behave. Not sure why this particular case and not the others though.

doc/dftb+/manual/dftbp.tex
adrielDom: I think this could lead to confusions as the ground state output files would correspond to a +2 charged system while the charge specified in the input would be 0. What do you think?

adrielDom: Otherwise the code would have to perform two calculations: one for the zero-charged ground state output and then a second one for the 2+ charged system needed for ppRPA.

bhourahine: Does it ever make sense to obtain spectra for anything other than the neutral closed shell systems (by, as you say, simulating the +2 system)? If not, we can remove the the charge choice  option for ppRPA (and explain, with reference to your paper about the reference system calculation).

adrielDom: it might make sense to calculate for charged closed-shell molecules, but I agree that the most requested calculations will probably be for neutral systems. I think we should at least provide this option: let the user set a nonzero charge, while the actual charge used for the calculations be the indicated in the input plus 2. 
what do you think? 
How do we proceed with the changes?

prog/dftb+/lib_timedep/pprpa.F90
bhourahine: Anything other than +2 and spin free then should give warnings (call warn()), but if its useful to allow other cases the code should not stop.

Issue #338, opened 2019-12-20
WITH_OMP option in make.config


inseonglee:
Dear DFTB+ developers,

I have a question about WITH_OMP option in compilation process.
In my understanding, when I use cmake I can decide WITH_OMP option in config.cmake file so that the dftb+ code considers WITH_OMP may become true. (I do not test and use with cmake)
But when I use make with make.config, WITH_OMP becomes false even though I use OpenMP parallelization with '-qopenmp' flag in ifort.

Due to this, it seems to be hard to use WITH_OMP option into the code development.
Can anyone explain about this?
If there is a wrong point for me, please tell me anything.

Best Regards
In Seong Lee

bhourahine:
This behaviour should be fixed in pull request #328  (commit 4238dd80 ).

inseonglee:
@bhourahine I tested WITH_OMP option with new commit (8a7ffcf). This commit includes current master branch as well as commit 4238dd8.

In the code, the following part is added for the test purpose in initprogram.F90 file.
`#:if WITH_OMP
  write(*,*) "omp run"
#:else
  write(*,*) "omp not run"
#:endif`
The flag of the code includes '-qopenmp' with ifort compiler.
But the test result shows "omp not run" sentence.

I think that this is caused from the make.config file. It means that current make.config file do not include WITH_OMP option unlike config.cmake file.
After I add WITH_OMP := 1 in the make.config file, the test result shows "omp run" sentence.
Thus, my conclusion is that WITH_OMP option should be added to make.config file for correct use of WITH_OMP option in the code development.

bhourahine:
Thanks for pointing this out. See #354 for your suggested fix.

inseonglee:
This behavior is now fixed (9551096), so I close this issue.

PR #354, opened 2020-01-15
Added openMP flag for makefile system

bhourahine:
Matching makefile bugfix for the cmake system in #328 as raised in #338

### DIFF
--- a/make.config
+++ b/make.config
@@ -20,6 +20,10 @@ WITH_GPU := 0
 # Whether DFTB+ should support MPI-parallelism
 WITH_MPI := 0
 
+# Whether OpenMP thread parallisation should be enabled, note that the
+# compiler flags for the make.arch file also need to support openMP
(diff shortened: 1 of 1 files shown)


inseonglee:
The option "WITH_OPENMP" is wrong variable. "WITH_OMP" is correct name of the variable. In the common.fypp, it regards `#:set WITH_OMP = defined('WITH_OMP')`.

Also, I think the make.arch files in the ./sys/ directory may be changed with respect to WITH_OMP.
For non-MPI setting, current FXXOPT is `-O2 -ip -standard-semantics -heap-arrays 10` in intel compiler case.
Thus, I think that the following parts have to be added.
`ifeq ($(strip $(WITH_OMP)),1)`
`FXXOPT += -qopenmp`
`endif`

Also, the flag for MPI case can be separated with WITH_OMP option.

PR #358, opened 2020-01-25
Build DFTB+ on Travis-CI

awvwgk:
[Build Status](https://travis-ci.com/awvwgk/dftbplus)

In case there is interest to build DFTB+ for every PR and commit on Travis-CI I setup a build on Ubuntu Bionic using GCC7 (with and without MPI). The build and the test suite take around 10 minutes on the CI server.

Things to discuss:

- Should DFTB+ also be build on OSX or Windows?
- Should different versions of GCC be tested?
  Does not work with current tool-chain file
- Should mpich as MPI provider be tested?
- Which optional features of DFTB+ should be included?
  from the build file: ELSI/PEXSI, ARPACK, transport, DFT-D3 (done), socket integration, API
- Should different compiler vendors be tested?
  Intel can be used on Travis-CI, but requires some work to setup
- Should another service be preferred?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,31 @@
+os: linux
+dist: bionic
+
+language: python
+python: 3.7
+
+
+env:
+ - WITH_MPI=false
+ - WITH_MPI=true
+
+addons:
+   apt:
+      packages:
+      - cmake
+      - gfortran
+      - libblas-dev
+      - liblapack-dev
+      - libopenmpi-dev
+      - libscalapack-openmpi-dev
+
+install:
+- git submodule update --init --recursive
+- echo "y" | ./utils/get_opt_externals ALL
+
+script:
+- mkdir _build
+- cd _build
+- cmake -DWITH_DFTD3=true -DWITH_MPI=${WITH_MPI} -DCMAKE_TOOLCHAIN_FILE=../sys/gnu.cmake ..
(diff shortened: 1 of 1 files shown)


aradi:
Yes, Travis integration is something we could consider.

Currently we have an internal buildbot based nightly testing. It tests the master branch, for various compiler (GNU, Intel, NAG), MPI-framework (MPICH, OpenMPI), MPI-process number (1, 2, 4) and thread-number (1, 2, 4) combination. I think, that variety won't be possible with Travis (the full test runs several hours..).

We plan to move the testing to a publicly available server next month (we will then be able to test GPU-functionality as well). So in theory we could even connect it with GitHub via a web-hook, but I  am just not sure, whether it makes sense to run the full test-suite whenever a commit had been made on a PR.

What do you think about using Travis for a quick test (one compiler, one MPI-framework) and integrate it into GitHubs CI framework, and keep the buildbot based nightly build with the full tests for master only?

awvwgk:
So the current setup in this PR would be a quick check for obvious errors, like non-compiling code and simple failing tests. So I will leave it that way.

If you set up Travis CI from https://travis-ci.com, you should get integration by a GH app, which provides usually a nice summary [like this](https://github.com/awvwgk/dftbplus/runs/409001483) for every commit.

PR #362, opened 2020-02-01
Support for xyz format

awvwgk:
Allows DFTB+ to read geometries in xyz format.

```
Geometry = xyzFormat {
24
caffeine C8H10N4O2
C          1.07317        0.04885       -0.07573
N          2.51365        0.01256       -0.07580
C          3.35199        1.09592       -0.07533
N          4.61898        0.73028       -0.07549
C          4.57907       -0.63144       -0.07531
C          3.30131       -1.10256       -0.07524
C          2.98068       -2.48687       -0.07377
O          1.82530       -2.90038       -0.07577
N          4.11440       -3.30433       -0.06936
C          5.45174       -2.85618       -0.07235
O          6.38934       -3.65965       -0.07232
N          5.66240       -1.47682       -0.07487
C          7.00947       -0.93648       -0.07524
C          3.92063       -4.74093       -0.06158
H          0.73398        1.08786       -0.07503
H          0.71239       -0.45698        0.82335
H          0.71240       -0.45580       -0.97549
H          2.99301        2.11762       -0.07478
H          7.76531       -1.72634       -0.07591
H          7.14864       -0.32182        0.81969
H          7.14802       -0.32076       -0.96953
H          2.86501       -5.02316       -0.05833
H          4.40233       -5.15920        0.82837
H          4.40017       -5.16929       -0.94780
}
```


prog/dftb+/lib_type/typegeometryhsd.F90
```diff
@@ -317,4 +318,129 @@ subroutine readTGeometryGen_help(node, geo, text)
...
+    integer :: ii, iSp, iTmp
+    real(dp) :: coords(3), rTmp, det
+    type(listString) :: speciesNames
+
+    ! Read first line of the xyz file: Number of atoms
+    iStart = 1
+    iEnd = nextLine(text, iStart)
+    call getNextToken(text(:iEnd), geo%nAtom, iStart, iErr)
+    call checkError(node, iErr, "Bad number of atoms.")
+    ! xyz files are always molecular, and we are not going to support extensions to this
```
bhourahine: The jmol variant format does support lattice vectors (probably several other hacked implementations as well), so that might be for a future extension to this routine.

awvwgk: My problem with all this flavours of xyz is that they are not well standardized. And usually you want a molecular calculation if you have a xyz file, so I don't see much use cases here.

awvwgk:
If you are happy with this implementation (especially my `nextLine` function), I will write a POSCAR reader as well.

prog/dftb+/lib_type/typegeometryhsd.F90
```diff
@@ -317,4 +318,129 @@ subroutine readTGeometryGen_help(node, geo, text)
...
+    allocate(geo%speciesNames(geo%nSpecies))
+    call asArray(speciesNames, geo%speciesNames)
+    call destruct(speciesNames)
+
+    if (geo%nSpecies /= maxval(geo%species) .or. minval(geo%species) /= 1) then
+      call detailedError(node, &
+          &"Nr. of species and nr. of specified elements do not match.")
+    end if
+
+    if (iStart <= len(text)) then
```
bhourahine: The original  ['standard'](http://www.ccl.net/chemistry/resources/messages/1996/10/21.005-dir/index.html) does have optional scalar and vector data after the 3 coordinates, so this probably needs to be handled.
DFTB+ output xyz can contains these fields where relevant, so would need user editing for re-starts. It does however weaken the possibilities for error checking (at least each coordinate line starts with characters).
In the longer term, parsing the atomic charges and vectors could be used to simplify restarts for MD & SCC.

awvwgk: In fact this is not a problem with this implementation, which comes along with the `nextLine` routine I introduced for this reader. Every token after the four columns is currently just ignored.

awvwgk:
Maybe a note on the implementation. The xyz format is defined in lines terminated by newline tokens, but since I only get the raw character buffer here and didn't knew how to search for newline tokens in HSD, I use the following approach:

```fortran
character(len=*) :: text
integer :: iStart, iEnd, iErr
iStart = 1
lpGetLines: do
  ! next two lines are wrapped by a function `newLine`, but shown here for clarity
  iEnd = index(text(iStart:), new_line(text)) + iStart - 1
  if (iEnd < iStart) iEnd = len(text)
  call getNextToken(text(:iEnd), ..., iStart, iErr)
  ... ! do something with this token
  iStart = iEnd + 1
end do lpGetLines
```

This allows reading the raw character buffer line by line and each line token by token. For the coordinate section, I decided to ignore everything that is trailing after the forth token in the line, to not interfere with additional columns.

prog/dftb+/lib_type/typegeometryhsd.F90
awvwgk: I added a test example including trailing scalars in every column.

awvwgk:
Looks like the GenFormat doesn't care about newlines at all, see #363.

PR #368, opened 2020-02-04
Added Get Stress Tensor to API

Pibemanden:
I made an interface function in the API of DFTB+ to get the stress tensor of the calculation.
The implementation is very close to that of the getGradients family of interfaces as I used those as a template for the implementation.


prog/dftb+/api/mm/dftbplus.h
```diff
@@ -207,6 +207,16 @@ void dftbp_get_energy(DftbPlus *instance, double *mermin_energy);
...
 
 
+/**
+ * Queries the stress tensor of the current periodic box
+ *
+ * \param[inout] instance Handler of the DFTB+ instance.
+ *
+ * \param[out] stresstensor Stress Tensor for the periodic box. Shape [3, 3]. Unit: Pascals.
+ */
+void dftbp_get_gradients(DftbPlus *instance, double *stresstensor);
```
awvwgk: `dftbp_get_stress_tensor` to match the `bind(C)` from `c_DftbPlus_getStressTensor`

Pibemanden: That is a good catch, I guess the compiler cannot catch all your copy paste errors after all.

This should be fixed in the current commit.

prog/dftb+/api/mm/capi.F90
```diff
@@ -283,6 +283,24 @@ subroutine c_DftbPlus_getGradients(handler, gradients) bind(C, name='dftbp_get_g
...
+    type(c_DftbPlus), intent(inout) :: handler
+
+    !> gradients, row major format
+    real(c_double), intent(out) :: stresstensor(3, 3)
+    
+    type(TDftbPlusC), pointer :: instance
+
+    call c_f_pointer(handler%instance, instance)
+
+    call instance%getStressTensor(stresstensor(:, :))
```
aradi: Please remove `(:, :)`. We use dummy indices only on the **left hand sides of assignments** (to indicate that no automatic allocation/reallocation occurs), but not otherwise, as the compiler may decide to create an unnecessary copy.

Pibemanden: Thank you for the comment I will fix this right away.

As for the test, I will try to see what I can do over the next couple of days.

Pibemanden: @aradi Just to make sure, it seems that currently the C tests are not run? Is that correct or is it just part of my current setup?

bhourahine: If built with cmake (and the api enabled), then ctest will run them. In the makefile system
make test_api
should do it.

Pibemanden: I tried building it with cmake and now the C tests works, however when I added the stress tensor test to the fileinitc test I get an error for the second Si calculation that the stress tensor is not available. I am not exactly sure what is causing this though, but to me it seems that it is some bug which is unrelated to what I am doing since each run through of the fileinitc test should initialize a unique calculation, so the third calculation should correctly be initialized as a clean copy of the first calculation.

Trying this test with the master branch of the DFTB+ code produces a "similar" error, in that the second Si calculation (The third calculation run) does not give any output about the Volume or Pressure, i.e. the output stops at the line "Force Related Energy ..." and then gives the timing results.

bhourahine: See a0d4e015 for the fix for this (problem was due to a initial value for a module variable, which then did not get re-set on later calls to initialise with new input).

PR #319, opened 2019-11-13
Implementation of the GFN1-xTB Hamiltonian

awvwgk:
Implementation of the GFN1-xTB method  (see [*JCTC*, **2017**, 13, 1989](https://dx.doi.org/10.1021/acs.jctc.7b00118)) in `dftb+`.

- [ ] Input parser
  - [ ] requires #371 to be merged
- [x] Parameters (probably hardcoded)
  - [x] GFN1-xTB parametrisation hardcoded
- [ ] xTB core Hamiltonian (H0)
  - [x] coordination number container
  - [x] STO-*N*G expansion
  - [x] Overlap integrals and derivatives over GTOs
  - [x] Gaussian-type orbital container
  - [ ] S/H0 evaluation
    - [x] shell polynomials, H0 scaling and EHT average
    - [ ] ordering of basis functions
    - [ ] CAO -> SAO trafo
  - [ ] check against `xtb`
- [ ] modified electrostatics and Ewald summation for Klopman–Ohno Gamma-function
  - [ ] requires #320 to be addressed
- [ ] repulsion energy + derivatives
  - [x] simple repulsion container
  - [x] connection of container with input
  - [ ] check against `xtb`
- [ ] Halogen-bond correction (with derivatives)
- [ ] tests
- [ ] documentation

#### New Input

- [x] preloading of defaults from a parametrisation (should always suffice as calculation input)

```cson
Hamiltonian = xTB {
  param = "gfn1"
}
```

- [x] The diagonal elements of the xTB core Hamiltonian can be shifted by the coordination number, to define the CN in an abstract way a CN container is introduced. This is populated in the input with

```cson
Hamiltonian = xTB {
  CoordinationNumber = exp {  # possible methods are "exp", "erf", "cov" and "gfn"
    Cutoff = 40   # real space cutoff for neighbour list
    MaxCN = 0   # disable cutting CN
    Radii = CovalentRadiiD3 {}  # cutoff radii to count neighbours as bonded
    Electronegativities = PaulingEN {}  # only used for covalent CN
  }
}
```

- [x] both `Radii` and `Electronegativities` can be explicitly populated with the `values` method
- [x] `Radii` accepts a modifier for the length unit used.

```cson
Electronegativities = values {
  H = 2.20
  O = 3.44
  C = 2.55
}
```

- [ ] repulsion polynomials can be setup from the parametrisation

```cson
Hamiltonian = xTB {
  param = "GFN1"
  Repulsion = fromParam {}
}
```

[...]

PR #371, opened 2020-02-05
Refactoring of parser/initprogram for second Hamiltonian

awvwgk:
Aim of this PR is to prepare DFTB+ for a second Hamiltonian. The main issue with the current implementation of `dftbp_parser` and `dftbp_initprogram` is that most of the reading routines are inlined in `readDFTBham`, but they should be shared with other Hamiltonians (like xTB).

- [x] refactor `dftbp_parser`
- [x] refactor `dftbp_initprogram`
- ~refactor `dftbp_main`~

This PR must be merged before #319 can be continued.

Note: This PR is likely to conflict with *every* PR modifying `dftbp_parser` and will require frequent rebasing against new commits to the master branch.

@bhourahine I cherry-picked your `parserCommonFactors` as a starting point for this onto `master`. Any commits on this branch are welcome.


prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -2089,7 +2137,7 @@ subroutine initProgramVariables(input, env)
       ! shell resolved)
       do iSp = 1, nType
         homoLoc = maxloc(atomEigVal(:orb%nShell(iSp), iSp),&
-            & mask=input%slako%skOcc(:orb%nShell(iSp), iSp) > 0.0_dp)
+            & mask=referenceN0(:orb%nShell(iSp), iSp) > 0.0_dp)
```
awvwgk: Is there a reason to use `input%slako%skOcc` instead of `referenceN0` here.

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -2102,7 +2150,7 @@ subroutine initProgramVariables(input, env)
         ! triplet or spin-polarised
         do iSp = 1, nType
           homoLoc = maxloc(atomEigVal(:orb%nShell(iSp), iSp),&
-              & mask=input%slako%skOcc(:orb%nShell(iSp), iSp) > 0.0_dp)
+              & mask=referenceN0(:orb%nShell(iSp), iSp) > 0.0_dp)
```
awvwgk: Same here, can `referenceN0` be used as substitute for the Slater–Koster specific field?

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -1694,7 +1734,15 @@ subroutine initProgramVariables(input, env)
     tForces = input%ctrl%tForces .or. tPrintForces
     tLinResp = input%ctrl%lrespini%tInit
 
-    referenceN0(:,:) = input%slako%skOcc(1:orb%mShell, :)
+    select case(hamiltonianType)
+    case default
+      call error("Invalid Hamiltonian")
+    case(hamiltonianTypes%dftb)
+      referenceN0(:,:) = input%slako%skOcc(1:orb%mShell, :)
```
awvwgk: Here `input%slako%skOcc` is copied to `referenceN0`, might need another copy, if  `referenceN0` is modified.

aradi:
@awvwgk The branch could be theoretically merged, but apparently, it does not build when transport is enabled. Could you try to build it with all possible extensions enabled?

awvwgk:
“All possible extensions” will take me a while to get everything setup on my machine. I think `libnegf` and the `WITH_TRANSPORT` preprocessor is the issue I have to look for, because it actively modifies the dummy arguments of most procedures.

Now it builds when `WITH_TRANSPORT` is set.

awvwgk:
The webhook has not been triggered yet, but build succeeded on my account: https://travis-ci.com/awvwgk/dftbplus/builds/150192278

Maybe we should think about including `libnegf` into the CI. I created a [branch](https://github.com/awvwgk/dftbplus/tree/ci-transport) with the necessary changes for Travis-CI. Timings are [here](https://travis-ci.com/awvwgk/dftbplus/builds/150192603), the transport tests seem to take around ~6 additional minutes, but we will have four concurrent jobs than.

aradi:
Had been rebased onto and merged into master.

awvwgk:
@aradi this has broken master, fix is pending with #386

aradi:
Yes, thanks, I also realized it in the mean time :wink:  I've already pushed the fix to upstream.

PR #374, opened 2020-02-06
Tool for numerical derivatives

awvwgk:
This came up in #314 and is generally useful for new energy expressions.


bhourahine:
We probably need to discuss location for this script ( utils/numcheck ) as there are various other derivative scripts (elastic constants, Born charges, ...) that use essentially the same structure, but are property calculators instead of checking tools.
Also in the longer terms there is the question of whether  to absorb this functionality into dp_tools and also to use the hsd python parser to get input (removing the need to have a specific geometry file name) and to check for autotest creation and settings in the dftb_in.hsd file.

PR #376, opened 2020-02-06
Fixed ford processing

bhourahine:
Replaced non-ascii characters in source and fixed for project files


awvwgk:
Is there a description how to build with the documentation with `ford` and `doxygen`? I know it should be obvious, but I tried using ford from the repo root first which failed, a note in INSTALL.rst would be helpful.

aradi:
I've added the info to `INSTALL.rst`.

PR #367, opened 2020-02-03
Generalized Born implicit solvation model

awvwgk:
This PR implements a generalized Born model used for tight binding methods, see #361 for a general discussion regarding implicit solvation. 
A similar model is used in the DFTB module of AMS and the `xtb` program.

#### Features provided by this PR

- [x] an abstract base class to implement different solvation models in the DFTB+ infrastructure
- [ ] an actual implementation of a generalized Born model providing an implementation for the solvation base class
  - [ ] correct gradient of charge dependent shift / hamiltonian potential
- [x] possibility to modify charges used in the generalized Born model with CM5
- [ ] optional solvent accessible surface area contributions (maybe different PR)

#### New Input

- [x] the solvation model is activated by populating the solvation entry in the HSD input file:

```cson
Hamiltonian = DFTB {
   Solvation = GeneralizedBorn {
      ...
   }
}
```

- [x] the solvent can be specified by name

```cson
Solvation = GeneralizedBorn {
   solvent = fromName { "Acetone" }
}
```

- [x] or by explicitly providing all the necessary constants

```cson
Solvation = GeneralizedBorn {
   solvent = fromConstants {
      epsilon = 20.7  # Dielectric constant of the solvent
      molecularMass = 58.08  # mass of the solvent molecule
      density [mol/cm^-3] = 0.79  # mol density of the solvent
   }
}
```

- [x] The solvation free energy can be calculated against different reference states, depending on the choice of reference the above constants will or will not influence the final energy. (`gsolv=reference` and `mol1bar` missing)
- [x] An absolute shift to the solvation free energy can (has to be) applied as well

```cson
Solvation = GeneralizedBorn {
   temperature = 298.15
   state = "gsolv"  # or "reference" or "mol1bar"
   shift [kcal/mol] = 2.5197963
}
```

- [x] to calculate the Born radii, van-der-Waals radii for each element have to been specified by

```cson
Solvation = GeneralizedBorn {
   radii [AA] = values {
      C = 1.45515
      H = 1.09155
   }
}
```

- [x] Or taken from a default list

```cson
Solvation = GeneralizedBorn {
   radii = vanDerWaalsRadiiD3 {}
}
```

- [x] Global model parameters for the Born radii can be specified with

```cson
Solvation = GeneralizedBorn {
   bornScale = 1.54870752
   bornOffset [AA] = 0.008071897
   OBC = {1.00 0.80 4.85}
}
```

- [x] Additionally descreening constants can be provided for each element to enhance the accuracy of the model.

[...]

PR #379, opened 2020-02-11
Storing data, atomic numbers and PSE “hash map”.

awvwgk:
For xTB, GBSA and DFT-D4 I need to store atomic data (a lot of it: *Z <= 86*) in a dense way, which is most convenient by using its atomic numbers and the PSE as perfect “hash function” to store them as array in the source code. But there is no concept of atomic numbers in DFTB+, so I found myself using the same pattern over and over again.

With the implementation provided in this PR, I could get the array of atomic masses for each species with this example code (ignore the style issues and missing comments for a moment):

```fortran
subroutine initAtomicMass(atomicMass, nSpecies, speciesNames)
  use dtfbp_atomicmass, only : getAtomicMass
  integer, intent(in) :: nSpecies
  character(len=*), intent(in) :: speciesNames(:)
  real(dp), allocatable, intent(out) :: atomicMass(:)
  allocate(atomicMass(nSpecies))
  atomicMass(:) = getAtomicMass(speciesNames)
end subroutine
```

To discuss if this is the right way to do it, I'm opening this PR, so we can discuss it independently of the PRs using it.

**Alternatives to consider:**

- the same way the UFF-dispersion parameters are stored
  - the disadvantage is that we still need to implement the search over the “map”
- make an external parameter file
  - has to be specified in the input parser for each element
- doing the conversion once and passing around the atomic number array instead of the species array
  - done in #314

*Note:* there is no need for this PR to be merged, but either #319 or #367 have to come up with a solution for this problem.

### DIFF
--- a/prog/dftb+/lib_common/CMakeLists.txt
+++ b/prog/dftb+/lib_common/CMakeLists.txt
@@ -3,6 +3,7 @@ set(curdir "lib_common")
 set(sources-fpp
   ${curdir}/accuracy.F90
   ${curdir}/assert.F90
+  ${curdir}/atomicmass.F90
   ${curdir}/constants.F90
   ${curdir}/environment.F90
   ${curdir}/fileregistry.F90
--- a/prog/dftb+/lib_common/atomicmass.F90
+++ b/prog/dftb+/lib_common/atomicmass.F90
@@ -1,0 +1,96 @@
+!--------------------------------------------------------------------------------------------------!
+!  DFTB+: general package for performing fast atomistic simulations                                !
+!  Copyright (C) 2006 - 2020  DFTB+ developers group                                               !
+!                                                                                                  !
+!  See the LICENSE file for terms of usage and distribution.                                       !
+!--------------------------------------------------------------------------------------------------!
+
+#:include 'common.fypp'
+
+!> Atomic masses for each element known in the PSE
+module dftbp_atomicmass
+  use dftbp_accuracy, only : dp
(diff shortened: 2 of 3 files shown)


aradi:
The compact storage is definitely appealing. There are two issues, I can see, however:
* We wont't be able to deal with isotopes (e.g. D or T).
*  Historically, the atomic mass is stored in the homonuclear SK-files (and read from there), so it is a user tunable parameter in some sense.

awvwgk:
> We wont't be able to deal with isotopes (e.g. D or T).

For custom species `D` or `CeIV` the user has to supply the parameters, the pattern for the input would be
```cson
Quantity = Defaults {}
```
or with a custom species
```cson
Quantity = Values {
  H = 1.0
  D = 2.0
  C = 12.0
}
```
Using the `Defaults` method should result in an error if a non-PSE species is encountered.

> Historically, the atomic mass is stored in the homonuclear SK-files (and read from there), so it is a user tunable parameter in some sense.

So in the context of xTB you suggest to add the atomic mass to the parametrisation? Also I'm not to keen to have parameter files for xTB on disk (86 for every parametrisation and I have four of them), so I would prefer to hardcode them, which brings me back to the issue on how to store the data than.

awvwgk:
Maybe the `symbolToNumber` routine should use `dftbp_charmanip` instead of implementing its own `tolower` functionality. But `tolower` is only `pure`, can it be changed to `elemental`?

PR #384, opened 2020-02-21
DFT-D4: Charge Model Container

awvwgk:
This PR abstracts the charge model implementation from the DFT-D4 implementation.

- [x] wrap EEQ calculation in a container
- [x] remove all Ewald specific data from DFT-D4 calculator
- [x] add dedicated input for the charge model used in DFT-D4
  - [x] support EEQ model and allow modification of species parameters
- [x] don't use internals of the EEQ container in DFT-D4
- [x] update the manual and the test inputs

#### New Input

EEQ charge model can be supplied with customized parameters. The `Defaults` method depends on the parametrisation supplied by the parent (here DFTD4), which can be explicitly overwritten in the method. The `Values` method can be used to supply all parameters explicitly.

```cson
Dispersion = DFTD4 {
  ChargeModel = EEQ {
    Chi = Defaults {}
    Gam = Defaults {}
    Kcn = Defaults {}
    Rad = Defaults {}
  }
}
```

### DIFF
--- a/doc/dftb+/manual/dftbp.tex
+++ b/doc/dftb+/manual/dftbp.tex
@@ -2780,11 +2780,9 @@ can be adjusted in the input.
   \kw{ChargeSteepness} & r & & 2.0 & \\
   \kw{ChargeScale} & r & & 3.0 & \\
   \kw{CutoffInter} & r & & 64 & \\
-  \kw{CutoffEwald} & r & & 40 & \\
   \kw{CutoffCount} & r & & 40 & \\
   \kw{CutoffThree} & r & & 40 & \\
-  \kw{EwaldParameter} & r & & 0.0 & \\
-  \kw{EwaldTolerance} & r & & 1.0e-9 & \\
+  \kw{ChargeModel} & m & & \is{EEQ} & \\
 \end{ptable}
 
 \begin{description}
@@ -2812,14 +2810,44 @@ can be adjusted in the input.
 \item[\is{CutoffInter}] \modif{\modtype{length}} Cutoff distance when
   calculating two-body interactions.
 
-\item[\is{CutoffEwald}] \modif{\modtype{length}} Cutoff distance when
-  calculating electrostatics interactions under PBC.
-
 \item[\is{CutoffCount}] \modif{\modtype{length}} Cutoff distance when
   calculating coordination numbers.
 
 \item[\is{CutoffThree}] \modif{\modtype{length}} Cutoff distance when
   calculating three-body interactions.
+
+\end{description}
+
+\subsubsection{DftD4 ChargeModel}
+
+This implementation of DFT-D4 supports only the \is{EEQ\cb} method to initialize
+the charge model. For each species four parameters (\kw{Chi}, \kw{Gam}, \kw{Kcn},
+and \kw{Rad}) have to be supplied in a \is{values\cb} method, since the model
+is instanciated inside the \is{DftD4\cb} method, defaults for all elements up to
+86 are supplied automatically.
+
+\begin{ptable}
+  \kw{Chi} & m & & \is{values} & \\
+  \kw{Gam} & m & & \is{values} & \\
+  \kw{Kcn} & m & & \is{values} & \\
+  \kw{Rad} & m & & \is{values} & \\
+  \kw{Cutoff} & r & & 40 & \\
+  \kw{EwaldParameter} & r & & 0.0 & \\
+  \kw{EwaldTolerance} & r & & 1.0e-9 & \\
+\end{ptable}
+
+\begin{description}
+
+\item[\is{Chi}] Electronegativities of all species.
+
+\item[\is{Gam}] Chemical hardnesses of all species.
+
+\item[\is{Kcn}] CN scaling factor of all species.
+
+\item[\is{Rad}] Charge width of all species in Bohr.
+
+\item[\is{CutoffEwald}] \modif{\modtype{length}} Cutoff distance when
+  calculating electrostatics interactions under PBC.
 
 \item[\is{EwaldParameter}] Sets the splitting parameter in the Ewald
   electrostatic summation for periodic calculations. This controls the fraction
--- a/doc/dftb+/manual/dftbp.tex
+++ b/doc/dftb+/manual/dftbp.tex
@@ -2822,15 +2822,15 @@ can be adjusted in the input.
 
 This implementation of DFT-D4 supports only the \is{EEQ\cb} method to initialize
 the charge model. For each species four parameters (\kw{Chi}, \kw{Gam}, \kw{Kcn},
-and \kw{Rad}) have to be supplied in a \is{values\cb} method, since the model
-is instanciated inside the \is{DftD4\cb} method, defaults for all elements up to
-86 are supplied automatically.
+and \kw{Rad}) have to be supplied in a \is{Values\cb} method, since the model
+is instanciated inside the \is{DftD4\cb} method, \is{Defaults\cb} for all elements
+up to 86 can be supplied automatically.
 
 \begin{ptable}
-  \kw{Chi} & m & & \is{values} & \\
-  \kw{Gam} & m & & \is{values} & \\
-  \kw{Kcn} & m & & \is{values} & \\
-  \kw{Rad} & m & & \is{values} & \\
+  \kw{Chi} & m & & \is{Defaults} & \\
+  \kw{Gam} & m & & \is{Defaults} & \\
+  \kw{Kcn} & m & & \is{Defaults} & \\
+  \kw{Rad} & m & & \is{Defaults} & \\
   \kw{Cutoff} & r & & 40 & \\
   \kw{EwaldParameter} & r & & 0.0 & \\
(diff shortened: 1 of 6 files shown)


prog/dftb+/lib_dftb/encharges.F90
```diff
@@ -8,28 +8,442 @@
...
+
+    !> New lattice vectors
+    real(dp), intent(in) :: latVecs(:, :)
+
+    real(dp) :: recVecs(3, 3), maxGEwald
+
+    @:ASSERT(this%tPeriodic)
+    @:ASSERT(all(shape(latVecs) == shape(this%latVecs)))
+
+    this%latVecs = latVecs
```
bhourahine: this%latVecs(:,:) =
as a habit to avoid chance of doing this to an allocatable at some point.

prog/dftb+/lib_dftb/encharges.F90
```diff
@@ -8,28 +8,442 @@
...
+    this%vol = determinant33(latVecs)
+    call invert33(recVecs, latVecs, this%vol)
+    this%vol = abs(this%vol)
+    recVecs = 2.0_dp * pi * transpose(recVecs)
```
bhourahine: recVecs(:,:)

prog/dftb+/lib_dftb/encharges.F90
```diff
@@ -8,28 +8,442 @@
...
+
+    if (this%tAutoEwald) then
+      this%parEwald = getOptimalAlphaEwald(latVecs, recVecs, this%vol, this%tolEwald)
+    end if
+    maxGEwald = getMaxGEwald(this%parEwald, this%vol, this%tolEwald)
+    call getLatticePoints(this%recPoint, recVecs, latVecs/(2.0_dp*pi), maxGEwald,&
+        & onlyInside=.true., reduceByInversion=.true., withoutOrigin=.true.)
+    this%recPoint = matmul(recVecs, this%recPoint)
```
bhourahine: (:,:)

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -3425,17 +3430,157 @@ subroutine readDispDFTD4(node, input, nrChrg)
...
     call convertByMul(char(buffer), lengthUnits, child, input%cutoffThree)
 
+    call getChildValue(node, "ChargeModel", value1, "EEQ", child=child)
+    call getNodeName(value1, buffer)
+    select case(char(buffer))
+    case default
+      call detailedError(value1, "Unknown method '"//char(buffer)//"' for ChargeModel")
+    case ("eeq")
+      allocate(d4Chi(geo%nSpecies))
+      d4Chi = getEeqChi(geo%speciesNames)
```
bhourahine: left brackets on these

prog/dftb+/lib_dftb/encharges.F90
```diff
@@ -8,28 +8,442 @@
...
+    integer, intent(in) :: nAtom
+
+    !> Lattice vectors, if the system is periodic.
+    real(dp), intent(in), optional :: latVecs(:, :)
+
+    real(dp) :: recVecs(3, 3), maxGEwald
+
+    this%tPeriodic = present(latVecs)
+
+    this%param = input%TEeqParam
```
awvwgk: @bhourahine What about this part? It is using automatic allocation of LHS on purpose. Here the purpose is to make this part of the initialization less error prone.

bhourahine: As that is the deliberate intention and its not in a loop, I don't see why. @aradi opinions ?

awvwgk:
Just pushed some minor style fixes, I noticed while working with this branch.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -3800,21 +3801,25 @@ end subroutine readDispDFTD3
   !> Here we additionally require a s9, since the non-addititive contributions
   !> tend to be expensive especially in the tight-binding context, s9 = 0.0_dp
   !> will disable the calculation.
-  subroutine readDispDFTD4(node, input, nrChrg)
+  subroutine readDispDFTD4(node, geo, input, nrChrg)
+    !> Do not pollute the global namespace of the parser -> import locally
+    use dftbp_dftd4param, only : getEeqChi, getEeqGam, getEeqKcn, getEeqRad
```
aradi: Could you please move the use-statement up to module level?

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -3832,17 +3837,157 @@ subroutine readDispDFTD4(node, input, nrChrg)
...
+    type(eeqInput), intent(out) :: input
+
+    !> Net charge of the system.
+    real(dp), intent(in) :: nrChrg
+
+    !> Electronegativities default values
+    real(dp), intent(in), optional :: kChiDefault(:)
+
+    !> Chemical hardnesses default values
+    real(dp), intent(in), optional :: kGamDefault(:)
```
aradi: Why `optional`? Do we (will we) ever invoke it without those paramters? If not, I'd prefer make them required arguments.

prog/dftb+/lib_dftb/encharges.F90
```diff
@@ -565,22 +979,22 @@ subroutine getEEQCharges(nAtom, coords, species, charge, nNeighbour, iNeighbour,
     real(dp), intent(in) :: dcndL(:, :, :)
 
     !> Updated energy vector at return
-    real(dp), intent(inout), optional :: energies(:)
+    real(dp), intent(inout), allocatable :: energies(:)
```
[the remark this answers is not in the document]
awvwgk: Using `optional` leads to a significant logical overhead upstream (`updateCoords`), therefore I decided to make them `allocatable` and perform this kind of evaluation here.

doc/dftb+/manual/dftbp.tex
```diff
@@ -2812,15 +2810,45 @@ \subsubsection{DftD4 settings}
...
+
+\item[\is{Gam}] Chemical hardnesses of all species.
+
+\item[\is{Kcn}] CN scaling factor of all species.
+
+\item[\is{Rad}] Charge width of all species in Bohr.
+
+\item[\is{Cutoff}] \modif{\modtype{length}} Cutoff distance when
+  calculating electrostatics interactions under PBC.
+
```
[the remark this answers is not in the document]
awvwgk: Since everything is set automatically the minimal input would be empty :wink:.

prog/dftb+/lib_dftb/encharges.F90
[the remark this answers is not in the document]
awvwgk: This is actually a really neat feature of Fortran. Certainly learned something today.

prog/dftb+/lib_dftbplus/parser.F90
awvwgk: Since it is a container, it should be reusable (together with its input parser), e.g. as electrostatic energy for a non-SCC tight binding calculation, which does not necessarily come with default parameters. But for a few more weeks, I guess, it will be used exclusively by DFT-D4, so let's require them for now.

prog/dftb+/lib_dftb/encharges.F90
```diff
@@ -8,28 +8,442 @@
...
+    !> cutoff distance in real space for EEQ
+    procedure :: getRCutoff
+
+  end type TEeqCont
+
+
 contains
 
+
+  subroutine initialize(this, input, withEnergies, withCharges, nAtom, latVecs)
```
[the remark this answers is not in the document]
awvwgk: Are there any arguments against using a `function` and overloading the constructor:

```fortran
interface TEqCont
  module procedure :: initialize
end interface TEqCont
```

prog/dftb+/lib_dftb/dftd4param.F90
```diff
@@ -12,6 +12,7 @@ module dftbp_dftd4param
   use dftbp_assert
   use dftbp_accuracy, only : dp
   use dftbp_constants, only : pi, AA__Bohr, symbolToNumber
+  use dftbp_encharges, only : eeqInput
```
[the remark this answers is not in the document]
awvwgk: I was a bit confused about this part, since it is not enforced consistently. Fortunately, this [is going to change](https://github.com/bhourahine/dftbplus/tree/typeNameConvnt).

prog/dftb+/lib_dftb/encharges.F90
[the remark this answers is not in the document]
awvwgk: The alternative is to keep the constructor type bound and call it from the pure type constructor, like you would call a type bound deconstructor from the finalizer. If you inherit from the class, you have to overwrite both constructor and deconstructor and call back to its parent, which would make more sense to me.

To be fair, I actually don't see a need to have the classes here extensible, it is just more convenient to have all the procedures type bound to avoid polluting the global namespace.

awvwgk:
Any update on the status of this PR?

bhourahine:
@awvwgk can you rebase up to the current master, then I'll merge.

awvwgk:
@bhourahine Just have. CI is still passing, so this fine to go.

PR #387, opened 2020-02-23
Generalized Born Implicit Solvation Model

awvwgk:
This PR implements the most basic generalized Born (GB) implicit solvation model. A similar model is used in the DFTB module of AMS and the `xtb` program.

#### Features provided

- [x] abstract interface for solvation models
  - [x] MPI aware interface
- [x] actual generalized Born solvation model

#### New Input

Introduces the `Solvation` keyword in the `Hamiltonian`:

```cson
Hamiltonian = DFTB {
  Solvation = GeneralizedBorn { # GFN2-xTB/GBSA(CS2)
    Solvent = fromConstants {
      Epsilon = 2.64  # Dielectric constant of the solvent
      MolecularMass [amu] = 76.14  # mass of the solvent molecule
      Density [kg/l] = 1.26  # mol density of the solvent
    }
    FreeEnergyShift [kcal/mol] = 2.81072250
    Radii = vanDerWaalsRadiiD3 [Bohr] {}
    Descreening = values {
       H = 0.93699367
       C = 0.83307834
       N = 1.02661619
       O = 0.96508008
    }
    BornScale = 1.40636177
    BornOffset = 1.653719965215E-03
    OBC = {1.00 0.80 4.85}
    Cutoff = 30
  }
}
```

#### Implementation Details

~The implementation is available for arbitrary boundary conditions via static neighbourlists. Despite being an electrostatic contribution, the Born energy should always converge within the largest Born radius, therefore, no need for an Ewald summation arises here.~
Forget about this, it does require an Ewald summation. PBCs for the GB model will be a separate PR.

Most parts of the interface are consistent with the `DispersionIFace` or `ThirdOrder` class, to be uniform in the existing infrastructure.


bhourahine:
Is there a specific reason for the 45a5751 periodic change?

awvwgk:
First, I did not implement the strain derivatives yet, so there is no stress tensor, and second, it does not really make sense for a 3D periodic system, despite being technically possible. For 2D PBCs or your helical boundary conditions it would be really useful.

I will do the strain derivatives later when I'm done with the PR for the surface term, which involves strain derivatives of an integration grid.

awvwgk:
Since we merged #388, I made this interface MPI aware as well. I think it is ready for review now.

There will be a nonpolar surface term, SASA model, in a different PR and a correction to the Mulliken charges by the Charge Model 5 in another PR once this is merged.

awvwgk:
Rebased against master because of #392.

prog/dftb+/lib_dftbplus/main.F90
```diff
@@ -5233,6 +5269,14 @@ subroutine getGradients(env, sccCalc, tExtField, isXlbomd, nonSccDeriv, EField,
       end if
     end if
 
+    if (allocated(solvation)) then
+      if (isXlbomd) then
+        call error("XLBOMD does not work with solvation yet!")
```
bhourahine: Needs equivalent tests for use with transport or excited states (but in the longer term would be good to get these working with the Born model).

doc/dftb+/manual/manual.bib
```diff
@@ -642,6 +642,18 @@ @article{dominguez15
 doi = {10.1021/acs.jpca.5b01732}
 }
 
+@article{onufriev2019
```
bhourahine, line 645: malformed line, missing comma at end.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -3516,6 +3536,139 @@ subroutine readOptions(node, ctrl)
...
+    case('fromconstants')
+      call getChildValue(value1, "Epsilon", solvent%dielectricConstant)
+      call getChildValue(value1, "MolecularMass", solvent%molecularMass)
+      call getChildValue(value1, "Density", solvent%density)
+    end select
+
+    input%keps = 1.0_dp / solvent%dielectricConstant - 1.0_dp
+
+    ! shift value for the free energy (usually fitted)
+    call getChildValue(node, "Shift", shift, modifier=modifier, child=field)
```
bhourahine, line 3611: A bit terse as a keyword, what about FreeEShift / ShiftFreeEnergy or similar?

doc/dftb+/manual/dftbp.tex
```diff
@@ -2875,6 +2876,138 @@ \subsection{DFTB3}
 \end{verbatim}
 \end{description}
 
+\subsection{Implicit Solvation Model}
+\label{sec:dftbp.solvation}
+
+\subsubsection{Generalized Born Model}
+\label{sec:dftbp.GeneralizedBorn}
+
+In generalized Born (GB) theory,\cite{onufriev} a molecule is considered as
```
bhourahine, line 2885: Broken reference, probably should be onufriev2019

doc/dftb+/manual/dftbp.tex
```diff
@@ -1090,6 +1090,7 @@ \section{Hamiltonian}
   \kw{RangeSeparated} & p &  & \is{None} & \pref{sec:dftbp.RangeSep} \\
   \kw{HubbardDerivs} & p & ThirdOrder(Full) = Yes  & - & \\
   \kw{OnSiteCorrection} & p & SCC = Yes & - & \pref{sec:dftbp.Onsites} \\
+  \kw{Solvation} & m & & - & \pref{sec:dftbp.Solvation}\\
```
bhourahine, line 1093: Seems to be a broken reference at compile time, probably change in capitalisation for 'Solvation' in the \label{}

doc/dftb+/manual/dftbp.tex
```diff
@@ -2875,6 +2876,138 @@ \subsection{DFTB3}
...
+approach.
+
+To use the generalized Born model in the SCC procedure use the
+\is{GeneralizedBorn\cb} method in the input to \kw{Solvation}.
+
+\begin{ptable}
+  \kw{Solvent} & m & & & \\
+  \kw{Shift} & r & & & \\
+  \kw{Temperature} & r & & 298.15 & \\
+  \kw{State} & s & & gsolv & \\
```
bhourahine, line 2942: Meaning of gsolv is not defined anywhere in the manual.

doc/dftb+/manual/manual.bib
```diff
@@ -642,6 +642,18 @@ @article{dominguez15
...
+author = {Onufriev, Alexey V. and Case, David A.},
+title = {Generalized Born Implicit Solvent Models for Biomolecules},
```
bhourahine, line 647: Due to bibtex behaviour, needs the first letter protecting from lower case conversion: {B}orn

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -3516,6 +3536,139 @@ subroutine readOptions(node, ctrl)
...
+    case default
+      call detailedError(child, "Unknown reference state: '"//char(state)//"'")
+    case("gsolv") ! just the bare shift
+      input%shift = shift
+    end select
+
+    call getChildValue(node, "BornScale", input%bornScale)
+    call getChildValue(node, "BornOffset", input%bornOffset, modifier=modifier, child=field)
+    call convertByMul(char(modifier), lengthUnits, field, shift)
+    call getChildValue(node, "OBC", input%obc, [1.00_dp, 0.80_dp, 4.85_dp])
```
bhourahine, line 3629: Very obtuse keyword name, OBCCorrection (?)
The manual also doesn't give any hints how to choose these values/where the defaults come from. I assume this is described in Onufriev Proteins 55:383–94, so would be good to have a reference as to where to find values/short description in the manual. I think you've used GB^{OBC}II values.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -3516,6 +3536,139 @@ subroutine readOptions(node, ctrl)
...
+      call detailedError(child, "Unknown method '"//char(buffer)//"' to generate radii")
+    case("vanderwaalsradiid3")
+      allocate(vdwRadDefault(geo%nSpecies))
+      vdwRadDefault(:) = getVanDerWaalsRadiusD3(geo%speciesNames)
+      do iSp = 1, geo%nSpecies
+        call getChildValue(value1, geo%speciesNames(iSp), input%vdwRad(iSp), &
+            & vdwRadDefault(iSp), child=child2)
+      end do
+      deallocate(vdwRadDefault)
+    case("values")
```
bhourahine, line 3645: Back to debate about units in these cases, probably should carry one here.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -3516,6 +3536,139 @@ subroutine readOptions(node, ctrl)
...
+    case('fromname')
+      call getChildValue(value1, "", buffer)
+      call SolventFromName(solvent, unquote(char(buffer)), found)
+      if (.not. found) then
+        call detailedError(value1, "Invalid solvent " // char(buffer))
+      end if
+    case('fromconstants')
+      call getChildValue(value1, "Epsilon", solvent%dielectricConstant)
+      call getChildValue(value1, "MolecularMass", solvent%molecularMass)
+      call getChildValue(value1, "Density", solvent%density)
```
bhourahine, line 3605: Not obvious here (or the manual) what unit system this is. Probably should have a modifier

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -3516,6 +3536,139 @@ subroutine readOptions(node, ctrl)
...
+      call detailedError(child, "Invalid solvent method '" // char(buffer) // "'")
+    case('fromname')
+      call getChildValue(value1, "", buffer)
+      call SolventFromName(solvent, unquote(char(buffer)), found)
+      if (.not. found) then
+        call detailedError(value1, "Invalid solvent " // char(buffer))
+      end if
+    case('fromconstants')
+      call getChildValue(value1, "Epsilon", solvent%dielectricConstant)
+      call getChildValue(value1, "MolecularMass", solvent%molecularMass)
```
bhourahine, line 3604: Not obvious here (or the manual) what unit system this is. Probably should have a modifier.

doc/dftb+/manual/dftbp.tex
```diff
@@ -2875,6 +2876,138 @@ \subsection{DFTB3}
...
+  \kw{BornScale} & r & & & \\
+  \kw{BornOffset} & r & & & \\
+  \kw{OBC} & r & & 1.00, 0.80, 4.85 & \\
+  \kw{Radii} & m & & & \\
+  \kw{Descreening} & m & & & \\
+  \kw{Cutoff} & r & & & \\
+\end{ptable}
+
+\begin{description}
+  \item[\is{Solvent}] Descriptors of the solvent, can be load from a database
```
bhourahine, line 2952: Needs a table (appendix?) of possible choices for solvent names.

prog/dftb+/lib_solvation/solvinput.F90
```diff
@@ -0,0 +1,82 @@
...
+  end type TSolvationInp
+
+  !> Wrapper to create a solvation model from its respective input
+  interface createSolvationModel
+    module procedure :: createGeneralizedBornModel
+  end interface
+
+contains
+
+  !> Print the solvation model used
```
bhourahine, line 35: For user experience, if they set a named solvent it would be useful to print out the name in the output, i.e. store a representation of the initialization of the model somewhere. Possibly the stdOut stream should have more information at the initialization stage (temperature, shift, ...). Of course they should be referring to the dftb_pin for this, but it doesn't always happen.

doc/dftb+/manual/dftbp.tex
```diff
@@ -2875,6 +2876,138 @@ \subsection{DFTB3}
...
+    Offset value for Born radii calculation.
+
+  \item[\is{OBC}] Parameters for volume polynomial to correct Born radii
+    calculation.
+
+  \item[\is{Radii}] Atomic radii for each element in Bohr, either takes
+    \is{VanDerWaalsRadiiD3\cb} for DFT-D3 van-der-Waals radii (can be overwritten)
+    or requires to provide \is{Values\cb} for all species.
+
+  \item[\is{Descreening}] Descreening values for each elements.
```
bhourahine, line 2979: "for each chemical element" ?

test/prog/dftb+/tests
```diff
@@ -107,6 +107,8 @@ scc/HH_custom-ref                      #? MPI_PROCS <= 1
 spin/Fe4                               #? MPI_PROCS == 2 or not WITH_MPI
 spin/Fe4_commonFermi                   #? MPI_PROCS <= 2
 
+solvation/gb_ionpair                   #? MPI_PROCS <= 1
```
bhourahine, line 110: Worth adding a solvent = ??? case as well ?

prog/dftb+/lib_dftbplus/parser.F90
aradi: `FreeEnergyShift` sounds to me descriptive enough. :wink:

prog/dftb+/lib_dftbplus/parser.F90
awvwgk: I've a proposal for modifiers here: d96d163. It's a bit hacky, but works in a robust way.

awvwgk: Having the modifier at `vanDerWaalsRadiiD3` or `values` instead of `radii` would be preferable, but I don't see how to do this with the `getChildValue` routines.

aradi: See 569fef6. :wink: 

Unfortunately, playing with it drew my attention to a bug in our current parser API: The parser allowed for a modifier on `VanDerWaalsRadiiD3` in the previous implementation, it was just ignored. For the current case, it is fixed now, but way have several places in the input, where an accidently specified modifier may be silently ignored, instead of triggering an error.....

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -3625,12 +3625,13 @@ subroutine readSolvGB(node, geo, input)
 
     call getChildValue(node, "BornScale", input%bornScale)
     call getChildValue(node, "BornOffset", input%bornOffset, modifier=modifier, child=field)
-    call convertByMul(char(modifier), lengthUnits, field, shift)
+    call convertByMul(char(modifier), lengthUnits, field, input%bornOffset)
```
awvwgk, line 3628: Nice catch.

doc/dftb+/manual/dftbp.tex
awvwgk: Maybe, we maintain a list [here](https://xtb-docs.readthedocs.io/en/latest/gbsa.html#parameterized-solvents) and a repo [here](https://github.com/grimme-lab/gbsa-parameters)

awvwgk: Eventually the HSD input should be a fallback if someone is keen enough to make their own parametrisation, in the end of this series of 4 or 5 PRs there will  be a convenient way to include a parameter file provided by some default channel (maybe even downloaded by `./utils/get_opt_externals`).

doc/dftb+/manual/dftbp.tex
awvwgk: Removing it for now.

prog/dftb+/lib_dftbplus/parser.F90
awvwgk: So every construct like
```cson
key {
  method {
    A = ...
    B = ...
    C = ...
  }
}
```
can carry an untracked modifier at the `method` level?

test/prog/dftb+/tests
awvwgk: With the currently implemented features this is just a shortcut for a dielectric constant. It is more testing the parser than the implementation, combined with non-default radii with non-default units it might be worth anyway.

awvwgk:
I decided to make the Born matrix evaluation explicit molecular by removing the neighbour list. This is more in line with the way `dftbp_coulomb` is dealing with the Coulomb matrix. This needs to be rewritten in case we come up with a unified framework, anyway. Maybe https://github.com/grimme-lab/xtb/pull/164 is taking a step in the right direction and I can port it to DFTB+ once I finished it in `xtb`.

prog/dftb+/lib_solvation/solvinput.F90
awvwgk: Does DFTB+ print the name of the SK-files used? The verbosity is at the same as with the dispersion corrections, maybe I can come up with an option to make it more verbose and print a detailed summary on the solvation model.
This will be definitely necessary for the final GBOBC+SASA+CM5+HB model we are aiming for.

awvwgk:
Rebased against master.

prog/dftb+/lib_common/unitconversion.F90
```diff
@@ -205,12 +201,45 @@ module dftbp_unitconversion
...
+      &unit("kg                  ", kg__au ), &
+      &unit("g                   ", 1.0e+3_dp*kg__au ) &
       &/)
 
+
+  !> Number of mass density units
+  integer, parameter :: nMassDensityUnit = 19
+
+
+  type(unit), parameter :: massDensityUnits(nMassDensityUnit) = [ &
```
awvwgk, line 223: Having a total of 19 units for the mass density is certainly overkill, but covers the ones I actually need + all the combinations from mass and volume.

doc/dftb+/manual/dftbp.tex
```diff
@@ -2968,6 +2969,141 @@ \subsection{DFTB3}
...
+\is{GeneralizedBorn\cb} method in the input to \kw{Solvation}.
+
+\begin{ptable}
+  \kw{Solvent} & m & & & \\
+  \kw{FreeEnergyShift} & r & & & \\
+  %\kw{Temperature} & r & & 298.15 & \\
+  %\kw{State} & s & & gsolv & \\
+  \kw{BornScale} & r & & & \\
+  \kw{BornOffset} & r & & & \\
+  \kw{OBCCorrection} & r & & 1.00, 0.80, 4.85 & \\
```
bhourahine, line 3038: 3r not r

doc/dftb+/manual/dftbp.tex
```diff
@@ -2968,6 +2969,141 @@ \subsection{DFTB3}
...
+    Real space cutoff for the calculation of the Born radii.
+\end{description}
+
+Example for CS2 as solvent:
+
+\begin{verbatim}
+Hamiltonian = DFTB {
+  Solvation = GeneralizedBorn { # GFN2-xTB/GBSA(CS2)
+    Solvent = fromName { "cs2" }
+    Temperature = 298.15
```
bhourahine, line 3089: Keyword is currently commented out

doc/dftb+/manual/dftbp.tex
```diff
@@ -2968,6 +2969,141 @@ \subsection{DFTB3}
...
+    State = "gsolv"
```
bhourahine, line 3090: likewise, currently inactive keyword

PR #390, opened 2020-02-24
Updated changelog as cmake was not mentioned

bhourahine:


CHANGELOG.rst
```diff
@@ -30,6 +30,9 @@ Changed
 
 - Input in GEN format now strictly follows the format described in the manual
 
+- New build system using CMake (the old makefile system will be supported for 1
+  further release)
```
awvwgk: Since we decided to support the old make build system only for one more week, I think we better drop the second part.

Issue #378, opened 2020-02-10
CI Test for Building Documentation


awvwgk:
Should there be a CI test for building the documentation with FORD and doxygen? Maybe even compiling the manual to check for completeness.

I could setup a GH action for this.

PR #393, opened 2020-02-28
Documentation building action

awvwgk:
[docs](https://github.com/awvwgk/dftbplus/actions)

Use GH actions to build the DFTB+ manual. Overall runtime is about ~three~ one minute~s~ ~(two minutes for the docker container)~.

- [x] build of the manual is done only for changes in the `doc/dftb+/manual` directory
- [x] uploads of the pdf only for this repository and push events
- [x] upload manual for push events on master

To discuss
- ~run doxygen and/or ford to generate the documentation from the source?~ not used anyway...

Related to #378

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -1,0 +1,20 @@
+name: docs
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: dante-ev/latex-action@v0.2.0
+      with:
+        root_file: manual.tex
+        working_directory: doc/dftb+/manual
+    - uses: actions/upload-artifact@v1
+      if: ${{ github.repository == 'dftbplus/dftbplus' && github.event_name == 'push' }}
+      with:
+        name: dftbplus-manual
+        path: doc/dftb+/manual/manual.pdf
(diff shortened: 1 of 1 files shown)


awvwgk:
Check the result here: https://github.com/awvwgk/dftbplus/actions/runs/46720402

awvwgk:
I think this is ready for review now.

aradi:
Wow, it seems to be more complex, than I thought. Building a special Docker-container just in order to run a simple LaTeX command is for me a huge task (with a big CO2 footprint :wink: ) for a very minor advantage. So, let's restrict it to the necessary commits only, when changes had been done in `doc/dftb+/manual`. As for the deployment, not sure, whether a PDF-file hidden somewhere in GitHub actions is of really much use for the users... We can have a quick discussion about this at the next meeting....

awvwgk:
Using an Ubuntu image an installing `texlive-full` will require 6 minutes, so pulling in the docker container saves us 4 minutes download and install for a full texlive stack.

I'm using an Ubuntu image now and only `texlive-latex-extra` and `texlive-science` which reduce the cost of this action to 1 minute (see https://github.com/awvwgk/dftbplus/actions/runs/81440651).

awvwgk:
With the current setup the build will be performed for changes in the `doc/dftb+/manual` directory and report failing manual builds in PR. Generally this should happen at least once per PR (in case it is implementing a new feature).

Merging the PR into master will trigger the build again and upload the manual.

PR #394, opened 2020-02-28
Update for Travis-CI build

awvwgk:
- enables WITH_TRANSPORT
- excludes expensive tests


awvwgk:
Still not working as expected...

PR #395, opened 2020-02-28
Update for Travis-CI build

awvwgk:
[Build Status](https://travis-ci.com/awvwgk/dftbplus)

- every test taking longer than a minute has to be guard with `LONG_TEST` and is disabled for Travis-CI
- enable compilation `WITH_TRANSPORT`
- remove redundant initialization of submodules

I looked up the Travis-CI specs, we have 2 cores on our virtual machine for building. With the new setup we overall save one minute and test most of the transport features.

Logs and timings are here: https://travis-ci.com/awvwgk/dftbplus/builds/151042525

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,12 +20,11 @@ addons:
       - libscalapack-openmpi-dev
 
 install:
-- git submodule update --init --recursive
 - echo "y" | ./utils/get_opt_externals ALL
 
 script:
 - mkdir _build
 - cd _build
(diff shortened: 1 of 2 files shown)


awvwgk:
Of course, this should be added, but where? It seems that there is no documentation around the test definition in the `tests` file except for the outdated [README in test directory](https://github.com/dftbplus/dftbplus/blob/master/test/prog/dftb%2B/README). At least the workflow described there is not the one I'm using for adding tests and testing.

Adding it in `INSTALL.rst` is suboptimal, because it is not only targeting developers. Adding it to developer guide would be best.

PR #398, opened 2020-02-29
Separate SCC and second order electrostatic

### LINKED ISSUE #320: Generalize Coulomb Matrix Setup / Ewald summation
Right now the Coulomb matrix setup seems to only support 1/r terms. Screened Coulomb interactions (like erf(r)/r or Klopman–Ohno) could also be handled by the current implementation, without the need of reproduce existing code like done in #314.

**Possible Solution**
1. A Coulomb container similar to the one used for the repulsion energy might work here and would allow to generalize the Coulomb matrix setup.
2. Alternatively one could use the existing routines, remove the 1/r afterwards and add the screened Coulomb interactions back. But this seems somewhat impractical at best.

awvwgk:
**WIP**: do not merge!

My take on abstracting the evaluation of the second order electrostatic / coulombic interactions from the SCC. Aim of this PR is to reuse the second order electrostatic for all implemented and future boundary conditions within a different context than the DFTB SCC.

Possible applications are the extended tight binding (xTB) SCC (see #319), the electronegativity equilibration model (EEQ) model in DFT-D4 (see #384) and the generalized Born (GB) implicit solvation model (see #387), which are right now either *blocked* by the current infrastructure or *reimplement* the second order electrostatic by themselves.

fixes #320

### DIFF
--- a/prog/dftb+/lib_dftb/coulomb.F90
+++ b/prog/dftb+/lib_dftb/coulomb.F90
@@ -24,7 +24,7 @@ module dftbp_coulomb
   use dftbp_constants, only : pi
   use dftbp_dynneighlist
   use dftbp_commontypes, only : TOrbitals
-  use dftbp_periodic, only : TNeighbourList
+  use dftbp_periodic, only : TNeighbourList, getLatticePoints, getCellTranslations
   implicit none
 
   private
@@ -53,6 +53,13 @@ module dftbp_coulomb
 
   !> Input data for coulombic interaction container
   type :: TCoulombInput
+
+    !> if > 0 -> manual setting for alpha
+    real(dp) :: ewaldAlpha = 0.0_dp
+
+    !> Ewald tolerance
+    real(dp) :: tolEwald = 0.0_dp
+
   end type TCoulombInput
 
 
@@ -68,6 +75,12 @@ module dftbp_coulomb
     !> lattice vectors if periodic
     real(dp) :: latVecs(3, 3)
 
+    !> reciprocal lattice vectors
+    real(dp) :: recVecs(3, 3)
+
+    !> Cell volume
+    real(dp) :: volume
+
     !> Lattice points for reciprocal Ewald
     real(dp), allocatable :: gLatPoint(:,:)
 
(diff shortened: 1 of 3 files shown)


awvwgk:
@bhourahine @aradi Any help regrading this PR (by commits or comments) is highly appreciated.

Right now, I'm not sure how to refactor the update of the charges and calculation of the shifts.

awvwgk:
My current strategy is to move code bits from `scc.F90` to `coulomb.F90` until I can reduce the imports from `dftbp_coulomb` to the container and its input, while trying not to break anything. This is a somewhat crude approach but usually works well with legacy code :wink:.

awvwgk:
Right now this PR provides a thin and transparent abstraction layer between the SCC and second order electrostatics. It is a good starting point for further refactoring, but not the final solution for the problems mentioned above.

awvwgk:
The logic for updating the shifts is really complicated, which is partly caused by the Poisson solver which is implemented in `dftbp_main` rather than `dftbp_scc`. I wouldn't move it to `dftbp_coulomb` but build the logic into the `updateShifts` routine.

https://github.com/dftbplus/dftbplus/blob/2ec42c063dda3d38301aa6fd62d95656cb21289e/prog/dftb%2B/lib_dftbplus/main.F90#L1749-L1790

If the select case was in `updateShifts` for the SCC it could internally set the shifts and the interface would look the uniform.

aradi:
That's a perfect first step IMO and helps to entangle the Coulomb and the SCC-modules. We should merge it at some point not to have ethernal branches. Also, I agree, that this part of the code is an entangled mess, so we need several steps to get it right.

For the long term, I would envision following architecture:

* Generic Coulomb class, offering generic interfaces for updating and querying things (e.g. `setCoords`, `addShift` etc, with specific implementations for 0D, 3D, cylindrical etc. The specific implementation should be initialized in `initprogram.F90`, and `scc.F90` should not be aware of the actual type. This way, we could make `scc.F90` boundary-condition agnostic.

* Generic SccCalculator class, with specific implementations using either Gamma-based electrostatics or Poission based one. Again, it should be decided in `initprogram.F90` and the main code should not be aware of the specific choice (if possible).

What would be the next step? Probably, we should try to disentangle the parts from `scc.F90` which need access to internal details of the Coulomb-calculator (e.g. querying `alpha`) or using its array to sum up over reciprocal lattice vectors. All those should be generated IMO in `initprogram.F90` and passed in some forms to all the modules which need them. Maybe, with some kind of proxy-object versioning it, so that the modules can be aware of geometry changes.

awvwgk:
Maybe we should move towards dropping the current `dftbp_scc` module completely and use the SCC implementation in `dftbp_main` as starting point for a new SCC container, this means everything in `dftbp_scc` either moves down to `dftbp_main` or up to `dftbp_coulomb`.

The new SCC should than just look like any other (geometry dependent) container and contain all the charge dependent containers of the method. This would also save us from reevaluating the gross charges in every charge dependent container, because we could rely on the SCC container to provide us with the correct ones.

PR #400, opened 2020-03-02
Non-polar Surface Area Solvation Model

awvwgk:
Molecular implementation of a solvent accessible surface area (SASA) model.

#### Features provided

- [x] All angular Lebedev-Laikov grids
- [x] SASA calculation and evaluation of surface tension

#### New Input

Introduces `SASA` solvation:

```cson
Solvation = SASA { # GFN1-xTB/GBSA(Toluene)
  ProbeRad [AA] = 1.59772343
  Smoothing [AA] = 0.3
  Offset [AA] = 2
  AngularGrid = 230
  Radii = vanDerWaalsRadiiD3 {}
  SurfaceTension = Values {
    H = -1.52312760
    C = -2.92375089
    O = 0.79482640
  }
}
```

~~requires #387~~


awvwgk:
We possibly have to do something about the unit of the surface tension currently in dyn/cm.

prog/dftb+/lib_solvation/lebedev.F90
```diff
@@ -0,0 +1,5541 @@
...
+!  Copyright (C) 2006 - 2020  DFTB+ developers group                                               !
+!                                                                                                  !
+!  See the LICENSE file for terms of usage and distribution.                                       !
+!--------------------------------------------------------------------------------------------------!
+
+#:include 'common.fypp'
+
+!> Generator for Lebedev-Laikov grids.
+!>
+!> V.I. Lebedev, and D.N. Laikov, "A quadrature formula for the sphere of the
```
bhourahine, line 12: Is this a fresh implementation from this paper, or is there additional history for these routines? Looks like it might be an offshoot of either the [LGPL routines](https://people.sc.fsu.edu/~jburkardt/f_src/sphere_lebedev_rule/sphere_lebedev_rule.f90) or the van Wuellen [Fortran port](http://www.ccl.net/cca/software/SOURCES/FORTRAN/Lebedev-Laikov-Grids/Lebedev-Laikov.F) of the original C, which carries a requirement on including the reference in publications and manuals (not sure at which stage the jburkardt implementation removed/re-implemented this).

doc/dftb+/manual/dftbp.tex
```diff
@@ -3103,6 +3103,91 @@ \subsubsection{Generalized Born Model}
...
+  \kw{Radii} & m & & & \\
+  \kw{SurfaceTension} & m & & & \\
+\end{ptable}
+
+\begin{description}
+  \item[\is{ProbeRad}] \modif{\modtype{length}}
+    Radius of the probe sphere used to determine the accessible surface area.
+
+  \item[\is{Smoothing}] \modif{\modtype{length}}
+    Smooting parameter for numerical integration.
```
bhourahine, line 3141: typo in smoothing

prog/dftb+/lib_solvation/lebedev.F90
awvwgk: Yes, you are right, I will reference it properly.

bhourahine:
In the sense of a bug or from a usability perspective?

awvwgk:
More from a usability perspective. Eventually it doesn't matter much, because the surface tension is a fit parameter in this model. This implementation allows to copy them directly from the `xtb` parameter files at https://github.com/grimme-lab/gbsa-parameters.

awvwgk:
Rebased against master (e968c7c). The grid generator has been moved to `lib_extlibs`.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -3711,6 +3716,96 @@ subroutine readSolvGB(node, geo, input)
...
+    type(fnode), pointer :: child, value1, value2, field, child2, dummy
+    integer :: iSp, ii, gridPoints
+    real(dp) :: conv
+    real(dp), allocatable :: vdwRadDefault(:)
+
+    if (geo%tPeriodic) then
+      call detailedError(node, "SASA model currently not available with PBCs")
+    end if
+
+    call getChildValue(node, "ProbeRad", input%probeRad, modifier=modifier, child=field)
```
bhourahine, line 3740: ProbeRadius might be clearer for users

awvwgk: Changed it, even if it means that I have to rerender all the parameter files.

doc/dftb+/manual/manual.bib
```diff
@@ -693,6 +693,27 @@ @article{Lee_JCTC_2019
 doi = {10.1021/acs.jctc.9b00132}
 }
 
+@article{im2003,
+author = {Im, Wonpil and Lee, Michael S. and Brooks III, Charles L.},
+title = {Generalized born model with a simple smoothing function},
```
bhourahine, line 698: {B}orn

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -3711,6 +3716,96 @@ subroutine readSolvGB(node, geo, input)
...
+        & modifier=modifier, child=field)
+    call convertByMul(char(modifier), lengthUnits, field, input%smoothingPar)
+
+    call getChildValue(node, "Tolerance", input%tolerance, 1.0e-6_dp, child=child)
+
+    call getChildValue(node, "AngularGrid", gridPoints, 230, child=child)
+    input%gridSize = 0
+    do ii = 1, size(gridSize)
+      if (gridPoints == gridSize(ii)) then
+        input%gridSize = ii
```
bhourahine, line 3753: loop is probably too short to be worth doing, but could add an exit statement

awvwgk: But it is sorted, I could use a bisection search here.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -3711,6 +3716,96 @@ subroutine readSolvGB(node, geo, input)
...
+      end if
+    end do
+    if (input%gridSize == 0) then
+      call detailedError(child, "Illegal number of grid points for numerical integration")
```
bhourahine, line 3757: Usability change for a later pull request would be to suggest the grid nearest to the request value.

awvwgk: Using a bisection search would automatically include this.

prog/dftb+/lib_dftbplus/parser.F90
bhourahine: Its over 32 items, so definitely not time critical. I'd guess the prefactor for bisection might just be larger than just doing the loop comparison.

prog/dftb+/lib_dftbplus/parser.F90
bhourahine: OK, so might as well then. If you've not found the included implementation, see lib_math/bisect.F90

prog/dftb+/lib_dftbplus/parser.F90
awvwgk: I will go with the bisection search. We can default to the next smaller grid in case of an illegal number of grid points. I was also thinking about defaulting to the next larger grid (by inverting the list), but this could turn out to be significantly more expensive.

prog/dftb+/lib_dftbplus/parser.F90
awvwgk: Now this can only happen for 5 or less requested grid points.

PR #407, opened 2020-03-08
Updated INSTALL.rst to reflect preprocessor requirement

bhourahine:
Python is needed to build, not just test.


bhourahine:
Yes, its Python2 capable. But as the rest of the codebase (dp_tools in particular) is Python3, we might as well ask for the supported Python version, rather than the obsolete one.

PR #409, opened 2020-03-09
Test for inverting no lattice vectors

bhourahine:
Minor bug, shows up with -ffpe-trap=invalid gnu compiler option.

### DIFF
--- a/prog/dftb+/lib_poisson/structure.F90
+++ b/prog/dftb+/lib_poisson/structure.F90
@@ -97,11 +97,15 @@ module structure
          endif
          if(boxsiz(i,i)<0.d0) boxsiz(i,:)=-boxsiz(i,:)
       enddo
-  
-      call inversebox()
-    
+
       period=st_isperiodic
-  
+
+      if (period) then
+        call inversebox()
+      else
+        xinvbox(:,:) = 0.0_dp
+      end if
+
       call log_gallocate(x,3,natoms)
       x(1:3,1:natoms)=st_x0(1:3,1:natoms)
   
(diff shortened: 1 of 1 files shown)


aradi:
@pecchia Any objections to get rid of `xinvbox` array and the `coordback` and `difback` routines which use it? Those routines are not called from anywhere apparently.

pecchia:
Hi,

xinvbox is defined in lib_poisson/structure.F90 

Actually now that I look at this I realize the poisson part requires some work for instances…

It is used for the internal ewald summation used in poisson. Sooner or later I will try to fix all that as well.

 

Alex

 

 

Da: Bálint Aradi [mailto:notifications@github.com] 
Inviato: lunedì 9 marzo 2020 16:00
A: dftbplus/dftbplus
Cc: Alessandro Pecchia; Mention
Oggetto: Re: [dftbplus/dftbplus] Test for inverting no lattice vectors (#409)

 

@pecchia <https://github.com/pecchia>  Any objections to get rid of xinvbox array and the coordback and difback routines which use it? Those routines are not called from anywhere apparently.

aradi:
@pecchia Yes, at some point the entire Poisson stuff will need a container, indeed. But for the current fix, we could just delete `xinvbox` and the two routines `difback` and `coordback`, as they are not called from anywhere. I usually prefer to remove dead code if possible.

bhourahine:
At least one of these is used elsewhere, as simply removing them gives a  SIGFPE in the eigensolver routines without NEGF.

bhourahine:
Ah, the SIGFPE is a consequence of testing with -ffpe-trap=invalid suggesting there are other bugs to find elsewhere, but its not due to the removal. So have committed with the routines/variables cut out.

bhourahine:
The floating point error seems to be inside the call to dsyevr in the eigensolvers, so somewhere in openblas on my system.

PR #410, opened 2020-03-09
Fix usage of uninitialized shift vectors with transport

aradi:
Fixes uninitialized usage of shift vectors in transport calculations.

### DIFF
--- a/prog/dftb+/lib_dftb/scc.F90
+++ b/prog/dftb+/lib_dftb/scc.F90
@@ -190,6 +190,9 @@ module dftbp_scc
     !> Coulombic interaction container
     type(TCoulombCont) :: coulombCont
 
+    !> Whether shift vector had been set externally -> skip internal shift calculation
+    logical :: hasExternalShifts
+
   contains
 
     !> Returns a minimal cutoff for the neighbourlist
@@ -386,6 +389,8 @@ contains
     if (this%tH5) then
       this%h5Correction = inp%h5Correction
     end if
+
+    this%hasExternalShifts = .false.
 
     this%tInitialised = .true.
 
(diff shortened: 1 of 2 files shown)


awvwgk:
We should get the CI for transport working, see #395.

prog/dftb+/lib_dftb/scc.F90
```diff
@@ -387,6 +390,8 @@ subroutine Scc_initialize(this, env, inp)
       this%h5Correction = inp%h5Correction
     end if
 
+    this%hasExternalShifts = .false.
```
awvwgk: Shouldn't we get this information from the input? Setting this here to a fixed value and flipping it later without a way to reset it is somewhat hacky.

prog/dftb+/lib_dftb/scc.F90
```diff
@@ -875,7 +890,8 @@ subroutine setShiftPerAtom(this, shift)
 
     @:ASSERT(this%tInitialised)
     @:ASSERT(size(shift) == size(this%shiftPerAtom,dim=1))
-    
+
+    this%hasExternalShifts = .true.
```
awvwgk: I would rather use something like `@:ASSERT(this%hasExternalShifts)` than setting the variable here.

prog/dftb+/lib_dftb/scc.F90
```diff
@@ -894,6 +910,7 @@ subroutine setShiftPerL(this, shift)
     @:ASSERT(size(shift,dim=1) == size(this%shiftPerL,dim=1))
     @:ASSERT(size(shift,dim=2) == size(this%shiftPerL,dim=2))
     
+    this%hasExternalShifts = .true.
```
awvwgk: Same here. It should be a requirement for setting the shift rather than a consequence.

bhourahine:
> This fix looks a bit hacky. Can't we determine from the input if there will be a shift set?

For the moment, getting the transport working again is the urgent thing. A follow up PR can clean up the logic.

PR #411, opened 2020-03-09
Sets Poisson requirement in scc from input

bhourahine:
Acts on  comments from #410


aradi:
OK, but then let's also add the opposite `@:ASSERT ` in `updateShifts()`, as calling it when external shifts are expected is probably an error in the programming logic.

PR #414, opened 2020-03-13
Libpoisson err return

bhourahine:
Converted stop statements in the Poisson solver into calls to the error routine (introducing a preprocessor macro to simplify a bit).
Also partly cleaned up memory logging for the solver (still using module variables, logging files not done cleanly and needs to convert to procedures of a type and open up for other parts of the code).


prog/dftb+/lib_poisson/bulkpot.F90
```diff
@@ -305,24 +314,29 @@ end subroutine destroy_phi_bulk
...
+
   do m = 1, ncont
    
     write(m_id,'(i2.2)') m
     inquire(file='contacts/BulkPot_'//m_id//'.dat',exist=lex)
     if (.not.lex) then
-      write(*,*) 'ERROR: file contacts/BulkPot_'//m_id//'.dat not found'
-      STOP
-    else    
+      @:error_handling(iErr, -1, 'ERROR: file contacts/BulkPot_'//m_id//'.dat not found')
```
awvwgk: Should the `ERROR:` part not be generated by the call to `error`?

prog/dftb+/include/error.fypp
```diff
@@ -0,0 +1,38 @@
...
+#!-------------------------------------------------------------------------------------------------!
+
+#!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+#! Error string handling wrappers for returns from routinese
+#!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+
+#! Macro to return an error flag if return variable available or throw
+#! an error and shut down otherwise
+#:def error_handling(errVar, errNumber, msg)
```
awvwgk: Maybe `ERROR_HANDLING` to make clear it is a preprocessor? Just like `ASSERT`.

PR #415, opened 2020-03-14
Coordination number container

awvwgk:
This PR generalizes the calculation of the coordination number

#### New Features

- [x] Coordination number container in four flavours
  - error function CN used in EEQ model and later GFN0-xTB
  - exponential CN (originial DFT-D3) for GFN1-xTB and DFT-D3
  - covalent CN used in DFT-D4
  - smooth exponential CN used in GFN2-xTB
- [x] reusable input for CN container
- [x] adjusted test inputs and manual entry

#### New Input

Coordination numbers can be generated by

```cson
CoordinationNumber = cov { # or exp/erf/gfn
  CutCN = 0  # for no cutting
  Electronegativities = PaulingEN {}  # only for cov-CN
  Radii = CovalentRadiiD3 {}
}
```

This PR is required for the xTB implementation.

### DIFF
--- a/doc/dftb+/manual/dftbp.tex
+++ b/doc/dftb+/manual/dftbp.tex
@@ -2781,7 +2781,7 @@ can be adjusted in the input.
   \kw{ChargeScale} & r & & 3.0 & \\
   \kw{CutoffInter} & r & & 64 & \\
   \kw{CutoffCount} & r & & 40 & \\
-  \kw{CutoffThree} & r & & 40 & \\
+  \kw{CoordinationNumber} & m & & \is{cov} & \\
   \kw{ChargeModel} & m & & \is{EEQ} & \\
 \end{ptable}
 
@@ -2810,9 +2810,6 @@ can be adjusted in the input.
 \item[\is{CutoffInter}] \modif{\modtype{length}} Cutoff distance when
   calculating two-body interactions.
 
-\item[\is{CutoffCount}] \modif{\modtype{length}} Cutoff distance when
-  calculating coordination numbers.
-
 \item[\is{CutoffThree}] \modif{\modtype{length}} Cutoff distance when
   calculating three-body interactions.
 
@@ -2834,6 +2831,7 @@ up to 86 can be supplied automatically.
   \kw{Cutoff} & r & & 40 & \\
   \kw{EwaldParameter} & r & & 0.0 & \\
   \kw{EwaldTolerance} & r & & 1.0e-9 & \\
+  \kw{CoordinationNumber} & m & & \is{erf} & \\
 \end{ptable}
 
 \begin{description}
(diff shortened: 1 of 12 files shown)


prog/dftb+/lib_dftb/coordinationnumber.F90
```diff
@@ -0,0 +1,844 @@
...
+    real(dp), intent(in) :: coords(:,:)
+
+    !> central cell chemical species
+    integer, intent(in) :: species0(:)
+
+    integer, allocatable :: nNeigh(:)
+
+    allocate(nNeigh(self%nAtom))
+    call getNrOfNeighboursForAll(nNeigh, neighList, self%rCutoff)
+    call getCoordinationNumber(self%nAtom, coords, species0, nNeigh, &
```
aradi: You can use line length up to 100 chars.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -3974,9 +3974,108 @@ subroutine readEeqModel(node, input, geo, nrChrg, kChiDefault, kGamDefault, &
...
+  #:for cnType in ("exp", "erf", "cov", "gfn")
+    case("${cnType}$")
+      input%cnType = cnType%${cnType}$
+  #:endfor
+    end select
+
+    call getChildValue(value1, "CutCN", input%maxCN, cutDefault, &
+        & child=child2)
+
+    call getChildValue(value1, "cutoff", input%rCutoff, 40.0_dp, &
```
aradi: Please capitalize input keyword (e.g. `Cutoff').

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -3974,9 +3974,108 @@ subroutine readEeqModel(node, input, geo, nrChrg, kChiDefault, kGamDefault, &
...
+        & modifier=modifier, child=field)
+    call convertByMul(char(modifier), lengthUnits, field, input%rCutoff)
+
+    allocate(input%en(geo%nSpecies))
+    if (input%cnType == cnType%cov) then
+      call getChildValue(value1, "electronegativities", value2, "PaulingEN", child=child2)
```
aradi: `Electronegativities`

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -3974,9 +3974,108 @@ subroutine readEeqModel(node, input, geo, nrChrg, kChiDefault, kGamDefault, &
...
+    real(dp), allocatable :: kENDefault(:), kRadDefault(:)
+    integer :: iSp
+
+    call getChildValue(node, "CoordinationNumber", value1, cnDefault, child=child)
+    call getNodeName(value1, buffer)
+
+    select case(char(buffer))
+    case default
+      call detailedError(child, "Invalid coordination number type specified")
+  #:for cnType in ("exp", "erf", "cov", "gfn")
```
aradi: Please do not to use pre-processor loop here. You only spare 4 lines of code all together, it is not worth of the obfuscation.

Also, I'd find it more logical to have the `case default` clause at the end (as it would be in an `if / else if / ... / else` construct)

test/prog/dftb+/dispersion/2H2O_dftd4/dftb_in.hsd
```diff
@@ -40,7 +40,14 @@ Hamiltonian = DFTB {
     s9 = 1.0
     s8 = 0.61
     CutoffInter = 60
-    CutoffCount = 34
+    CoordinationNumber = cov {
+       Cutoff = 34
+    }
+    ChargeModel = EEQ {
+       CoordinationNumber = erf {
```
aradi: We should probably capitlize the coordination number containers as well (e.g. `Erf`, `Cov`, ...) Similarly also in the manual then.

prog/dftb+/lib_dftb/coordinationnumber.F90
awvwgk: *Can* or *must*?

prog/dftb+/lib_dftb/coordinationnumber.F90
```diff
@@ -0,0 +1,844 @@
...
+    real(dp), allocatable :: covRad(:)
+
+    !> Electronegativity
+    real(dp), allocatable :: en(:)
+
+  end type TCNInput
+
+
+  !> Coordination number container
+  type :: TCNCont
```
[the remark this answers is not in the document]
awvwgk: Yes, I need direct access to some of the components and don't want a getter/setter here. But I will make most of them `private`.

doc/dftb+/manual/dftbp.tex
```diff
@@ -2883,6 +2881,46 @@ \subsubsection{DftD4 ChargeModel}
 }
```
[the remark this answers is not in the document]
awvwgk: I'm a bit concerned about speed and readability here. How fast is building the input when I also allow modifiers? Also I don't want to overload the input by attaching modifiers to each species symbol. The alternative to attach the modifier to the parent turned out to be fragile in case we use a hybrid of defaults and user values.

prog/dftb+/lib_dftbplus/parser.F90
awvwgk: I prefer to have the default on-top because it usually contains the error handling, which I find easier to understand when it is close to the failing condition. I also negate most of my `if` constructs to reflect this behaviour, so the if/then/else is not really an argument, at least for me.

doc/dftb+/manual/dftbp.tex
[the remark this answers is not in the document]
awvwgk: The unit of the chemical hardness should be energy per square-charge, because it is the second derivative of the energy w.r.t. the population. Not sure if you really want to have units on that one.

awvwgk:
The webhook jammed again? Test is passing, report: https://travis-ci.org/github/dftbplus/dftbplus/builds/667155857.
Maybe we should migrate to travis-ci.com?

prog/dftb+/lib_dftb/coordinationnumber.F90
aradi: We have several old parts which are still 80 width formatted. In newer contributions I prefer 100 characters.

prog/dftb+/lib_dftbplus/parser.F90
aradi: Makes sense, I can live with that logic as well. :wink:

PR #426, opened 2020-03-26
Add GPU-options to CMake build

aradi:
Trivial change with hand-specified options as we have for all other extensions as well. I am working on a more automatized version, but that should follow in a later PR.


aradi:
I've forgot to add the option to `config.cmake`, now it should be fine. :wink:

PR #427, opened 2020-03-26
Updated to new paper citation

bhourahine:


README.rst
```diff
@@ -2,10 +2,12 @@
...
-DFTB+ is a software package for carrying out fast quantum mechanical atomistic
-calculations based on the Density Functional Tight Binding method. It can be
-either used as a standalone program or integrated into other software packages
-as a library.
+DFTB+ is a software package for carrying out fast quantum mechanical
+atomistic calculations based on the Density Functional Tight Binding
+method. It can be either used as a standalone program or integrated
+into other software packages as a library. The most recent features of
+the code are described in this recent `paper
+<https://doi.org/10.1063/1.5143190>`_.
```
awvwgk: Two times recent, maybe shorten it to:
```rst
... are described `here <https://doi.org/10.1063/1.5143190>`_
```

Also I not a huge fan of calling QC program codes. It is called software package above, which is clearer.

PR #428, opened 2020-03-27
Implement ctypes interface for DFTB+

vanderhe:
An implementation of an interface to perform DFTB+ calculations out of Python, using the foreign function library ctypes.

Missing features:

- [x] proper tests, that have to be integrated into the existing test system

- [ ] callback function for population dependant external potential


tools/pythonapi/dftbplus.py
```diff
@@ -0,0 +1,274 @@
...
+            self._relcoords = None
+
+        if self._periodic:
+            self._dftbpluslib.dftbp_set_coords_and_lattice_vecs(
+                self._dftb_handler, self._coords, self._latvecs)
+        else:
+            self._dftbpluslib.dftbp_set_coords(self._dftb_handler, self._coords)
+
+
+    def set_ext_pot(self, extpot, extpotgrad=None):
```
aradi, line 133: Please use a name which corresponds to the on in the C-interface (e.g. `set_external_potential`)

tools/pythonapi/dftbplus.py
```diff
@@ -0,0 +1,274 @@
...
+        Args:
+            extpot (2darray):     External potential at the position of each
+                                  atom. Shape: [natom]. (in atomic units)
+            extpotgrad (2darray): Gradient of the external potential at each
+                                  atom. Shape: [natom, 3]. (in atomic units)
+                                  This parameter is optional, you can pass None
+                                  if you did not ask DFTB+ to calculate forces.
+        '''
+
+        self._extpot = extpot
```
aradi, line 145: I think, we should not to store the queried/passed entities in `self`. It is not needed anywhere.

tools/pythonapi/dftbplus.py
vanderhe: Thanks for the comment, I have renamed accordingly (cf. e0400dd).

tools/pythonapi/dftbplus.py
vanderhe: You are right, storing those entities was unnecessary in many cases. I have changed this (cf. e0400dd).

test/api/pyapi/testcases/extpot/test_extpot.py
```diff
@@ -0,0 +1,103 @@
...
+set an external potential and its gradients.
+'''
+
+
+import numpy as np
+import dftbplus
+from testhelpers import write_autotest_tag
+
+
+LIB_PATH = '../../../../../prog/dftb+/libdftbplus.so'
```
awvwgk, line 19: The extension of library is particular system dependent, there is no guarantee that it will end on `so`. If the build is performed on OSX we will most probably have a `dynlib` here.

awvwgk: The `numpy.ctypeslib.load_library` wrapper for `CDLL` might be worth a try for this purpose, it uses `numpy.distutils.misc_util.get_shared_lib_extension` and `distutils` under the hood to figure out the operating system and select the right extension.

Even better, at least for the test, would be to pass the absolute path to the library from the CMake build system to the tester, since this is the only way to know for sure the right path is actually used.

vanderhe: As a Linux-only user, I honestly forgot that compatibility with Win, OSX,... could be a thing. I'm actually not sure at all whether compiling is possible under these operating systems, but maintaining compatibility will definitely not harm.

Thanks for pointing this out, I'll try the suggested Numpy wrapper.

vanderhe:
Since the basic features are built in and it is not yet clear to me whether the implementation of suitable callbacks (which I am currently working on) via ctypes is possible, I am now releasing this pull request for review.

The use of the interface is quite simple and very similar to that of the already existing C interface. Nevertheless, I would write corresponding recipes entries, which could be particularly helpful for the callback functions (if they should work). It is anticipated that these functions are not self-explanatory.

test/api/pyapi/testcases/extpot/test_extpot.py
```diff
@@ -0,0 +1,103 @@
...
+    print('(H2O) Obtained nr. of atoms: {:d}'.format(natoms))
+    print('(H2O) Expected nr. of atoms: {:d}\n'.format(NATOM0))
+
+    # evaluate mermin free energy
+    print('(H2O) Obtained Mermin-energy: {:15.10f}'.format(merminen))
+    print('(H2O) Expected Mermin-energy: {:15.10f}\n'.format(-3.9854803392))
+
+    # evaluate gradients
+    print('(H2O) Obtained gradient of atom 1: {:15.10f} {:15.10f} {:15.10f}'
+          .format(gradients[0, 0], gradients[0, 1], gradients[0, 2]))
```
aradi, line 76: You could use `.format(*gradients[0])` which is considerably shorter. Also for the other similar cases.

tools/pythonapi/dftbplus.py
```diff
@@ -0,0 +1,253 @@
...
+    def get_energy(self):
+        '''Performs the energy (Mermin free) calculation
+           and queries the energy of the current geometry.
+
+        Returns:
+            energy[0] (float): calculated Mermin free energy
+                               (in atomic units)
+        '''
+
+        energy = np.zeros(1)
```
aradi, line 141: `np.empty()`?

test/api/pyapi/testcases/extpot/test_extpot.py
vanderhe: Applied in 57a8873581acd0755186f6aeb363fd3feca0aab3

tools/pythonapi/dftbplus.py
vanderhe: Applied in 57a8873581acd0755186f6aeb363fd3feca0aab3

PR #434, opened 2020-04-03
Degeneracy module added and used in printing

bhourahine:
Start of specific identification of degeneracy in excited state routines. Currently only used for printing purposes.


prog/dftb+/lib_math/degeneracy.F90
```diff
@@ -0,0 +1,263 @@
+!--------------------------------------------------------------------------------------------------!
+!  DFTB+: general package for performing fast atomistic simulations                                !
+!  Copyright (C) 2006 - 2020  DFTB+ developers group                                               !
+!                                                                                                  !
+!  See the LICENSE file for terms of usage and distribution.                                       !
+!--------------------------------------------------------------------------------------------------!
+
+#:include 'common.fypp'
+
+module dftbp_degeneracyDetect
```
awvwgk, line 10: I'm not too keen starting with camelCase in module names. Also degeneracyDectect is more than the 15 character limit from the style guide. Maybe use just degeneracy or detectdeg here?

prog/dftb+/lib_math/degeneracy.F90
```diff
@@ -0,0 +1,263 @@
...
+    class(TDegeneracyDetection), intent(out) :: self
+
+    !> Tolerance for degeneracy testing
+    real(dp), intent(in), optional :: tolerance
+
+    if (present(tolerance)) then
+      self%tolerance = tolerance
+    else
+      ! a few times eps, just in case of minor symmetry breaking
+      self%tolerance = 128.0_dp*epsilon(0.0_dp)
```
awvwgk, line 71: Maybe make it a module parameter (`toleranceDefault`)?

prog/dftb+/lib_math/degeneracy.F90
```diff
@@ -0,0 +1,263 @@
...
+    integer, intent(in), optional :: levelSubRange(2)
+    
+    integer :: ii, jj, nOrb, iStart, iEnd
+    real(dp) :: localTol
+
+    if (present(tol)) then
+      localTol = tol
+    else
+      ! a few times eps, just in case of minor symmetry breaking
+      localTol = 128.0_dp * epsilon(0.0_dp)
```
awvwgk, line 211: Same constant as in the initialization, just make it a parameter.

prog/dftb+/lib_math/degeneracy.F90
```diff
@@ -0,0 +1,263 @@
...
+      ! set states before group as not of interest
+      iStart = levelSubRange(1)
+      iEnd = levelSubRange(2)
+    else
+      ! full range
+      iStart = 1
+      iEnd = nOrb
+    end if
+    if (iStart >= iEnd) then
+      call error("Degeneracy range is broken, should not be here")
```
awvwgk, line 226: Isn't this more suitable for an assertion macro?

prog/dftb+/lib_math/degeneracy.F90
```diff
@@ -222,9 +223,7 @@ subroutine degeneracyRanges(levelRange, nGrp, Levels, tol, grpMembership, levelS
       iStart = 1
       iEnd = nOrb
     end if
-    if (iStart >= iEnd) then
-      call error("Degeneracy range is broken, should not be here")
-    end if
+  @:ASSERT(iStart <= iEnd)
```
awvwgk, line 226: The negation of `iStart >= iEnd` would be `iStart < iEnd`. What is the reason for using less equal here?

bhourahine: A range of x:x is OK for a single non-degenerate level, so the original version of the test is broken.

PR #435, opened 2020-04-09
Uses passed prerun arguments

bhourahine:
The tests then become thread aware if compiled with openMP

### DIFF
--- a/test/api/mm/testcases/extcharges/testrun.sh
+++ b/test/api/mm/testcases/extcharges/testrun.sh
@@ -1,1 +1,2 @@
-../../testers/test_extcharges
+RUN_CMD=$*
+$RUN_CMD ../../testers/test_extcharges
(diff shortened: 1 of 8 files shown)


bhourahine:
As its a stopgap fix for a PR to do something like [this](https://gist.github.com/jehiah/855086) and matches other script usage, I didn't bother using "$@", but will change them over.

awvwgk:
Yeah, still looks hacky, would be nice if CMake could handle this directly. Exporting an environment shouldn't be to much to ask from a build system, at least in meson it is pretty straight-forward.

There seems to be at least something in the CMake documentation, that looks exactly like the right tool for the job: https://cmake.org/cmake/help/latest/prop_test/ENVIRONMENT.html

bhourahine:
That would require a re-write of the test framework, as it currently all funnels through autotest2 which is command line driven and handles the test return. So probably one for a later PR.

PR #437, opened 2020-04-11
Temp profile fix

### LINKED ISSUE #436: Temperature profiles
Both linear and exponential profiles set temperatures to zero if first in the profile list.

bhourahine:
Fixes initial temperature profile steps that are not constant. Fixes #436


prog/dftb+/lib_md/tempprofile.F90
```diff
@@ -161,33 +155,55 @@ subroutine TempProfile_next(self)
...
+    character(*), intent(in) :: profileName
+
+    !> was the profile correctly identified
+    logical, intent(out) :: success
+
+    integer :: ii
+
+    success = .false.
+    do ii = 1, size(temperatureProfileNames)
+      if (trim(profileName) == tolower(trim(temperatureProfileNames(ii)))) then
```
awvwgk, line 200: A `select case` would work as well and avoid the obscure loop here, also we would not have to rely on the fact that the enumerator corresponds to a loop counter.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -4146,16 +4140,16 @@ subroutine readTemperatureProfile(node, modifier, ctrl)
...
     call destruct(lr1)
     allocate(ctrl%tempMethods(size(tmpC1)))
-    lp2: do ii = 1, size(tmpC1)
-      do jj = 1, size(tempMethodNames)
-        if (trim(tmpC1(ii)) == tolower(trim(tempMethodNames(jj)))) then
-          ctrl%tempMethods(ii) = jj
-          cycle lp2
+    do ii = 1, size(tmpC1)
+      block
+        logical :: success
```
awvwgk, line 4145: Is this really worth introducing a `block` here? We already refactored the parser to have local scopes by subroutines and since we are in `readTemperatureProfile` I don't see a need for a separate `block` here.

PR #440, opened 2020-04-15
Partial fix for d4 crash

bhourahine:
Error retun in the lapack wrapper was broken, so segfaulted instead
of reporting error on matrix factorization of matrix of equations.
Note, still does not address why DFTD4 was generating a singular
matrix with nag7 compiler.

### DIFF
--- a/prog/dftb+/lib_math/lapackroutines.F90
+++ b/prog/dftb+/lib_math/lapackroutines.F90
@@ -639,16 +639,30 @@ contains
     nn = size(aa, dim=1)
     allocate(ipiv(nn))
     call sytrf(aa, ipiv, uplo, info0)
-    if (info0 == 0) then
-      call sytri(aa, ipiv, uplo, info0)
+
+    if (info0 /= 0) then
+      write(error_string, "(A,I10)") "Matrix inversion failed because of &
+          &error in sytrf. Info flag:", info0
+      if (present(info)) then
+        call warning(error_string)
+        info = info0
+        return
+      else
+        call error(error_string)
+      end if
     end if
 
-    if (present(info)) then
-      info = info0
-    elseif (info0 /= 0) then
+    call sytri(aa, ipiv, uplo, info0)
+
+    if (info0 /= 0) then
       write(error_string, "(A,I10)") "Matrix inversion failed because of &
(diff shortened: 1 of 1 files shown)


awvwgk:
The easiest way to have a singular matrix from the EEQ is by forgetting the Lagrangian:

https://github.com/dftbplus/dftbplus/blob/92871ac0b54ba059717ad8b5e1ab17d60401aa96/prog/dftb%2B/lib_dftb/encharges.F90#L1058-L1060

Might there be an issue with the nag-compiler assigning a scalar to a slice? We could explicitly make it an array assignment  by spreading the value to the correct dimension with:

```patch
diff --git a/prog/dftb+/lib_dftb/encharges.F90 b/prog/dftb+/lib_dftb/encharges.F90
index ec3c7ef7..eb215352 100644
--- a/prog/dftb+/lib_dftb/encharges.F90
+++ b/prog/dftb+/lib_dftb/encharges.F90
@@ -1055,8 +1055,8 @@ contains
     else
       call getCoulombMatrixCluster(nAtom, coords, species, gam, rad, aMat)
     end if
-    aMat(nDim, 1:nAtom) = 1.0_dp
-    aMat(1:nAtom, nDim) = 1.0_dp
+    aMat(nDim, 1:nAtom) = spread(1.0_dp, 1, nAtom)
+    aMat(1:nAtom, nDim) = spread(1.0_dp, 1, nAtom)
     aMat(nDim, nDim) = 0.0_dp
 
     ! Step 3: invert linear system
```

bhourahine:
Its an openMP issue, have a fix (which will be checked in as another PR shortly, but I don't fully understand what the nag compiler did not like there).
More generally, its numerically better to solve A x = y directly (something like ?sysvxx solves to machine precision most of the time even if not full rank or fairly ill conditioned) instead of y = A^-1 x. (Started work on a branch for this a while ago but didn't get time to do much).

awvwgk:
The main disadvantage we saw is that the number of right hand side vectors will become fairly large for the response (3*n+9), therefore we decided to go with the inversion. Exchanging it for a solver is fairly easy, with some tricks (read `procedure, pointer`) we can even have both and benchmark which is faster and more stable.

PR #441, opened 2020-04-15
Fix for nag openMP issue in encharge

bhourahine:
Compiler problem with more than one thread, leading to matrix of block form
0 I
I 0


bhourahine:
See my [ompEnchargeNone](https://github.com/bhourahine/dftbplus/tree/ompEnchargeNone) branch for a default none that might work (regression tests running now, will switch it in as a PR if it passes everything).

bhourahine:
Addressed in #442 using default(none).

PR #442, opened 2020-04-15
Omp encharge none

bhourahine:
Fix for OMP with nag7 compiler.

### DIFF
--- a/prog/dftb+/lib_dftb/encharges.F90
+++ b/prog/dftb+/lib_dftb/encharges.F90
@@ -714,6 +714,7 @@ contains
     @:ASSERT(volume > 0.0_dp)
 
     ! Reciprocal space part of the Ewald sum.
+    ! Workaround:nagfor 7.0 with combined DO and PARALLEL
     !$OMP PARALLEL DO DEFAULT(NONE) REDUCTION(+:aMat)&
     !$OMP& SHARED(nAtom, alpha, volume, coords, recPoint)&
     !$OMP& PRIVATE(iAt2f, vec, rTerm)&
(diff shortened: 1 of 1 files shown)


awvwgk:
Looking up the [omp 3.1 specs](https://www.openmp.org/wp-content/uploads/OpenMP3.1.pdf) at 2.6.1 line 2:
> The parallel loop construct is a shortcut for specifying a `parallel` construct containing one or more associated loops and no other statements.

This looks like a nag7 bug to me, but I cannot find a bug tracker for this compiler to see if it already has been reported.

bhourahine:
Yes, it does look that way. The only significant difference to other loops in that file is reduction over one matrix instead of several and a change in line continuations. The apparently harmless
program test
  implicit none
  real, allocatable :: aMat(:,:)
  integer :: ii, jj, n

  read(*,*)n
  allocate(aMat(n,n))
  aMat(:,:) = 0.0

  !$OMP PARALLEL DEFAULT(NONE) REDUCTION(+:aMat) SHARED(n) PRIVATE(jj)
  !$OMP DO SCHEDULE(RUNTIME)
  do ii = 1, n
    do jj = 1, n
      aMat(jj, ii) = ii * jj
    end do
  end do
  !$OMP END DO
  !$OMP END PARALLEL
  write(*,*)aMat
  
end program test

dies with
Runtime Error: test.f90, line 18: Invalid deallocation of Z'1510614BF0D0' size 32: not an allocated block
Also the aMat is apparently treated as a private variable, so something happens at deallocating the thread copies at the OMP END PARALLEL

awvwgk:
Let's merge the “fix” for this issue and file a bug report for the nag7 compiler.

aradi:
I'll file the bug report to the NAG-support, usually they are quite fast in responding. (And no, they do not have a public bug-tracker as far as I know...)

PR #443, opened 2020-04-18
Create abstract base class for existing X-bond correction

awvwgk:
Abstract halogen bond correction to allow adding other classical correction schemes, like the GFN1-xTB halogen bond correction and the GFN0-xTB polar bond correction.

### DIFF
--- a/prog/dftb+/lib_dftb/CMakeLists.txt
+++ b/prog/dftb+/lib_dftb/CMakeLists.txt
@@ -3,6 +3,7 @@ set(curdir "lib_dftb")
 set(sources-fpp
   ${curdir}/chargeconstr.F90
   ${curdir}/charges.F90
+  ${curdir}/clscorr.F90
   ${curdir}/coordnumber.F90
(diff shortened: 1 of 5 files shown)


awvwgk:
@aradi @bhourahine The `THalogenX` container does not fit nicely into the usual container scheme due to missing `updateCoords` and `updateLatVecs` interfaces, therefore the evaluation of energies, gradients and stresses requires to perform a calculation on the same energy expression thrice. So it might be worth refactoring this first. Any opinion on this?

prog/dftb+/lib_dftb/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ set(curdir "lib_dftb")
 set(sources-fpp
   ${curdir}/chargeconstr.F90
   ${curdir}/charges.F90
+  ${curdir}/clscorr.F90
```
aradi, line 6: What about `classicalcorr.F90`? Still within the 15 char limit and probably more informative. :wink:

PR #446, opened 2020-04-20
Fix missing free energy shift in GB model

awvwgk:
The free energy shift was missing in the first GB model implementation, this PR amends it and also includes the missing reference states.

- states “reference” (`cosmotherm` option `gsolv=reference`) and “mol1bar” included, naming could be improved, but this way it is consistent with `cosmotherm` and the AMS implementation
- add test for each reference state, just copies of the `gb_ionpair` test with the energy changed
- update parser and manual on reference states
- include GB model in changelog


doc/dftb+/manual/dftbp.tex
```diff
@@ -3048,14 +3048,20 @@ \subsubsection{Generalized Born Model}
...
+    \kw{MolecularMass} and \kw{Density} only affect the calculation if
+    ``reference'' is chosen as state of the solvent.
 
-  \item[\is{Shift}] \modif{\modtype{energy}} Shift for free energy calculations.
+  \item[\is{FreeEnergyShift}] \modif{\modtype{energy}} Shift for free energy
+    calculations.
 
-  %\item[\is{Temperature}] Temperature for free energy calculations. Default is
-    %ambient temperature: 298.15 K.
+  \item[\is{Temperature}] Temperature for free energy calculations. Default is
```
bhourahine, line 3057: Need to mention this is always in Kelvin

doc/dftb+/manual/dftbp.tex
```diff
@@ -3048,14 +3048,20 @@ \subsubsection{Generalized Born Model}
...
+    ambient temperature: 298.15 K.
+    Only affects the calculation for if ``reference'' or ``mol1bar'' is chosen
```
bhourahine, line 3059: Might want to briefly explain what mol1bar corresponds to

doc/dftb+/manual/dftbp.tex
awvwgk: Is there a way to add temperature units to it? I have only found energy units, which might not really be the most suitable choice, unless I can supply the default value + modifier.

bhourahine: I don't see a problem with adding a new category for temperature units in prog/dftb+/lib_common/unitconversion.F90

As you convert it by the Boltzmann constant in the parser, isn't it an energy proxy anyway?

doc/dftb+/manual/dftbp.tex
awvwgk: The safest choice in my opinion is always “gsolv” as state, as you can more easily arrive from there at the really wanted state. Of course you than have to keep in mind that you might have to add the volume work yourself afterwards.

doc/dftb+/manual/dftbp.tex
awvwgk: It would be more convenient to have the temperature in K in the processed input, I find this more intuitive than Hartree. But it is indeed used as energy, I will change it here.

doc/dftb+/manual/dftbp.tex
awvwgk: Just added a short description on the available states.

doc/dftb+/manual/dftbp.tex
```diff
@@ -3031,8 +3031,8 @@ \subsubsection{Generalized Born Model}
 \begin{ptable}
   \kw{Solvent} & m & & & \\
   \kw{FreeEnergyShift} & r & & & \\
-  %\kw{Temperature} & r & & 298.15 & \\
-  %\kw{State} & s & & gsolv & \\
+  \kw{Temperature} & r & & 298.15 & \\
```
bhourahine, line 3034: Needs a unit on the specified default value

awvwgk:
Is this ready for a merge now?

PR #392, opened 2020-02-25
Convert types to TType convention

bhourahine:
Also changed two colliding logical variables from tGeoOpt to
isGeoOpt and tLinResp to isLinResp

PR #449, opened 2020-04-23
Change REKS input setting

inseonglee:
This PR is WIP.

Change REKS input block in dftb_in.hsd.
- [x] Enumeration of REKS types
- [x] Replace some inputs by logical variables (e.g. SaveMemory, ReadEigenvectors)
- [x] Add Energy and Gradient sub-blocks for convenience
- [x] Modify dftb_in.hsd for all REKS test calculations
- [x] Modify documentation for REKS part

For test files, I will change inputs after the modifications are finished.


inseonglee:
@aradi @bhourahine I finished to modify REKS input arguments for convenience. The main purpose is to replace the integer type variables (slightly confused) by logical or enumeration types (intuitive) so that the users can easily exploit the REKS functionality. Please review this PR.

Currently, only remaining works are to modify documentation related to REKS part. This will be done after all modifications are complete.

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -4726,7 +4726,7 @@ subroutine checkReksConsistency(reksIni, onSiteElements, kPoint, nEl, nKPoint, t
       end if
     end if
 
-    if (reksIni%Efunction /= 1 .and. tLatOpt) then
+    if (reksInp%Efunction /= 1 .and. tLatOpt) then
```
bhourahine, line 4729: Is there a reason not to also use an enumerated type for the energy function?

inseonglee: In `Energy` block of current input, it reads string list such as {"PPS" "OSS"}. This options can be different when we use SSR(4,4) case, which means I think that we have to increase enumerated types when the active space size increases.

For example, SSR(4,4) can use the following functionals.
1) only PPS (which is single-state REKS in (4,4) case)
2) PPS + OSS1 + OSS2,  3) PPS + DSPS,  4) and so on.

Thus, the second option in SSR(2,2) is PPS + OSS, but the second one in SSR(4,4) becomes PPS + OSS1 + OSS2. So, I do not use enumerated type for `reksInp%Efunction` at the moment.

inseonglee: @bhourahine How about this? I think that we can sort the enumerated type like this. The ten-th digit indicates the active space size (kind) and the digit indicates the possible choice. For example, certain enumerated type `reksEfunc` = 0 for PPS and = 1 for PPS + OSS in SSR(2,2) case. Likewise, `reksEfunc` = 10 for PPS and 11 for PPS + OSS1 + OSS2 and 12 for PPS + DSPS and so on. If you want to agree this opinion, I will try to change the `Efunction` to `reksEfunc` enumerated type.

bhourahine: Seems reasonable, but assumptions about order or additivity of the constants outside of the dedicated reks routines should be removed.

inseonglee: Then I will maintain current status for this parser routine. Is there other comments? If not I will change the documentation from now on.

prog/dftb+/lib_dftbplus/main.F90
```diff
@@ -7519,8 +7519,8 @@ subroutine getReksEnProperties(eigenvecs, coord0, reks)
 
       call getUnrelaxedDensMatAndTdp(eigenvecs(:,:,1), reks%overSqr, reks%rhoSqrL, &
```
bhourahine, line 7520: Please pass reks instead of its components (the main code should know as little as possible about details of various extensions, as this more robust for future changes and extensions).

inseonglee: @bhourahine I see, but I will fix this with other PR indicating refactoring for some REKS routines (I will create new PR after this PR is merged).

inseonglee:
@aradi @bhourahine I am still waiting your response for this REKS input. If there is no more fix, then I will finish the modification for REKS documentation within this week.

inseonglee:
@aradi @bhourahine Now, I finished to modify the manual parts of REKS method. Please review this PR.

doc/dftb+/manual/dftbp.tex
```diff
@@ -3911,133 +3914,163 @@ \subsection{SSR22}
...
+  \kw{ReadEigenvectors} & l & & No \\
+  \kw{FonMaxIter} & i & & 20 & \\
+  \kw{Shift} & r & & 0.3 & \\
+  \kw{SpinTuning} & $(\text{r})^*$ & & \cb & \\
+  \kw{TransitionDipole} & l & Functional $\neq$ \{ "PPS" \}, TargetMicrostate = 0 & No & \\
+  \kw{Gradient} & m & & ConjugateGradient \cb & \\
+  \kw{RelaxedDensity} & l & & No & \\
+  \kw{NonAdiabaticCoupling} & l & Functional $\neq$ \{ "PPS" \}, StateInteractions = Yes & No & \\
+  \kw{VerbosityLevel} & i & & 1 & \\
+\end{ptable}
```
inseonglee, line 3948: In the pdf file, this part shows too long length for the width. Is there a good idea for this table?

bhourahine: A parbox might solve this for the moment, something like
\parbox{0.3\textwidth}{Functional $\neq$ \{ "PPS" \}, TargetMicrostate = 0}

In the longer term, we probably should re-define the table macro to set that column as a p{width}

bhourahine:
OK, have brought this up to date with the master repository and will run some regression tests now.

doc/dftb+/manual/dftbp.tex
inseonglee: Thanks. This problem is solved by using \parbox. Now, it shows normal length for the width in REKS keyword tables.

prog/dftb+/lib_reks/reksgrad.F90
```diff
@@ -571,14 +572,14 @@ subroutine getHxcKernel(iSquare, getAtomIndex, getDenseAO, over, overSqr, &
...
     !> Algorithms to calculate analytic gradients
     integer, intent(in) :: Glevel
 
-    !> Memory level used in calculation of gradient
-    integer, intent(in) :: Mlevel
+    !> Save 'A' and 'Hxc' to memory in gradient calculation
+    logical, intent(in) :: tSaveMem
+
+    !> Whether to run a range separated calculation
+    logical, intent(in) :: tRangeSep
```
bhourahine, line 582: Can you rename tRangeSep to isRangeSep. From #392 onwards TRangeSep is now a type name elsewhere.

doc/dftb+/manual/dftbp.tex
```diff
@@ -4121,133 +4124,163 @@ \subsection{SSR22}
...
 
-  \item[\is{MemoryLevel}] Specifies the memory allocation of the variables used in CP-REKS equations.
-    First option shows fast computation speed but it requires large memory allocation, increasing as
-    $O(N_\mathrm{basis}^4)$. Second option shows relatively slow computation speed but it directly calculate
-    the CP-REKS variables without saving the variables. In general, the second option is recommended for large systems.
+    \item[\is{SaveMemory}] If set to \is{Yes}, some variables (A orbital hessian matrix and H-xc kernel) which need large
+      memory allocation are saved in the memory. If these variables are saved, then the computational
+      speed also increases but it shows large memory allocation, increasing as $O(N_\mathrm{basis}^4)$.
+      If set to \is{No}, the CP-REKS equations are solved without saving these variables, thus it is relatively
+      slower than the case that you set to \is{Yes}. In general, \is{No} option is recommended for the large systems.
```
bhourahine, line 4243: "for large systems."

doc/dftb+/manual/dftbp.tex
```diff
@@ -4121,133 +4124,163 @@ \subsection{SSR22}
...
-    energy information (\is{PrintLevel} = 0), gradient information (\is{PrintLevel} = 1), detailed information about
-    SCC cycle and timing in gradient calculation (\is{PrintLevel} = 2).
+    \item[\is{Preconditioner}] If set to \is{Yes}, it uses a preconditiner in the conjugate-gradient based algorithm. In general,
+      the convergence speed is increased when this option sets to \is{Yes}, thus this option is recommended.
 
-  \item[\is{MemoryLevel}] Specifies the memory allocation of the variables used in CP-REKS equations.
-    First option shows fast computation speed but it requires large memory allocation, increasing as
-    $O(N_\mathrm{basis}^4)$. Second option shows relatively slow computation speed but it directly calculate
-    the CP-REKS variables without saving the variables. In general, the second option is recommended for large systems.
+    \item[\is{SaveMemory}] If set to \is{Yes}, some variables (A orbital hessian matrix and H-xc kernel) which need large
```
bhourahine, line 4239: "(an orbital hessian matrix and the H-xc kernel)"

prog/dftb+/lib_reks/reksgrad.F90
inseonglee: Done with commit 8467c28. Please see this commit.
Here, there are remaining parts using `tRangeSep` in main.F90, mainio.F90, initprogram.F90 files.
At the moment, I do not modify the parts that are not related to REKS.

doc/dftb+/manual/dftbp.tex
inseonglee: Done with commit 65d0cb5.

doc/dftb+/manual/dftbp.tex
inseonglee: Done with commit 65d0cb5.

PR #451, opened 2020-04-25
Add include path for Ford processing

bhourahine:
Fixes a missing path for ford processing of D4 code


bhourahine:
Doxygen uses an internal c-like preprocessor (there doesn't seem to be a method to externally specify an alternative), so is unable to deal with fypp constructs and so is currently unusable anyway.

This could presumably be worked around by providing a mechanism to externally pre-process the source tree and then run doxygen on the result.

awvwgk:
Or maybe just drop doxygen in favour of FORD, I don't really see a point in using two documentation generators, especially if one is somewhat broken.

bhourahine:
Ford doesn't cope with C code, so is broken in a different sense.

PR #455, opened 2020-04-27
Removed flag not relevant to gcc

bhourahine:


sys/gnu.cmake
```diff
@@ -94,5 +94,5 @@ set(OTHER_INCLUDE_DIRS "" CACHE STRING "Other include directories to consider")
 set(CMAKE_Fortran_FLAGS_DEBUG "-g -Wall -std=f2008ts -pedantic -fbounds-check" CACHE STRING
```
awvwgk, line 94: While we are already on it, from the man page of `gfortran`:

> The deprecated option `-std=f2008ts` acts as an alias for `-std=f2018`. It is only present for backwards compatibility with earlier gfortran versions and should not be used any more.

awvwgk: If it doesn't break with GCC 7.5.0 on Travis CI it should be fine to update this keyword as well.

bhourahine:
Might as well make this atomic: #458

PR #458, opened 2020-04-28
Changed 2008ts to 2018 standard for gfort

bhourahine:

PR #459, opened 2020-04-28
Enable debug builds on PRs

awvwgk:
This setup will perform builds in debug mode on PRs and in release mode on branches.
See https://docs.travis-ci.com/user/pull-requests/ for details.

The branch build for this PR: https://travis-ci.com/github/awvwgk/dftbplus/builds/162396959

#### Timings

MPI | Release | Debug
--- | --- | ---
true | ~9 min | ~ ~~12~~ 10 min
false | ~8 min |  ~ ~~11~~ 8 min

Might require to mark more tests as `LONG_TEST`: 
- `dftb+_md/SiC64-xlbomdfast-T0`
- `dftb+_md/SiC64-xlbomdfast`
- `dftb+_transport/CH4-poisson`
- `dftb+_transport/C-chain+U`
- `dftb+_transport/SiH-chain`
- `dftb+_transport/SiH-chain_semiInf`

#### Conflicts

Does not work with #458 unless we bump the GCC version for Travis-CI, which then does not work with the current toolchain file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,6 @@ install:
 script:
 - mkdir _build
 - cd _build
-- cmake -DWITH_DFTD3=true -DWITH_TRANSPORT=true -DWITH_MPI=${WITH_MPI} -DFYPP_FLAGS="-DTRAVIS" -DCMAKE_TOOLCHAIN_FILE=../sys/gnu.cmake ..
-- make -j 2
-- ctest -j 2
+- if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then BUILD_TYPE=Debug; else; BUILD_TYPE=Release; fi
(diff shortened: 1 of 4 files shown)


awvwgk:
The yaml was pretty hard to read already, therefore I decided we better have a script here.
Just recycled my conda-build recipe for `xtb`: https://github.com/grimme-lab/xtb-feedstock/blob/master/recipe/build.sh
The conda-forge CI is running in an Ubuntu container at Azure DevOps, similar to the Ubuntu image used on Travis-CI, so this should be compatible.

We can of course get rid of the `bash` array to change back to the more POSIX compliant shell, but I found the array for the arguments quite readable.

awvwgk:
Need to work on the timings first, we should be under 10 minutes per build.

awvwgk:
Rebased against master (6bc362c), additional fix for build error in debug mode due invalid format string at

https://github.com/dftbplus/dftbplus/blob/6bc362c4a207ba9c564de0b919d49e8c7c86e3ae/prog/dftb%2B/lib_poisson/bulkpot.F90#L77

PR #461, opened 2020-04-28
Complete GBSA model

### LINKED ISSUE #361: Implicit Solvation Model?
It seems like DFTB+ is missing an implicit solvation model, like a polarizable continuum model (PCM) or a generalized Born (GB) model. Is there interest to have an implicit solvation model in DFTB+?

The [DFTB module of AMS](https://www.scm.com/doc/DFTB/DFTB_Model_Hamiltonian.html#solvation) got recently a GB model. And the `xtb` program is supporting a GB model as well.

awvwgk:
This PR implements the complete GBSA model as currently present in `xtb` (both GFN1-xTB and GFN2-xTB) and the DFTB module in AMS.

This is done in a completely modular way by reusing the SASA container inside the generalized Born model and does not require any additional implementation effort. Additionally the generalized Born model can now use the SASA information for correcting HB effects (+ potential shift to the Hamiltonian).

The output for the GBSA model is now also more detailed

```
Solvation:                   generalized Born model
Dielectric constant:           7.000000E+00
Free energy shift:             4.231604E-03 H      1.151478E-01 eV
Born radii integrator:       GBOBC
SASA model:                  Yes
Grid points:                   5.810000E+04 total          5810 per atom
CM5 correction:              Yes
Hydrogen bond correction:    No
```

A DFTB+ compatible HSD parameter file reflecting the GBSA(H2O) parametrisation for GFN2-xTB will look like this (elements with Z>18 removed for brevity):

```cson
GeneralizedBorn {
  Solvent = fromConstants { # GFN2-xTB/GBSA(H2O)
    Epsilon =   80.20000000
    MolecularMass [amu] =   18.00000000
    Density [kg/L] =    1.00000000
  }
  Temperature [K] = 298.15
  FreeEnergyShift [kcal/mol] =    1.16556316
  BornScale =    1.55243817
  BornOffset =   2.462811043694508E-02
  Radii = vanDerWaalsRadiiD3 {}
  Descreening = Values {
    H  =    0.71893869
    He =    2.06028647
    Li =    0.40608033
    Be =    0.92184652
    B  =    0.77279023
    C  =    0.74298311
    N  =    0.90261230
    O  =    0.75369019
    F  =    0.61625017
    Ne =    1.84783975
    Na =    1.20080317
    Mg =    0.65802384
    Al =    0.72103171
    Si =    0.83603641
    P  =    0.66440982
    S  =    0.89558269
    Cl =    0.62985587
    Ar =    1.27858461
  }
  SASA {
    ProbeRadius =   1.843075777670416E+00
    AngularGrid = 230
    Radii = vanDerWaalsRadiiD3 {}
    SurfaceTension = Values {
      H  =  -3.349830600000000E-01
      He =   1.268718610000000E+00
      Li =  -4.026822030000000E+00
      Be =  -6.613238720000000E+00
      B  =  -6.811420300000000E-01
      C  =  -7.476906500000000E-01
      N  =  -2.312912920000000E+00
      O  =   9.179791100000000E-01
      F  =   1.255433480000000E+00
      Ne =   8.576002100000000E-01
      Na =  -6.741013230000000E+00
      Mg =  -1.190967580000000E+00
      Al =  -9.427821000000000E-02
      Si =   1.539815520000000E+00
      P  =  -2.856607400000000E-01
      S  =  -4.855542000000000E-01
      Cl =  -1.020804500000000E-01
      Ar =   2.701570800000000E-01
  }
  HBondCorr = Yes
  HBondStrength = Values {
      H  =  -7.172800544988973E-02
      He =   0.000000000000000E+00
      Li =   0.000000000000000E+00
      Be =   0.000000000000000E+00
      B  =   0.000000000000000E+00
      C  =  -2.548469535762511E-03
      N  =  -1.976849501504001E-02
      O  =  -8.462476828587280E-03
      F  =  -6.300005896906867E-03
      Ne =   0.000000000000000E+00
      Na =   0.000000000000000E+00
      Mg =   0.000000000000000E+00
      Al =   0.000000000000000E+00
      Si =   0.000000000000000E+00
      P  =  -1.145584928393385E-02
      S  =  -2.333053852509530E-04
      Cl =  -1.688957105919989E-03
      Ar =   0.000000000000000E+00
  }
}
```

HSD parameter files will be provided in https://github.com/grimme-lab/gbsa-parameters, after this PR is merged. To make life easier, this also implements a reader for the `xtb` parameter format.

~~requires #400~~
might resolve #361


awvwgk:
As discussed, I moved the parsing for all solvation related parts of the HSD tree to a separate module.

The good thing is, the parser does not need to import anything except for the `dftbp_solvparser` module from the `lib_solvation` now. This leaves the `dftbp_solvation` as definition of the abstract base class and the `dftbp_solvinput` module as only other entry points necessary to be included elsewhere in DFTB+.

awvwgk:
This is ready up to the point, that we explicitly copy parameter files from https://github.com/grimme-lab/gbsa-parameters. As soon as we agree on the final HSD input keywords, I will merge https://github.com/grimme-lab/gbsa-parameters/pull/2 and release the parameter files, which allows to add them as external dependency to `./utils/get_opt_externals` and cleanup the test directory.

prog/dftb+/lib_solvation/born.F90
```diff
@@ -200,6 +222,16 @@ subroutine initialize(self, input, nAtom, species0, speciesNames, latVecs)
     self%param = input%TGBParameters
     self%rho(:) = input%vdwRad(:) * input%descreening(:)
 
+    if (allocated(self%sasaCont) .and. allocated(input%hBondPar)) then
+      if (any(input%hBondPar /= 0.0_dp)) then
+        allocate(self%hBondStrength(nAtom))
+        do iAt1 = 1, nAtom
+          iSp1 = species0(iAt1)
+          self%hBondStrength(iAt1) = input%hBondPar(iSp1) / self%sasaCont%probeRad(iSp1)**2
```
bhourahine, line 230: Might want to consider a future PR to pre-compute the inverse of the probeRad data. Unless the compiler is very clever it won't work out that there are only a few of these and it can save (x~25) flops with a multiplication by the stored inverses.

prog/dftb+/lib_solvation/born.F90
```diff
@@ -374,7 +471,13 @@ subroutine getStress(self, stress)
...
     @:ASSERT(self%volume > 0.0_dp)
 
-    stress(:,:) = self%stress / self%volume
+    if (allocated(self%sasaCont)) then
+      call self%sasaCont%getStress(stress)
+    else
+      stress(:, :) = 0.0_dp
+    end if
+
+    stress(:,:) = stress + self%stress / self%volume
```
bhourahine, line 480: Isn't self%stress actually self%virial at this point? Might consider re-naming, if we're going to tidy all of these up.

doc/dftb+/manual/dftbp.tex
```diff
@@ -3028,7 +3028,10 @@ \subsubsection{Generalized Born Model}
 
 To use the generalized Born model in the SCC procedure use the
 \is{GeneralizedBorn\cb} method in the input to \kw{Solvation}.
-Note that the GB model implemented is only available for finite
+The non-polar solvent area model can be combinded with the GB model
+enabling to additionally correct for hydrogen bonding, the resulting
+model is called GBSA.
+Note that the GB(SA) model implemented is only available for finite
```
bhourahine, line 3034: Probably also a good idea to add a line giving the location of the [repository](https://github.com/grimme-lab/gbsa-parameters) and explaining the state of fitting (since the regression tests and get_external are using data from there).

bhourahine: Something like:
Parameters for the model, as fitted for the xtb GFN hamiltonians, are available \href{https://github.com/grimme-lab/gbsa-parameters}{online}.

test/prog/dftb+/tests
```diff
@@ -133,6 +133,11 @@ solvation/gb_ionpair                   #? MPI_PROCS <= 1
 solvation/gb_reference                 #? MPI_PROCS <= 1
 solvation/gb_mol1bar                   #? MPI_PROCS <= 1
 solvation/gb_cm5                       #? MPI_PROCS <= 1
+solvation/gbsa_e35                     #? MPI_PROCS <= 1
+solvation/gbsa_caffeine                #? MPI_PROCS <= 1
+solvation/gbsa_param1                  #? MPI_PROCS <= 1
+solvation/gbsa_param2                  #? MPI_PROCS <= 1 and not WITH_MPI
```
bhourahine, line 139: just 
#? not WITH_MPI
should work here

prog/dftb+/lib_solvation/solvparser.F90
```diff
@@ -0,0 +1,477 @@
...
+          do iSp = 1, geo%nSpecies
+            call getChildValue(value1, trim(geo%speciesNames(iSp)), input%hBondPar(iSp), child=child2)
+          end do
+        end select
+      end if
+    end if
+
+  end subroutine readSolvGB
+
+
```
bhourahine, line 302: !> Reads in solvent accessible surface area parameters

prog/dftb+/lib_solvation/solvparser.F90
```diff
@@ -0,0 +1,477 @@
...
+    end select
+
+    call getChildValue(node, "Offset", input%sOffset, 2.0_dp * AA__Bohr, &
+        & modifier=modifier, child=field)
+    call convertByMul(char(modifier), lengthUnits, field, input%sOffset)
+
+  end subroutine readSolvSASA
+
+
+  subroutine readCM5(node, input, geo)
```
bhourahine, line 415: "!> Read in CM5 charge model settings"
or similar

prog/dftb+/lib_solvation/born.F90
awvwgk: This is actually storing the inverses in `hBondStrength` for efficiency reasons. I see no need to optimize this part of the constructor, except for memory usage of the `hBondStrength` array, but this version makes it more convenient to use later (also think about a more empirical model for the GBSA where this parameter is indeed atom resolved by some function of the geometry / topology).

prog/dftb+/lib_solvation/born.F90
awvwgk: Usually going for `sigma` but `virial` should work as well.

prog/dftb+/lib_solvation/born.F90
bhourahine: Was actually thinking of this _specific_ implementation, so something like
invRad = 1.0_dp / self%sasaCont%probeRad
do iAt1 = 1, nAtom
as nAtom >> nSpecies most of the time and nSpecies < ~10. But this isn't time or memory critical here, so as you wish.

awvwgk: It is still a step in the constructor, so I don't see the point of requiring the allocation of a scratch array. Maybe one could simplify a bit by a temporary scalar with
```suggestion
          invRad = 1.0_dp / self%sasaCont%probeRad(iSp1)
          self%hBondStrength(iAt1) = input%hBondPar(iSp1) * invRad * invRad
```

bhourahine: The scratch array is tiny and goes out of scope rapidly (and is probably what a sufficiently smart compiler will effectively do). Not seeing the proposed change as either simpler or faster, so happy to stay with original.

prog/dftb+/lib_solvation/born.F90
bhourahine: sigma is good.

awvwgk:
Obviously, the file format is dictated by `xtb` here. We can add a version number, but this will require a change in `xtb` as well.

The other alternative would be #470 and a little `awk` script to convert them to HSD input.

awvwgk:
Also happy to see that my parser for the parameter file is not to convoluted for DFTB+, I was actually expecting a comment on the use of newlines in `error` and `warning` messages.

bhourahine:
> Obviously, the file format is dictated by `xtb` here. We can add a version number, but this will require a change in `xtb` as well.

This is why I mention it as a meta issue, but obviously not a problem at that moment. 
Cross that bridge when you come to it / Kümmere dich nicht um ungelegte Eier.

bhourahine:
> Also happy to see that my parser for the parameter file is not too convoluted for DFTB+, I was actually expecting a comment on the use of newlines in `error` and `warning` messages.

That would be @aradi, not me. Perhaps a topic for the next dev meeting.

doc/dftb+/manual/dftbp.tex
awvwgk: Having the parameters on `dftb.org` as well would also be a nice addition.

PR #464, opened 2020-04-28
Parallelise short gamma construction

bhourahine:
Seems to have large performance improvement on some REKS test cases for multiple threads.

### DIFF
--- a/prog/dftb+/lib_dftb/scc.F90
+++ b/prog/dftb+/lib_dftb/scc.F90
@@ -12,6 +12,7 @@ module dftbp_scc
   use dftbp_assert
   use dftbp_accuracy
   use dftbp_blasroutines
+  use dftbp_schedule
   use dftbp_charges, only : getSummedCharges
   use dftbp_chargeconstr
   use dftbp_commontypes
@@ -464,7 +465,7 @@ contains
 
     call updateNNeigh_(this, species, neighList)
 
-    call initGamma_(this, species, neighList%iNeighbour)
+    call initGamma_(this, env, species, neighList%iNeighbour)
 
     if (allocated(this%extCharge)) then
       if (this%tPeriodic) then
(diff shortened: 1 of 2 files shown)


prog/dftb+/lib_dftb/scc.F90
```diff
@@ -1332,9 +1332,11 @@ subroutine initGamma_(this, env, species, iNeighbour)
     ! some additional symmetry not used in these loops, as the value of gamma for atoms interacting
     ! with themselves is the same for all atoms of the same species
 
+    ! Workaround for bug in Intel18 compiler can not use OMP REDUCTION(:+work) in debug mode
```
awvwgk, line 1335: This is not a bug in Intel18, I don't see a + reduction on `work` in your implementation, it _should_ be shared.

bhourahine: See the previous commit [6faf7d36](https://github.com/bhourahine/dftbplus/commit/6faf7d36c459b639946dfc97ded52c58dfa276ab)
which works for intel19 but not 18, weirdly showing up as a segfault on a test using GPUs.

awvwgk: Weird segfault is usually an omp stacksize problem. The GPU testcases are huge and an omp reduction can be quite wasteful with memory, especially when it is not needed in this context. We could try it again and increase the `OMP_STACKSIZE` environment variable, which is usually set by default to a (too) small value. Point is, that is not a bug, but some weird default behaviour of the compiler (which tend to change between releases, unfortunately).

bhourahine: The matrix being reduced is quite small (setting the OMP stack size to 16M is enough to avoid the segfault). Removing the reduction operation runs a little bit slower, at least on two threads (due to a difference in cache behaviour presumably, since each thread isn't needing to write against the same matrix until the reduce at the end) but avoids that segfault.

awvwgk: If you ever find a way to set (or even check) the stacksize from Fortran, let me know.

bhourahine:
> Is there no performance benefit for regular DFTB? What makes REKS special?

@thomas-niehaus  Yes there is some speed up with a few of the regression tests, but its < 10% in the best cases on 2 threads as compared to ~40% for some of the REKS cases. I assume REKS is either 1) repeatedly re-building this matrix or 2) its better OMP parallel than the standard code, so notices this routine, or 3) something went wrong with the reference timings.

bhourahine:
Test with intel19 and ggc9 don't show substantial performance improvements, so withdrawing PR

Issue #389, opened 2020-02-24
Restart for hessian matrix calculation


bhourahine:
Currently there is no way to do this. Alternatively, optional calculation of selected rows of the hessian and a tool to assemble them after would also work (and allow for better parallelisation).

bhourahine:
#465 addresses this

PR #465, opened 2020-04-29
Parallel hessian

pecchia:
This feature allows to compute the Hessian splitting the job over several subsections.
This is done by adding into block SecondDerivatives{ computedAtoms=start:end }
It was necessary in Dresden as the HPC queue was limiting maximum running times.
In principles different portions can be computed in parallel if wrapped by an appropriate script.

- [x] A test has been added. 
- [ ] Documentation to be added.
- [x] Fix the changed submodules
- [ ] example script for combining outputs


bhourahine:
Can you add an example script, as this would address #389 ?

prog/dftb+/CMakeLists.txt
```diff
@@ -77,7 +77,7 @@ if(WITH_DFTD3)
 endif()
 
 if(WITH_TRANSPORT)
-  target_link_libraries(dftbplus PUBLIC negf sparskit mudpack)
```
bhourahine, line 80: Is this on top of the changes in #463 ?

pecchia: This is the usual problems I have with submodules... I was working on another branch, then I checked out the libnegf head which requires this change and kept it when creating this new branch. If we do not really like it here I can checkout the old libnegf revision. I was wandering if I can squash some commit in the github remote to cleanup history.

bhourahine: If its merged after #463 there shouldn't be a problem.

PR #467, opened 2020-04-30
Lammps integration

gsjung0419:
Dear administrator,

I am developing a code for lammps to use dftb+. 

In order to better use, I added some functions in the api. 

One is to get repulsive and electronic energy separately. 

The other is to get virial term to use in lammps. 

Pleas let me know if there is further modification is needed. 

Thank you.

Best,

GS

### DIFF
--- a/prog/dftb+/lib_dftbplus/mainapi.F90
+++ b/prog/dftb+/lib_dftbplus/mainapi.F90
@@ -128,6 +128,11 @@ contains
     !> resulting gradients wrt atom positions
     real(dp), intent(out) :: virial(:,:)
 
+    if (.not. tStress) then
+      call error("Virial not available, you must initialise your calculator with&
+          & this property enabled.")
+    end if
+
     call recalcGeometry(env)
     virial(:,:) = totalStress*CellVol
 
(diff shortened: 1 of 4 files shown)


prog/dftb+/lib_dftbplus/mainapi.F90
```diff
@@ -104,6 +120,18 @@ subroutine getStressTensor(env, stress)
...
+   subroutine getVirial(env, virial)
+
+    !> instance
+    type(TEnvironment), intent(inout) :: env
+
+    !> resulting gradients wrt atom positions
+    real(dp), intent(out) :: virial(:,:)
+
+    call recalcGeometry(env)
+    virial(:,:) = totalStress*CellVol
```
awvwgk, line 132: Since the stress tensor is saved in DFTB+ and not the virial, it might not be available for molecular system where the stress tensor is ill-definied. A check like in the `getStressTensor` routine would be necessary here:

```suggestion
    if (.not. tStress) then
      call error("Virial not available, you must initialise your calculator with&
          & this property enabled.")
    end if

    call recalcGeometry(env)
    virial(:,:) = totalStress*CellVol
```

aradi:
@gsjung0419 Are still working on this?

PR #470, opened 2020-04-30
Implement a reader for a list of species data

awvwgk:
This PR implements a reader for a list of species data, which I used frequently in my previous contributions. I also found a similar pattern in the pp-RPA and replaced it with the new reader. There is other species data in the parser, but it did not fit in the simple proof-of-principle scheme of this PR.

Additionally the species list reader will check for additional nodes present, which do not correspond to a species present in the current geometry. It will check the content of those nodes for the right type and mark them as read. The additional nodes are limited to the first 118 elements in the periodic system, which should cover most of the elements usable together with DFTB+.

The aim of this PR is to use this reader also for the `MaxAngularMomentum` or the `HubbardDerivs` nodes to make the HSD input less cumbersome and more user-friendly.

@dftbplus/maintainers This PR is open for discussion. I hope there is a more elegant way to implement this and maybe get the reader faster for long lists like this:

```cson
    Descreening = Values {
      H  =    0.30000000
      He =    1.53998804
      Li =    0.67578863
      Be =    0.97377845
      B  =    0.81140818
      C  =    0.70398030
      N  =    1.32614006
      O  =    0.85560034
      F  =    0.62220267
      Ne =    1.32703368
      Na =    0.74891818
      Mg =    0.30000000
      Al =    0.68528486
      Si =    0.90746438
      P  =    0.90580718
      S  =    0.94868883
      Cl =    0.77839591
      Ar =    0.72209903
      K  =    0.69511400
      Ca =    0.30000000
      Sc =    1.05559918
      Ti =    1.05826730
      V  =    1.04042554
      Cr =    0.80035125
      Mn =    0.89695582
      Fe =    0.98962185
      Co =    0.63768205
      Ni =    1.02347946
      Cu =    0.72529321
      Zn =    0.46166339
      Ga =    0.70264395
      Ge =    0.79017965
      As =    0.69270983
      Se =    0.94010259
      Br =    0.30000000
      Kr =    0.65877050
      Rb =    0.30000000
      Sr =    0.65424275
      Y  =    1.07262150
      Zr =    0.73384526
      Nb =    1.04760606
      Mo =    0.76043320
      Tc =    1.19945969
      Ru =    0.88344350
      Rh =    0.80070141
      Pd =    1.17400917
      Ag =    0.94331092
      Cd =    0.72523165
      In =    0.76250523
      Sn =    0.73447181
      Sb =    0.55871535
      Te =    0.66384378
      I  =    0.59834033
      Xe =    0.79993610
      Cs =    0.59824016
      Ba =    0.49615254
      Hf =    0.77923135
      Ta =    0.92170732
      W  =    0.70388295
      Re =    0.89972537
      Os =    0.97299413
      Ir =    0.85244821
      Pt =    1.09001850
      Au =    0.95252827
      Hg =    0.60967486
      Tl =    0.78738062
      Pb =    0.90805503
      Bi =    0.30000000
      Po =    0.80000000
      At =    0.80000000
      Rn =    0.80000000
    }
```


prog/dftb+/lib_dftbplus/specieslist.F90
```diff
@@ -0,0 +1,90 @@
...
+      end do
+    else
+      do iSp = 1, size(speciesNames)
+        call getChildValue(node, speciesNames(iSp), array(iSp), child=child)
+      end do
+    end if
+
+    do iSp = 1, size(pse)
+      call getChild(node, pse(iSp), value1, requested=.false.)
+      if (associated(value1)) then
```
aradi, line 78: The `getChild()` call already sets the processed flag, you do not need the `if` block afterwards.

awvwgk: But we should verify the node could be read into the respective type, otherwise the processed input might contain a string value instead of real there.

awvwgk: Removing the if leads to accepting input like
```cson
    Descreening = Values {
      H  =    0.59677857
      C  =    0.87412731
      O  =    0.94134128
      Se = "real number"
    }
```

aradi: :+1 But then at least, we could spare the repeated search in the linked lists of children by re-using `value` as `getChildValue(value1, "", ...)` within the `if` branch.

aradi: (By the way, could we then call `value1` maybe `child` instead to make the intention more explicit?) :wink:

awvwgk:
Looks much simpler now, thanks @aradi.

bhourahine:
@awvwgk you should be able to merge this yourself as its approved.

awvwgk:
@bhourahine I'm short on time right now, but I will look into updating and merging this branch soon.

Issue #469, opened 2020-04-30
dftbp_io module duplicates stdIo from dftbp_globalenv


bhourahine:
But sets it as a parameter instead.

Also stdIn and stdErr are exposed by dftbp_io, but should probably be wrapped the same way.

bhourahine:
Should eventually make this thread save and remove module variable use, but resolved for the moment.

PR #471, opened 2020-04-30
Wrapped stdErr and changed file open

bhourahine:
Also changed a few modules that should be using globalenv instead
of io definition of stdOut. Addresses most of #469.


prog/dftb+/lib_common/globalenv.F90
```diff
@@ -56,22 +62,31 @@ module dftbp_globalenv
...
 
     !> Customised global standard output
     integer, intent(in), optional :: outputUnit
 
     !> Customised global MPI communicator
     integer, intent(in), optional :: mpiComm
 
-    integer :: mpiComm0, outputUnit0
+    !> Customised global standard input
+    integer, intent(in), optional :: inputUnit
```
awvwgk, line 74: This variable is not used, why have it in the dummy arguments?

bhourahine:
> Can you remove the `dftb_io` module completely with this PR? Otherwise the changes are not really useful.

Unfortunately a few places need to know the stdOut unit before the comm world processor 0 is identified, so it can't be removed. But its use can be minimized.

awvwgk:
> Unfortunately a few places need to know the stdOut unit before the comm world processor 0 is identified, so it can't be removed. But its use can be minimized.

Initializing `stdOut` with `stdOut0` and `stdErr` with `stdErr0` in `dftbp_globalenv` would solve this issue. Alternatively this places could directly use `stdOut0` and `stdErr0`.

bhourahine:
> Initializing `stdOut` with `stdOut0` and `stdErr` with `stdErr0` in `dftbp_globalenv`

Good point.

prog/dftb+/lib_common/globalenv.F90
```diff
@@ -56,22 +62,28 @@ module dftbp_globalenv
...
 
-    integer :: mpiComm0, outputUnit0
+    !> Customised global standard error
+    integer, intent(in), optional :: errorUnit
+
+    integer :: mpiComm0, outputUnit0, errorUnit0
 
-    if (present(outputUnit)) then
-      outputUnit0 = outputUnit
+  #:for VAR, NAME in [("output", "stdOut"), ("error", "stdErr")]
```
awvwgk, line 78: This preprocessor does not look to useful here, it just makes the code obscure.

Issue #445, opened 2020-04-19
write statements in libPoisson


bhourahine:
Still some write(*,*) statements present there

PR #472, opened 2020-04-30
Replaced write(*,*) in Poisson

bhourahine:
Addresses #445


prog/dftb+/lib_poisson/bulkpot.F90
```diff
@@ -342,7 +343,7 @@ Subroutine readbulk_pot(phi_bulk, iErr)
     if (a.ne.phi_bulk(m)%iparm(14) .or. &
       b.ne.phi_bulk(m)%iparm(15) .or. &
       c.ne.phi_bulk(m)%iparm(16)) then
-      if(id0) write(*,*) 'Warning: incompatible BulkPot: will be recomputed'   
+      if(id0) write(stdOut,*) 'Warning: incompatible BulkPot: will be recomputed'
```
awvwgk, line 346: This should be a call to `warning` from `dftbp_message`.

prog/dftb+/lib_poisson/bulkpot.F90
```diff
@@ -71,15 +72,15 @@ subroutine write_super_array(SA)
...
-     write(*,*) 'val',size(SA%val)
+     write(stdOut,*) SA%a,SA%b,SA%c
+     write(stdOut,*) SA%dla,SA%dlb,SA%dlc
+     write(stdOut,*) 'size',SA%ibsize
+     write(stdOut,*) 'iparm',SA%iparm
+     write(stdOut,*) 'fparm',SA%fparm
+     write(stdOut,*) 'natm_PL',SA%natm_PL
+     write(stdOut,*) 'L_PL',SA%L_PL
+     write(stdOut,*) 'rhs',size(SA%rhs)
+     write(stdOut,*) 'val',size(SA%val)
```
awvwgk, line 83: Better use some consistent style with the rest of DFTB+ here, maybe:

```suggestion
     write(stdOut, '(a, ":", t30, g14.10)') 'size',SA%ibsize
     write(stdOut, '(a, ":", t30, g14.10)') 'iparm',SA%iparm
     write(stdOut, '(a, ":", t30, g14.10)') 'fparm',SA%fparm
     write(stdOut, '(a, ":", t30, g14.10)') 'natm_PL',SA%natm_PL
     write(stdOut, '(a, ":", t30, g14.10)') 'L_PL',SA%L_PL
     write(stdOut, '(a, ":", t30, g14.10)') 'rhs',size(SA%rhs)
     write(stdOut, '(a, ":", t30, g14.10)') 'val',size(SA%val)
```

bhourahine:
> Most of the changes are done in comments, which is not really changing anything.
> 
> Do we need the commented out blocks for debugging or are they just leftover and can be removed?

Most of them are leftover rubbish, but a few might mean something to someone (was planning on further cleanup for this routine, just want to quickly get rid of behaviour that breaks the MPI aware writing of variables)

prog/dftb+/lib_poisson/bulkpot.F90
```diff
@@ -342,7 +335,7 @@ Subroutine readbulk_pot(phi_bulk, iErr)
     if (a.ne.phi_bulk(m)%iparm(14) .or. &
       b.ne.phi_bulk(m)%iparm(15) .or. &
       c.ne.phi_bulk(m)%iparm(16)) then
-      if(id0) write(*,*) 'Warning: incompatible BulkPot: will be recomputed'   
+      call warning('incompatible BulkPot: will be recomputed')
```
awvwgk, line 338: Is this automatically protected for the MPI master?

bhourahine: Yes, the stdOut unit is directed to /dev/null on all but the master

awvwgk: Sounds hacky, but okay.

PR #474, opened 2020-05-03
Poisson sans negf

bhourahine:
Decouples use compilation and use of the Poisson solver from the WITH_TRANSPORT preprocessing to use libNEGF.
Note, the execute_command_line() call is specific to posix-like systems, so eventually should be generalised to other OS choices (and moved into its own module).
Should be orthogonal to #473 #472 #471


pecchia:
Seems fine to me

pecchia:
Unclear how to approve. Do I have write access ?

pecchia:
Before the subroutine create_directory() was taken from libnegf/ext_system 
In there I wrapped a C library call that was handling some non-posix case (windows). This was the primary reason why poisson was still dependent on transport. I guess we should write an independent library system interfacing relevant C calls like mkdir, rm, mv, cp.
Notably there are system calls in dftd3/extras.f90

bhourahine:
Yes, we need a general cross platform implementation for common functions (the extras.f90 routines seem to only cover posix). This should be a separate pull request.

bhourahine:
Approval in [comment](https://github.com/dftbplus/dftbplus/pull/474#issuecomment-623180137)

PR #476, opened 2020-05-04
Replace matrix inversion in Broyden mixer

awvwgk:
Found this superfluous matrix inversion in the Broyden mixer, when working on an iterative solver for linear equations, which should probably not use a mixer defeating the purpose of the solver by needlessly inverting matrices.

Unfortunately, this cannot use the default `gesv` wrapper for `getrf` and `getrs`, because the matrix must be transposed before solving which is not available in the `gesv` interface on the LAPACK side. Also I had to remove and replace the already present wrapper for `getrs` (not used anywhere), because it was also doing the factorisation using `getrf`, which it is not supposed to do.

We should probably think about implementing `gesv` ourselves instead of using the LAPACK interface.

### DIFF
--- a/prog/dftb+/lib_math/lapackroutines.F90
+++ b/prog/dftb+/lib_math/lapackroutines.F90
@@ -99,8 +99,20 @@ module dftbp_lapackroutines
   end interface gesvd
 
 
+  !> Solves a system of linear equations
+  !>  A * X = B  or  A**T * X = B
+  !> with a general N-by-N matrix A using the LU factorization computed by getrf.
+  interface getrs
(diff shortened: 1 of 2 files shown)


bhourahine:
There is something similar in the Anderson mixer as well.
Yes, something like gesvxx (in terms of keeping the very good regularisation for poorly conditioned matrices, not all of the extra settings it has) would be a good future addition.

awvwgk:
For this we should define a similar wrapper as for the eigensolvers, so they all look the same from the outside. Also Anderson and DIIS are fine, they use `gesv` directly.

PR #478, opened 2020-05-04
Enable builds with `arpack` on Travis-CI

awvwgk:
This PR enables `arpack` based features in the MPI build on Travis-CI.

#### Timings

No noticeable additional cost in the Travis-CI environment

ARPACK | MPI | Release | Debug
--- | --- | --- | ---
false | false | 7 min | 9 min
true | false | 7 min | 9 min
false | true | 8 min | 10 min

Release build on my fork at: https://travis-ci.com/github/awvwgk/dftbplus/builds/163701087

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ addons:
       - liblapack-dev
       - libopenmpi-dev
       - libscalapack-openmpi-dev
+      - libarpack2-dev
 
 install:
 - echo "y" | ./utils/get_opt_externals ALL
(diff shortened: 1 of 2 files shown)


bhourahine:
Probably should be done in consideration of #438.

We also need to discuss what use of ARPACK in an MPI build does, since the dependent routines will probably run n identical serial cases at best and more likely random pieces of a BLACS matrix with current code structure.

awvwgk:
At least there is a parallel version of `arpack` around: https://packages.ubuntu.com/de/bionic/libparpack2-dev

aradi:
For the moment, this has the same logic as our buildbot server (ARPACK with non-MPI builds only). We can change the logic, once the appropriate code changes has been made.

PR #485, opened 2020-05-08
Fix for subsitution of missing finite diff method

bhourahine:
Richardson method not implemented for 2nd derivative, so stop code instead silently using a different method.


bhourahine:
> For compatibility with older versions, the parser should attempt to change the differentiation from `richardson` to `finiteDiff` in the HSD input. Otherwise some processed input will become invalid.

At the moment this is unused code (coverage tests...) and will only become accessible for analytical hessian evaluation in a future PR.

PR #486, opened 2020-05-08
Fractional error in tests

bhourahine:
Required for #327 for properties with large dynamical range between test examples.


test/prog/dftb+/bin/tagdiff
```diff
@@ -76,9 +77,12 @@ class Diff(object):
 class DiffElement(Diff):
     """Calculates the difference as the maximal difference of the elements"""
 
-    def __call__(self, orig, new):
+    def __call__(self, orig, new, modif):
         try:
-            diff = max(map(lambda x,y: abs(x-y), orig.value, new.value))
+            if (modif) :
+                diff = max(map(lambda x,y: abs(x-y)/(abs(x)+np.finfo(float).eps), orig.value, new.value))
```
awvwgk, line 83: This can still become a huge value, maybe a custom regularisation constant can be provided for relative errors?

aradi:
Any opinion on using a generalized diff-concept (as in numpy `isclose()` or Julia `isapprox()`) instead of having separate cases for relative and absolute errors? If we check for `|x-x0| < tol + tol * max(|x0|)`, we should cover all cases, unless the quantity being calculated has a very small order of magnitude. Would be probably anyway a better test as the one we have now...

bhourahine:
Feature still needed, but should do this from scratch on current main

PR #489, opened 2020-05-11
Test framework and example for modes

bhourahine:
Only runs for serial code, as modes is not mpi parallel.


test/prog/modes/C24O6H8/dftb_in.hsd
```diff
@@ -0,0 +1,39 @@
...
+
+# This file uses the 5th input format
+ParserOptions = {
+  ParserVersion = 7
+}
+
+# if run in MPI parallel, allow openMP threads to be used
+Parallel = {
+  UseOmpThreads = Yes
+}
```
awvwgk, line 39: For what reason is a `dftb+` input file needed for `modes`?

test/prog/modes/README
```diff
@@ -0,0 +1,92 @@
+AUTOTEST
+========
```
awvwgk, line 2: How about converting this one to reStructuredText, even if it is only for the nice preview on GH?

test/prog/modes/C24O6H8/dftb_in.hsd
bhourahine: It isn't (parent file for hessian generation), will remove

test/prog/modes/README.rst
```diff
@@ -0,0 +1,53 @@
...
+The autotest suite contains some tests for the modes code from the DFTB+
+project, with precalculated results for automated comparison. In order to test
+this binary on your system, you should execute the following command:
+
+  ./bin/autotest2 \
+     -w <tempdir> \
+     -p <binary> \
+     -f <testfile> \
+     -s A \
+     -v
```
awvwgk, line 17:
```suggestion
.. code:: bash

  ./bin/autotest2 \
     -w <tempdir> \
     -p <binary> \
     -f <testfile> \
     -s A \
     -v
```

test/prog/modes/README.rst
```diff
@@ -0,0 +1,53 @@
...
+where
+
+  <tempdir> is an existing directory which should contain the results of the
+             calculations to be validated (this directory is not deleted
+             afterwards)
+
+  <binary> is the path to your binary
+
+  <testfile> contains the list of tests which should be calculated (you can use
+             the 'tests' file in the this directory or create your own file)
```
awvwgk, line 28:
```suggestion
tempdir
             is an existing directory which should contain the results of the
             calculations to be validated (this directory is not deleted
             afterwards)

binary
             is the path to your binary

testfile
             contains the list of tests which should be calculated (you can use
             the 'tests' file in the this directory or create your own file)
```

test/prog/modes/README
bhourahine: Done. Of course it means the DFTB+ test README also needs a PR.

bhourahine:
> Maybe change the first block to code and the second to description. My `rst` is not that good to trust in it blindly, better look at the preview before accepting my suggested changes.

Thanks for the rst suggestions, seems to work.

Issue #490, opened 2020-05-13
DEBUG_CODE never activates


bhourahine:
Missing some internal preprocessor definitions, so the code block in main.F90 that gives the dipole moment from the H-F theorem never activates.

aradi:
Hopefully fixed in #500.

PR #326, opened 2019-11-21
API for MPI. First pass 

AlexBuccheri:
Subroutines and functions added for updating species and data dependent on species ordering. Number of atoms must be kept constant. Placeholders left for support of additional functionality.

PR #491, opened 2020-05-13
Module to check coherence across global comm world

bhourahine:
Currently only deployed as a DEBUG_CODE check with API use, as the
input is externally supplied by whatever is driving the code.
The coord0 array is tested as an example, but will be relevant
for other data in the MPI enabled API.
Requires #490 to be resolved with current use, but is relevant to #420 / #326.


bhourahine:
Merge after #500

AlexBuccheri:
I've closed this PR as PR #420 is more up-to-date

PR #493, opened 2020-05-16
Xml clean up

bhourahine:
Removed the residual (non-working) XML read and write from the HSD side of the parser, leaving detailed.xml export as the only public facing XML part of the DFTB+ experience.

Probably should be merged after #489 as changes the modes parser as well (and a few other ancillary codes, skderivs, waveplot).


prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -260,14 +257,10 @@ subroutine readParserOptions(node, root, flags)
...
-    if (.not. (flags%tWriteHSD .or. flags%tWriteXML)) then
-      call detailedWarning(node, &
-          &"WriteHSDInput and WriteXMLInput both turned off. You are not&
-          & guaranteed" &
-          &// newline // &
-          &" to able to obtain the same results with a later version of the&
-          & code!")
+    if (.not. flags%tWriteHSD) then
+      call detailedWarning(node, "WriteHSDInput turned off. You are not guaranteed" // newline // &
+          &" to able to obtain the same results with a later version of the code! (the dftb_pin.hsd&
```
awvwgk, line 262: Maybe another newline before the parenthesis to avoid overlong lines in the output?

PR #500, opened 2020-05-24
Fix debug build

aradi:
It turns out, debug was not passed to the preprocessor at all, so no ASSERTS and alike were executed, probably since I implemented the cmake build framework...  :disappointed:

### DIFF
--- a/prog/dftb+/lib_dftbplus/main.F90
+++ b/prog/dftb+/lib_dftbplus/main.F90
@@ -1483,18 +1483,20 @@ contains
 
     integer :: nSpin
 
-    #:call ASSERT_CODE
-      @:ASSERT(size(over) == size(ham, dim=1))
-      @:ASSERT(size(H0) == size(ham, dim=1))
-      @:ASSERT(all(shape(rhoPrim) == shape(ham)))
-      if (allocated(iRhoPrim)) then
-        @:ASSERT(all(shape(iRhoPrim) == shape(ham)))
-        @:ASSERT(all(shape(iHam) == shape(ham)))
+    #:block ASSERT_CODE
+      @:ASSERT(size(H0) == size(over))
+      if (.not. allocated(reks)) then
+        @:ASSERT(size(ham, dim=1) == size(over))
+        @:ASSERT(all(shape(rhoPrim) == shape(ham)))
+        if (allocated(iRhoPrim)) then
+          @:ASSERT(all(shape(iRhoPrim) == shape(rhoPrim)))
+          @:ASSERT(all(shape(iHam) == shape(ham)))
+        end if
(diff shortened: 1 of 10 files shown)


prog/dftb+/lib_math/eigensolver.F90
```diff
@@ -2228,15 +2228,15 @@ end subroutine dblecmplx_zhbgv
     if (present(vl)) then
       jobvl = 'V'
       ldvl = size(vl, dim=1)
-      @:ASSERT(all(shape(vr)>=[n,n]))
+      @:ASSERT(all(shape(vl)>=[n,n]))
```
aradi, line 2231: @bhourahine Could you please confirm, that the change is OK? (Otherwise the ASSERT is triggered in the pp-RPA examples.)

prog/dftb+/lib_math/eigensolver.F90
```diff
@@ -2228,15 +2228,15 @@ end subroutine dblecmplx_zhbgv
...
     else
       jobvl = 'N'
       ldvl = 1
     end if
     if (present(vr)) then
       jobvr = 'V'
       ldvr = size(vr, dim=1)
-      @:ASSERT(all(shape(vl)>=[n,n]))
+      @:ASSERT(all(shape(vr)>=[n,n]))
```
aradi, line 2239: @bhourahine  The same here...

prog/dftb+/lib_math/eigensolver.F90
bhourahine: Can confirm that this is now correct.

prog/dftb+/lib_math/eigensolver.F90
bhourahine: likewise, now correct.

prog/dftb+/lib_dftbplus/main.F90
```diff
@@ -1483,18 +1483,20 @@ subroutine reallocateSparseArrays(sparseSize, reks, ham, over,&
 
     integer :: nSpin
 
-    #:call ASSERT_CODE
```
bhourahine, line 1486: The Hellman-Feynman block should probably also use the same construction. Likewise #491

aradi: `#:call` and `#:block` are the same constructs in Fypp (they call fypp-macro), just using different names for improved readability.

As for `ASSERT_CODE` vs. `DEBUG_CODE`: Former is included, if asserts are active (and should only contain assert related stuff), latter if the build type is debug (and could contain any kind of debug stuff). So in theory they are different. (However, as currently, debug and asserts are turned together on and off, there is no practical difference, I agree.)

PR #501, opened 2020-05-25
Add callback functionality to Python interface

vanderhe:
Add callback functionalities for population dependent external potentials to the Python ctypes interface (as a supplement to #428).

- [x] callback function for population dependant external potentials


tools/dptools/setup.py
```diff
@@ -24,7 +24,7 @@
         "Programming Language :: Python",
         "Environment :: Console",
         "Intended Audience :: Science/Research",
-        "License :: OSI Approved :: MIT License",
+        "License :: OSI Approved :: LGPL License",
```
awvwgk, line 27: Somewhat unrelated change.

PR #505, opened 2020-05-25
Serial codes

bhourahine:
Traps various serial codes against being run on more than 1 MPI processor.

Includes #504


prog/modes/modes.F90
```diff
@@ -29,6 +34,18 @@ program modes
...
   character(lc) :: lcTmp, lcTmp2
+  integer :: fdUnit
+
+#:if WITH_MPI
+  !> MPI environment, if compiled with mpifort
+  type(TMpiEnv) :: mpiEnv
+
+  ! As this is serial code, trap for run time execution on more than 1 processor with an mpi enabled
+  ! build
+  call mpifx_init_thread(requiredThreading=MPI_THREAD_FUNNELED)
```
awvwgk, line 45: Any reason to only have this for modes?

bhourahine: Probably should make this a function in a module that can be imported by programs in the main build system (there are some Fortran snippets that aren't fully integrated inside prog/transporttools though).

I think the eternal module should be a separate PR though, so will just merge.

PR #506, opened 2020-05-26
Fortran return of build and API version

bhourahine:
Examples added to API testers. Also fixed source line length in
one of them.

### DIFF
--- a/prog/dftb+/api/mm/mmapi.F90
+++ b/prog/dftb+/api/mm/mmapi.F90
@@ -108,20 +108,9 @@ contains
     !> patch level for API
     integer, intent(out) :: patch
 
-    character(*), parameter :: version = '${APIVERSION}$'
-    character(len=256) :: majorS, minorS, patchS
-    integer :: iPos, jPos
-
-    ! parse a string of the form 'x.y.z'
-    iPos = index(version, '.')
-    jPos = index(version, '.', back=.true.)
-
-    majorS = version(1:iPos-1)
-    minorS = version(iPos+1:jPos-1)
-    patchS = version(jPos+1:)
-    read(majorS, *) major
-    read(minorS, *) minor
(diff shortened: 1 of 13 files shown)


bhourahine:
@awvwgk not sure I understand. The api version _is_ returned as 3 integers from getDftbPlusApi(). Do you mean:
- the DFTB+ build string (which is git descriptive of the whole code so non numeric, and is logically distinct from the api version talking to it).
- internal macro handling? (major.minor.patch is coming from a file, matching the previous behavior).
- Something else???
Yes, the C API return is something else to discuss (define macro and/or function calls).

prog/dftb+/api/mm/mmapi.F90
```diff
@@ -84,6 +84,48 @@ module dftbp_mmapi
...
+    ! parse a string of the form 'x.y.z'
+    iPos = index(version, '.')
+    jPos = index(version, '.', back=.true.)
+
+    majorS = version(1:iPos-1)
+    minorS = version(iPos+1:jPos-1)
+    patchS = version(jPos+1:)
+    read(majorS, *) major
+    read(minorS, *) minor
+    read(patchS, *) patch
```
awvwgk, line 124: This part is quite complicated to parse the API version string and assumes a lot about the `APIVERSION` preprocessor, which offers various possibilities for breakage (`x.y`, `x.y.z-a1`). Failure in this part of the API happens to be always fatal, since it cannot be recovered due to missing guards for the `read`.

This version here is of course fine, but it can be easily broken if one is not cautious about the exact implementation details of the API function which ties in directly with CMake build system.

Maybe I'm a bit overcautious here.

bhourahine: The api version number is a flat checked in file at the moment. We should discuss at the next meeting with @aradi.

PR #507, opened 2020-05-30
 Refactor initprogram

AlexBuccheri:
This PR refactors initprogram by moving:   
* equivalency relations 
* charge initialisation and setting
* N-electrons setting

to separate subroutines within the initprogram module, where they can subsequently be used by the DFTB+ API.

### DIFF
--- a/prog/dftb+/lib_dftbplus/initprogram.F90
+++ b/prog/dftb+/lib_dftbplus/initprogram.F90
@@ -545,6 +545,15 @@ module dftbp_initprogram
   !> geometry optimiser
   type(TNumDerivs), allocatable, target :: derivDriver
 
+  !> Total charge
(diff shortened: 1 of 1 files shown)


prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -3441,6 +3174,482 @@ subroutine initProgramVariables(input, env)
...
+
+       return
+    endif
+
+
+    ! Charges not read from file
+    notChrgRead: if (.not. tReadChrg) then
+
+      if (allocated(initialCharges)) then
+         !TODO(Alex) Address use of magic number
```
AlexBuccheri, line 3442: Remove TODO. Address what this magic number is and replace with parameter.

PR #513, opened 2020-06-04
REKS Refactor

inseonglee:
This PR is WIP.

The main purpose is to change the several routines (e.g. getReksEnergyL) to original existing routines (e.g. getEnergies) so that duplicated parts can be removed for convenient coding.

Please tell me if there are more parts that you want to fix or refactor in REKS.


inseonglee:
@bhourahine @aradi Now refactoring for REKS routines is finished. Most of the fixes we talked about during the last meeting were included. Please see the commits carefully. I will wait your reviews for this PR.

Please tell me if there are more parts that you want to fix or refactor in REKS.

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -4868,11 +4893,14 @@ subroutine TReksCalc_init(reks, reksInp, electronicSolver, orb, spinW, nEl,&
       case (electronicSolverTypes%onlyTransport)
         call error("REKS is not compatible with OnlyTransport-solver")
       case(electronicSolverTypes%qr, electronicSolverTypes%divideandconquer,&
-          & electronicSolverTypes%relativelyrobust, electronicSolverTypes%elpa)
+          & electronicSolverTypes%relativelyrobust)
         call REKS_init(reks, reksInp, orb, spinW, nSpin, nEl(1), nExtChrg, extChrg,&
             & blurWidths, is3rd, isRangeSep, tForces, tPeriodic, tStress, tDipole)
-      case(electronicSolverTypes%omm, electronicSolverTypes%pexsi, electronicSolverTypes%ntpoly)
-        call error("REKS is not compatible with density matrix ELSI-solvers")
+      case(electronicSolverTypes%magma_gvd)
```
bhourahine, line 4899: The magma solver should behave like a standard serial eigensolver. Is there a particular reason for this incompatibility?

inseonglee: @bhourahine Currently, it is not tested so I do not include the case. If needed, I will test and add the magma solver case for REKS calculation if there is no problems.

inseonglee: Also, there may exist a problem with MAGMA sovler. Currently, REKS code is openmp-parallelized, especially in gradient part. So, I am not sure about the process how the OMP parallelization works under GPU case.

prog/dftb+/lib_dftbplus/main.F90
```diff
@@ -2125,8 +2121,10 @@ subroutine getSccHamiltonian(H0, over, nNeighbourSK, neighbourList, species, orb
 
     nAtom = size(orb%nOrbAtom)
 
-    ham(:,:) = 0.0_dp
-    ham(:,1) = h0
+    if (.not. isREKS) then
+      ham(:,:) = 0.0_dp
```
bhourahine, line 2125: The variable ham is intent out, so should probably be set to zero before this if test. Internally the add_shift() routine treats it as intent(inout) and just adds to its value, so it might not be initialized for REKS.

prog/dftb+/lib_dftbplus/main.F90
```diff
@@ -6722,463 +6632,255 @@ subroutine getHamiltonianLandEnergyL(env, denseDesc, sccCalc, orb, &
...
-      if (mod(iL,2) == 1) then
-        ! If iL = 3, then qm = 3u + 3d, 3u - 3d and my_qm = 3u + 3d
+      ! qm representation is converted to my_qm representation
+      if (iL <= reks%Lpaired) then
+        ! If iL = 1, then qm = 1u + 1d, 1u - 1d and my_qm = 1u + 1d
         ! calculate charge part
-        tmpBlock(:,:,:,1) = intBlock(:,:,:,1)
-        hamSp(:,1) = h0(:)
+        potential%intBlock(:,:,:,1) = reks%intBlockL(:,:,:,1,iL)
+        tmpHamSp(:,1) = h0
```
bhourahine, line 6762: The getSccHamiltonian call below is intent(out) for tmpHamSp so will over-write the stored values.

prog/dftb+/lib_dftbplus/main.F90
```diff
@@ -6722,463 +6632,255 @@ subroutine getHamiltonianLandEnergyL(env, denseDesc, sccCalc, orb, &
...
       else
-        ! If iL = 4, then qm = 4u + 4d, 4u - 4d and my_qm = 3u - 3d (= -(4u - 4d))
-        ! calculate magnetization part
-        tmpBlock(:,:,:,1) = -intBlock(:,:,:,2)
-        hamSp(:,:) = 0.0_dp
+        if (mod(iL,2) == 1) then
+          ! If iL = 3, then qm = 3u + 3d, 3u - 3d and my_qm = 3u + 3d
+          ! calculate charge part
+          potential%intBlock(:,:,:,1) = reks%intBlockL(:,:,:,1,iL)
+          tmpHamSp(:,1) = h0
```
bhourahine, line 6768: as above

prog/dftb+/lib_dftbplus/main.F90
```diff
@@ -6722,463 +6632,255 @@ subroutine getHamiltonianLandEnergyL(env, denseDesc, sccCalc, orb, &
...
+        else
+          ! If iL = 4, then qm = 4u + 4d, 4u - 4d and my_qm = 3u - 3d (= -(4u - 4d))
+          ! calculate magnetization part
+          potential%intBlock(:,:,:,1) = -reks%intBlockL(:,:,:,2,iL)
+          tmpHamSp(:,1) = 0.0_dp
```
bhourahine, line 6773: as above

prog/dftb+/lib_dftbplus/main.F90
inseonglee: Done with commit b2264d1.

prog/dftb+/lib_dftbplus/main.F90
inseonglee: In commit b2264d1, the intent is now changed, so tmpHamSp is normally updated when it enters to getSccHamiltonian routine. The input value is H0 or zero with respect to current microstate index.

inseonglee:
@bhourahine Here, I found one typo in the file. In the lib_dftb/ directory, the file name is hamitonian.F90, it may be a typo.

inseonglee:
@bhourahine Also, I think that the routine `hermitianSquareMatrix` in the rangeseparated.F90 should be moved to the sparse2dense.F90 file like symmetrizeHS routine.

In addition, test for REKS fails for periodic case, so I will see these cases.

bhourahine:
> In addition, test for REKS fails for periodic case, so I will see these cases.

Is this cases from test/prog/dftb+/reks/ ? All of these passed with several compilers, so I would be very interested in when the problem appears.

The other issues you mention need to be separate pull requests.

inseonglee:
In my cluster, the following 4 tests fail.

The following tests FAILED:
         51 - dftb+_reks/C6H8_extchrg_periodic_nosave (Failed)
         54 - dftb+_reks/HBI+H2O_periodic_stress (Failed)
         55 - dftb+_reks/HBI+H2O_periodic_stress_disp (Failed)
         57 - dftb+_reks/PSB3_2SSR_extchrg_periodic (Failed)
Errors while running CTest

My compiler is intel19, so I am still finding the reason for this difference.

bhourahine:
> My compiler is intel19, so I am still finding the reason for this difference.

Definitely interested if you find the reason, as one of the test compilers is intel 19.0.5.281

inseonglee:
The version of my intel compiler is 19.1.0.166.

I found the routine to cause the difference.
For the test of PSB3_2SSR_extchrg_periodic case, current code do not add the reciprocal part to the localDeriv0 in the line ~1785 of coulomb.F90 file. It means that only real space part is added to localDeriv0 in my current code. The previous version of the code shows correct addition for reciprocal part.

In my case, the error occurs for all periodic cases, so these phenomena may occur in the case without external charges.

bhourahine:
@inseonglee do you want to make the changes to fix this?

inseonglee:
@bhourahine I am not sure, I am just testing the reciprocal part by changing the line including derivEwaldReciprocal().

bhourahine:
@inseonglee I have just made a pull request into your branch to remove the parsing of filling and mixer for REKS and also to correctly label the resulting energy (as this changed in the main code depending on whether TS is available).

I have now tested with intel19.3 and see failures for the same cases as you get for 19.1, so will also try finding the bug.

prog/dftb+/lib_dftbplus/initprogram.F90
bhourahine: Probably would not cause problems, but this can be revisited after the 20.1 release.

inseonglee:
With changed setting of optimization, when I build using cmake, the followings appear.

......
(Adjust variables defined in this file to change compiler, linker and library settings)
-- Flags for Fortran-compiler:  -standard-semantics -O3 -O2 -ip -heap-arrays 10
-- Flags for C-compiler:   -O3 -DNDEBUG -O2 -ip
-- Build type: Release
-- Found OpenMP_C: -qopenmp (found version "5.0")
-- Found OpenMP_Fortran: -qopenmp (found version "5.0")
......

Where are these -O3 settings given from which file?
I am compiling with the new changes.

bhourahine:
The compiler options are set by cmake, but can be over-riden in the sys/intel.cmake file and also on the command line. Commit e5f7a22a4058ee5c2ef446bf652d73b9760b9c7c set the intel.cmake to -O2.
What version of cmake and command are you using when configuring?

inseonglee:
The version of cmake is 3.14.5 and I use following command.

$ env FC=ifort CC=icc cmake -DTOOLCHAIN=intel ..

I also checked the -O2 setting. All tests are passed including periodic cases that previous cases fail.

bhourahine:
The -O3 is coming from cmake itself (and is apparently considered to be a semi-bug for cmake). I've just pushed some changes to remove it from the compile settings, so when I run cmake for intel, I get:

-- Flags for Fortran-compiler:  -standard-semantics  -DNDEBUG -O2 -ip -heap-arrays 10

inseonglee:
@bhourahine Thanks for fix. I also checked the changed things. So, are there more parts to want to refactor for REKS part? In my side, only one thing remains. In reksinterface.F90 file, it contains `getMullikenPopFromRelaxedDensity_` and `getDipoleMoment_` routines used in `getReksGradProperties` routine. Actually, these routines are almost same with original routine in main.F90.

It means that `getDipoleMoment_` is perfectly same with `getDipoleMoment` in main.F90 and
`getMullikenPopFromRelaxedDensity_` is not currently refactored with `getMullikenPopulation` like commit b416c24.

If possible, I want to refactor these two routines to remove duplicated parts.

bhourahine:
Ideally we want to make the 20.1 release later this week, so would you be OK with this possibly happening after the release?

As your proposed changes are internal and do not change the user experience of REKS calculations, could you make another pull request after this branch is merged?

inseonglee:
@bhourahine OK, this part will be modified after 20.1 release, I also hope the release of 20.1 version within this week. Then, this PR is good to merge.

Issue #460, opened 2020-04-28
CMake Toolchain files incompatible with multiple GCC installations


awvwgk:
For developing with GCC one usually has several GCC installations at hand (8.4.0 as `gfortran-8` and 9.3.0 as `gfortran` for example), but with the current `sys/gnu.cmake` one can only select the default GCC.

Unfortunately there is no easy way to overwrite it in the standard way:

```
> FC=gfortran-8 CC=gcc-8 cmake -DCMAKE_TOOLCHAIN_FILE=./sys/gnu.cmake -B_build -S.
-- Build type: Release
-- The Fortran compiler identification is GNU 9.3.0
-- Check for working Fortran compiler: /usr/bin/gfortran
-- Check for working Fortran compiler: /usr/bin/gfortran - works
-- Detecting Fortran compiler ABI info
-- Detecting Fortran compiler ABI info - done
-- Checking whether /usr/bin/gfortran supports Fortran 90
-- Checking whether /usr/bin/gfortran supports Fortran 90 - yes
-- The C compiler identification is GNU 9.3.0
-- Check for working C compiler: /usr/bin/gcc
-- Check for working C compiler: /usr/bin/gcc - works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Found OpenMP_C: -fopenmp (found version "4.5") 
-- Found OpenMP_Fortran: -fopenmp (found version "4.5") 
-- Found OpenMP: TRUE (found version "4.5")  
-- Found library lapack as /usr/lib/liblapack.so
-- Found library blas as /usr/lib/libblas.so
-- Configuring done
-- Generating done
```

As the setup will ignore the default `FC` and `CC` environment variables. Of course one can overwrite the lengthy CMake variables with

```
> cmake -DCMAKE_TOOLCHAIN_FILE=./sys/gnu.cmake -DCMAKE_Fortran_COMPILER=gfortran-8 -DCMAKE_C_COMPILER=gcc-8 -B_build -S.
-- Build type: Release
-- The Fortran compiler identification is GNU 8.4.0
-- Check for working Fortran compiler: /usr/bin/gfortran-8
-- Check for working Fortran compiler: /usr/bin/gfortran-8 - works
-- Detecting Fortran compiler ABI info
-- Detecting Fortran compiler ABI info - done
-- Checking whether /usr/bin/gfortran-8 supports Fortran 90
[...]

PR #519, opened 2020-06-13
Enable testing of API on CI builds

awvwgk:
If we have PRs working on the API (#501, #420) we should continuously test them to avoid the most basic failures. This PR enables testing on Linux and OSX for the API.

*OSX build was failing, as clang is unable to link the API testers due to missing OMP symbols.*

libomp (for clang) is installed via homebrew, but the versions don't match (as expected):

```
-- Found OpenMP_C: -Xclang -fopenmp (found version "3.1") 
-- Found OpenMP_Fortran: -fopenmp (found version "4.5")
```

The easy fix would be to disable OMP on OSX builds, but let's not talk about this option. The probably better fix would be to not use clang in the first place, but use GCC. Which will require some hacks because we cannot easily do this with the present cmake toolchain files due to #460.

@dftbplus/administrators any opinion on this?

Interestingly, the GCC7 build on OSX is failing for

```
160 - dftb+_timedep/C6H6-Sym_Arnoldi (Failed)
```

while the clang build is passing.

Decision is to bump the GCC version to 8.

### DIFF
--- a/.github/workflows/fortran-build.yml
+++ b/.github/workflows/fortran-build.yml
@@ -17,10 +17,7 @@ jobs:
         pip3 install numpy
         echo "y" | ./utils/get_opt_externals ALL
     - name: Configure build files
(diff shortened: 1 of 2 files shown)


awvwgk:
Okay, let's check first if it works for the new MPI aware API tests as well.

PR #520, opened 2020-06-14
Mpi structure replicas

bhourahine:
Structure replication on top of the spin/k-point group in MPI.

Useful for paralleising things like numerical Hessian (but also could be used for statically balanced replica hamiltonians or REMD).

Recovers most of the functionality that was present at commit [b74ac92f](https://github.com/bhourahine/dftbplus/commit/b74ac92f7f63be16907993fe33be8045315371ae)

- [ X ] Group replication and comm splitting
- [ X ] SCC / forces etc. in each replica
- [  ] Clean termination if replica > 0  finishes after 0 in comm
- [  ] Check for code functionality breaks in other code modes
- [  ] Identify other modes that can use this technology cleanly


doc/dftb+/manual/dftbp.tex
```diff
@@ -980,6 +988,13 @@ \subsubsection{Masses}
 \end{verbatim}
 where \kw{Atoms} specifies the atom or atoms which each have a mass of \kw{MassPerAtom} assigned.
 
+\subsection{Replicas}
+\label{sec:dftbp.Replicas}
+
+Mutiple structures can be simultaneously simulated. Currently the input
+structure is replicated accross these. This option is only currently available
+for MPI compled code.
```
awvwgk, line 996:
```suggestion
for MPI compiled code.
```

prog/dftb+/lib_type/typegeometry.F90
```diff
@@ -33,8 +33,8 @@ module dftbp_typegeometry
     !> atomic species
     integer, allocatable :: species(:)
 
-    !> coordinates of atoms
-    real(dp), allocatable :: coords(:,:)
+    !> coordinates of atoms, potentially in multiple replicas
+    real(dp), allocatable :: coords(:,:,:)
```
awvwgk, line 37: This is probably breaking any access to the coords array in the code, needs to be very carefully checked.
See: https://github.com/dftbplus/dftbplus/pull/520/checks?check_run_id=770056167#step:6:18102

PR #525, opened 2020-06-24
Replace use of allocation status by optionals

vanderhe:
The use of the allocation status to check for presence of variables has been kept as a workaround to maintain compatibility with some older compilers; until this point. Since this is probably a thing of the past and the corresponding buggy compilers no longer have to be supported, a change to the consistent use of optional arguments is possible.

### DIFF
--- a/prog/dftb+/lib_dftb/dftbplusu.F90
+++ b/prog/dftb+/lib_dftb/dftbplusu.F90
@@ -260,10 +260,7 @@ contains
   !> Calculates the energy contribution for the DFTB+U type functionals
   !>
   !> Note: factor of 0.5 in expressions as using double the Pauli spinors
-  subroutine E_dftbU(egy, qBlock, species, orb, functional, UJ, nUJ, niUJ, iUJ, qiBlock)
-
-    !> energy contribution
-    real(dp), intent(out) :: egy(:)
+  subroutine E_dftbU(qBlock, species, orb, UJ, nUJ, niUJ, iUJ, egy, functional, qiBlock)
 
     !> charge block populations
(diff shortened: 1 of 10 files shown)


bhourahine:
Again, could you look at resolving conflicts?

vanderhe:
Conflicts resolved.

bhourahine:
It looks like several compilers have problems, intel19 in particular, with the regression tests. Do you have access to this compiler?

vanderhe:
I'm not quite getting through the new build system yet, but just ran a few tests on dpdevel. For these tests, only the Python-API related tests (pyapi_*) fail, which is actually a bit surprising because I haven't made any changes to them, at least not directly. Is it possible that this is due to some changes of the build system?

vanderhe:
I found this in one of the output files (of the pyapi_* tests), maybe you have some advice for me:
INTEL MKL ERROR: /opt/intel/2019/compilers_and_libraries_2019.5.281/linux/mkl/lib/intel64/libmkl_avx512.so: undefined\
 symbol: mkl_sparse_optimize_bsr_trsm_i8.
Intel MKL FATAL ERROR: Cannot load libmkl_avx512.so or libmkl_def.so.

vanderhe:
Conflicts resolved.

Please note that this PR breaks compatibility with Intel18. In relation to this, I have a question: I assumed that the broader introduction of optional arguments would inevitably break the compatibility. However, the Intel18-compiler is not listed under the workarounds of utils/workarounds/unallocopt*.f90. For me this means that the support of Intel18 together with using optional arguments is basically possible, but I made a mistake somewhere. Any thoughts on this?

Issue #523, opened 2020-06-20
Changing order of contacts with transport geometry fails cryptically


gpenazzi:
Maybe this is always been there or it might be upstream from libnegf, but I noticed it only now. In a transport geometry, if we do not declare contact in ascending atomic order, we get a cryptic error during the calculation `STOP ERROR (zextract_dns): bad indeces specification`.
An example of the offending input:
```
Transport {
    Device {
      AtomRange = 1 80  
    }
    Contact {
      Id = "source"
      AtomRange = 121 160 #Note that these atoms come after the ones in the drain
    }
    Contact {
      Id = "drain"
      AtomRange = 81 120
    }
 }
```
It would be nice to have at least an early check and return with a meaningful error.

dryndyk:
I checked with one of my files. 
Confirm the problem.

bhourahine:
Seems to be an ancient libNEGF line in sparsekit_drv.F90 that's been there a while. 

^da54b03 (pecchia            2010-06-07 11:39:09 +0000 2794)     IF ((i1.GT.i2).OR.(j1.GT.j2).OR.(i2.GT.A_csr%nrow).OR.(j2.GT.A_csr%ncol)) THEN
e7df07df (Gabriele Penazzi   2013-02-15 22:33:58 +0100 2795)        STOP 'ERROR (zextract_dns): bad indeces specification';
^da54b03 (pecchia            2010-06-07 11:39:09 +0000 2796)     ENDIF

The  recent libNEGF changes that pull request #528 picks up touches that line but doesn't make the message more informative.

gpenazzi:
@bhourahine yes the stop triggers there and the message is a bit uninformative because it occurs in the block matrix manipulation routines, we don't know the root cause there. There could have an earlier check and better message in libnegf where we can be sure that it is a contact-device partitioning problem, but it would be nice to deal with this case normally. I'll try to give it a look if no one fixes this before.

pecchia:
I think this is a bug. I will investigate. The error message is indeed uninformative.
But it is there to trigger further debugging. In fact the code there only checks that the block indices make any sense. Probably the bug comes much before.

I can add info. 
In general we should remove the stops and define a sort of exception handling.
We can discuss how to do this without adding a return error to any function (that I really do not like).

PR #528, opened 2020-06-29
Libnegf fixes

pecchia:
Performance speed up of matrix inversion with better threaded mkl routine xgetrs (up to x7 on large matrices).
CMake fixes in the mpifx dependence.


pecchia:
Now this branch address the contact order in #523 with a more robust implementation. So now any order works.
We should decide what to do with the reordering implemented by Ben. We can still  decide that in dftb+ the contacts are internally reordered with increasing atom indices.

bhourahine:
I would suggest we drop PR #523 as your PR is a better solution.

PR #530, opened 2020-07-01
PROGRESS-BML Integration

cnegre:
This (2018) version of the code contains the integration with PROGRESS and BML libraries.


prog/dftb+/lib_extlibs/bml.F90
```diff
@@ -0,0 +1,24 @@
...
+!  Copyright (C) 2018  DFTB+ developers group                                                      !
+!                                                                                                  !
+!  See the LICENSE file for terms of usage and distribution.                                       !
+!--------------------------------------------------------------------------------------------------!
+
+#:include 'common.fypp'
+
+!> Provides interface to the BML library.
+module dftbp_bml
+  use bml
```
bhourahine, line 12: This only seems to be pulling in BML interfaces, is there an equivalent for progress itself?

cnegre:
Those should be under 
prog/dftb+/lib_extlibs/progress.F90

ckoe-bccms:
Just one observation: 

With bml 2.0.0 and qmd-progress 1.1.0 this does not link if -standard-semantics is used. The -standard-semantics switch is the default setting in the make.x86_64-linux-intel of cnegre:master. Unfortunately I cannot compile qmd-progress with a gcc newer than gcc8, so I did not test how building with gcc behaves. 

After fiddling with the SK-Files, symlinks are missing under test/prog/dftb+, the test/prog/dftb+/progress/h2o_dense can be executed manually. The mermin free energy calculated with the compiled binary appears to be correct compared to autotest.tag. So I assume an earlier bml version, i.e. 1.3.1 or 1.2.3, is not the required bml version. Note that my qmd-progress build.log still says 
```
-- Checking for module 'bml'
--   Found bml, version 1.3.1
```
although the archive the source was extracted from is the 2.0.0 version of bml.

PR #533, opened 2020-07-06
Fix for contact ordering

bhourahine:
Addresses #523 from the DFTB+ side.

Note, in the case of non-compliant contact ordering, a reordering
happens with the new order printed in the stdout.
Note, various places in libNEGF print things like current between
contacts 1 and 2 (etc.) and in these cases the new ordering is
the relevant one, not the user supplied input for DFTB+.

- [x] Contact order from input processed
- [ ] Labels for printing in DFTB+ for contact properties which are currently just numeric
- [ ] libNEGF contact printing labels that are just numeric


pecchia:
I think we where fixing things in opposite ways. 
At the moment the initialization in libnegf.F90 init_str() assumes the contacts are in increasing order. 
However this is an internal problem (or my mistake) that can be fixed by passing an additional parameter.
Each contact can in general include a surface part (originally included to inject from bound states, e.g. in a RTD). So I have Contact_start(:), Surface_end(:) and contact_end(:) but I initialize these only using surfend(:) and contend(:) input. Hence I set contact_start(i+1)=contend(i)+1, which obviously assumes the implicit ordering.
 
I would like to add the input parameter surfstart(:)  to make things more robust.

bhourahine:
Changes in PR #528  make this unneeded.

Issue #529, opened 2020-06-29
Geometry output


Mverena:
Hello,

I observed two problems with the gen output, when performing calculations with > 10 atomic species and using dummy names for those species that contain > 2 digits as name.

1) When using >2 digits as a specifier for the atomic species, like "H11", next to a different specifier like O they get combined to OH11 instead of O H11. Note: I am using these dummy names due to a modification of the SK files from an original Hydrogen file.

2) When having > 100 atoms in the system and > 10 atomic species the atomic indices' and atomic species' numeric specifier get combined to e.g. 11011 instead of 110 11.

These issues require me to always manually modify the gen output in order to use it.

Best
Verena

bhourahine:
Can you test the code in #538 to see if this resolves your issue?

Mverena:
Hi Ben,

thanks a lot for the fix. The code resolves my issue.
The gen output looks fine now.

Best
Verena

PR #538, opened 2020-07-12
Fix gen format fails with >10 species

bhourahine:
Solution for #529

### DIFF
--- a/prog/dftb+/lib_io/formatout.F90
+++ b/prog/dftb+/lib_io/formatout.F90
@@ -152,7 +152,7 @@ contains
     real(dp) :: invLatVec(3,3)
 
     ! Format first line of a gen file
-    character(len=*), parameter :: formatHead1 = '(1X,I0," ",A2)'
+    character(len=*), parameter :: formatHead1 = '(1X,I0,1X,A2)'
 
     ! Format a vector
     character(len=*), parameter :: formatVec = '(3E20.10)'
@@ -207,7 +207,7 @@ contains
 
     write(formatCoordinates, '("(I5,2X,I",I0,",3E20.10)")') floor(log10(real(nSpecies)))+1
     if (tFractional) then
-      invLatVec(:,:) = latVec(:,:)
+      invLatVec(:,:) = latVec
       call matinv(invLatVec)
       do ii = 1, nAtom
         write(fd, formatCoordinates) ii, species(ii), matmul(invLatVec,coord(:, ii) + origin)
--- a/prog/dftb+/lib_io/formatout.F90
+++ b/prog/dftb+/lib_io/formatout.F90
@@ -146,16 +146,16 @@ contains
     logical, intent(in), optional :: tFracCoord
 
     integer :: nAtom, nSpecies
-    character(mc) :: formatSpecies, formatCoordinates
+    character(mc) :: formatCoordinates
     integer :: ii
     logical :: tFractional, tHelical, tPeriodic
     real(dp) :: invLatVec(3,3)
 
-101 format(1X,I0," ",A2)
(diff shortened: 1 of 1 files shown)


prog/dftb+/lib_io/formatout.F90
```diff
@@ -189,34 +188,38 @@ subroutine writeGenFormat_fid(fd, coord, species, speciesName, latVec, origin, t
...
 
-    write(formatSpecies, 101) nSpecies
-    write(fd, formatSpecies) (trim(speciesName(ii)), ii = 1, nSpecies)
+    do ii = 1, nSpecies
+      write(formatSpecies, "('(A',I0,')')")len(trim(speciesName(ii)))+1
+      write(fd, formatSpecies, advance='NO')trim(speciesName(ii))
+    end do
+    write(fd,*)
 
+    write(formatCoordinates, 102) floor(log10(real(nSpecies)))+1
```
awvwgk, line 209: Please, don't mix `format` statements and format strings, this makes IO routines even harder to read.

prog/dftb+/lib_io/formatout.F90
```diff
@@ -146,14 +146,13 @@ subroutine writeGenFormat_fid(fd, coord, species, speciesName, latVec, origin, t
...
-    character(6) :: formatSpecies
+    character(mc) :: formatSpecies, formatCoordinates
     integer :: ii
     logical :: tFractional, tHelical, tPeriodic
     real(dp) :: invLatVec(3,3)
 
-100 format(I5," ",A2)
-101 format("(",I2.2,"A3)")
-102 format(I5,I2,3E20.10)
+101 format(1X,I0," ",A2)
```
awvwgk, line 154: Is there a specific need for the leading space? Also I would prefer to use either `1X` or `" "` for spaces to keep the format statements consistent.

prog/dftb+/lib_io/formatout.F90
```diff
@@ -189,34 +188,38 @@ subroutine writeGenFormat_fid(fd, coord, species, speciesName, latVec, origin, t
...
     else
-      write(fd, 100) nAtom, "C"
+      write(fd, 101) nAtom, "C"
     end if
 
-    write(formatSpecies, 101) nSpecies
-    write(fd, formatSpecies) (trim(speciesName(ii)), ii = 1, nSpecies)
+    do ii = 1, nSpecies
+      write(formatSpecies, "('(A',I0,')')")len(trim(speciesName(ii)))+1
+      write(fd, formatSpecies, advance='NO')trim(speciesName(ii))
```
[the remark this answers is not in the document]
bhourahine: A near miss, but '(1x,a)' will do it correctly.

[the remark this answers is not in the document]
awvwgk: Oh yeah, now I see, `len(trim(speciesName(ii)))+1` to sneak the space in was something I missed.

prog/dftb+/lib_io/formatout.F90
bhourahine: Point taken (was just sticking to the messy style of the existing routine).

prog/dftb+/lib_io/formatout.F90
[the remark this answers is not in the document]
bhourahine: (depending on view about leading or trailing spaces).

prog/dftb+/lib_io/formatout.F90
bhourahine: Not really, resulting file just looks neater to me. The above '(a)' discussion replicates this behaviour anyway.

Issue #535, opened 2020-07-07
Suspect formula in transition dipole moment


francesco-bosia:
There is a suspect formula [here](https://github.com/dftbplus/dftbplus/blob/a4e65b65da71b61a866a42ec79649972ba85ec86/prog/dftb%2B/lib_timedep/linrespcommon.F90#L814).

I think it should be 

```
    rtmp = eval(ii)**(-1/4) ,
```
as also suggested by by the [literature](https://doi.org/10.1016/j.theochem.2009.04.034) and by the comment
```
! 1._dp / sqrt(sqrt(eval(ii))) .
```
In this subroutine, eval are the eigenvalues of the coupling matrix, i.e. the square of the vertical transition energies. 
I think one could use the coefficients directly, as done in [here](https://github.com/dftbplus/dftbplus/blob/a4e65b65da71b61a866a42ec79649972ba85ec86/prog/dftb%2B/lib_timedep/linrespcommon.F90#L782), instead of transforming the eigenvectors in-place. 
In fact, one gets wild transition dipoles, on the order of 1e4 - 1e6 Debye for a bright state. 

Or am I missing something?

aradi:
@francesco-bosia Thanks a lot! indeed, this is an implementation bug! We will fix it asap!

PR #539, opened 2020-07-12
Fix for transition dipole error and testing

bhourahine:
Fix for #535


bhourahine:
The transition dipole vector isn't portable as its directly proportional to the eigenvector, so caries an arbitrary phase

prog/dftb+/lib_timedep/linrespgrad.F90
```diff
@@ -2034,6 +2035,10 @@ subroutine writeExcitations(sym, osz, nexc, nmatup, getij, win, eval, evec, wij,
       if (.not.tDegenerate) then
         call taggedWriter%write(fdTagged, tagLabels%excEgy, eval)
         call taggedWriter%write(fdTagged, tagLabels%excOsc, osz)
+        if (fdTradip > 0) then
```
aradi, line 2038: That's a risky check. If units are auto-generated (with ``newunit=``), they will be negative according to the standard.

aradi:
@thomas-niehaus Do you have any ideas how to fix the non-portability of the transition dipole vectors?

prog/dftb+/lib_timedep/linrespgrad.F90
bhourahine: Yes, the whole file and linresp.F90 both need to be refactored to remove generated file id as these are used for _all_ output files.

aradi: OK, I see, it's all over in the file :disappointed:

PR #540, opened 2020-07-12
Parser20.1

bhourahine:
Parser version numbers now able to match code release label.

Mutually exclusive with #546, therefore closes #546


bhourahine:
Probably should also use this opportunity to move the ElectronDynamics block out of Analysis{}. Its not a release feature yet, but v8 backward compatibility would be possible to cover between 19.1 and the 20.1 release.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -237,25 +233,40 @@ subroutine readParserOptions(node, root, flags)
...
-    integer :: inputVersion
     type(fnode), pointer :: child
+    real(dp) :: rTmp, rVersion
+
+    ! Parser input version in form x.y with year and release number, matching code (after version 8)
+    call getChild(node, "ParserVersion", child, requested=.false.)
+    if (associated(child)) then
+      call getChildValue(child, "", rTmp, rVersion)
+      flags%version%major = nint(rTmp)
+      flags%version%minor = nint(10.0_dp*(rTmp - flags%version%major))
```
awvwgk, line 246: Really? A real value for the parser version? I would advocate for a version *string* here.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -237,25 +233,40 @@ subroutine readParserOptions(node, root, flags)
...
+      if (abs(rTmp - flags%version%major + 0.1_dp*flags%version%minor) > epsilon(1.0)) then
```
awvwgk, line 247: This arithmetic error allows only integer values to pass the test (`20.0` will pass, `20.1` will always fail)
```suggestion
      if (abs(rTmp - flags%version%major - 0.1_dp*flags%version%minor) > epsilon(1.0)) then
```
I would still prefer to replace the entire thing.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -237,25 +233,40 @@ subroutine readParserOptions(node, root, flags)
...
+        call detailedError(child, "Invalid parser version string")
+      end if
+    else
+      flags%version = currentParser
```
awvwgk, line 251: Should set the `ParserVersion` node to make it appear in the processed input

awvwgk:
A simple alternative to this PR can be found in: https://github.com/awvwgk/dftbplus/tree/parser-version

PR #541, opened 2020-07-13
Remove unused variables (trivial ones)

vanderhe:
A clear number of warnings from the compiler is essential for how warnings are generally perceived by the developer. At the moment, the overwhelming number of warnings means that no one would notice another warning or even care about it. The cause of the warnings is predominantly (apparently) unused variables.

In this sense, this PR removes some unused variables (trivial ones).


bhourahine:
The main code is nearly stable for the next release. Could you resolve the conflicting files and I'll then review

vanderhe:
Conflicts resolved. The last commit message was wrong and I wanted to correct it (hope nothing messed up ...).

PR #544, opened 2020-07-14
Cmake install of timeprop analysis scripts

bhourahine:
Adds the two analysis scripts for time propagation to the install in cmake.


aradi:
I've restructured it a bit to have install instructions in the folder where the script to be installed are....

PR #546, opened 2020-07-16
Allow usage of program version to specify parser version

awvwgk:
- `InputVersion` in root disallows `ParserVersion` in `ParserOptions`
- To ease usage of the versioning in the code `InputVersion` is mapped to a sequential `ParserVersion`
- `InputVersion` "20.1" translate to `ParserVersion` 8
- in case an older program input version is requested, the node will be deleted and replaced with the correct `ParserOptions/ParserVersion` node
- modifies an existing input test to test conversion from `InputVersion = "19.1"` to `ParserOptions/ParserVersion = 7` and sub-sequentially to `ParserOptions/ParserVersion = 8`

```cson
InputVersion = "20.1"
```

equivalent to

```cson
ParserOptions {
   ParserVersion = 8
}
```

Mutually exclusive with #540, therefore closes #540


test/prog/dftb+/input/caffeine_xyz/dftb_in.hsd
```diff
@@ -27,9 +27,7 @@ Analysis {
   CalculateForces = Yes
 }
 
-ParserOptions {
-  ParserVersion = 8
-}
+InputVersion = "19.1"
```
bhourahine, line 30: 8 == 20.1 ?

awvwgk: You're right, but than this would only test the reader and not the conversion, rolling the version back to "19.1" allows to test both (including the unsetting of the `InputVersion` node and the creation of the new `ParserVersion` node).

bhourahine: Will that input work with the 19.1 code though (isn't the xyz format read post-release)? Probably better to have an example using features from that release.

awvwgk: Added a new test for this case using `genFormat`, but this input will probably not work with version "18.2" because it is using the `InputVersion` node :wink:.

PR #548, opened 2020-07-16
Moved electron dynamics to its own block

bhourahine:
Update affected tests, moving most of them to 20.1 input format


test/prog/dftb+/timeprop/CH3_dftbu_singlet/dftb_in.hsd
```diff
@@ -70,10 +70,8 @@ ParserOptions = {
   ParserVersion = 3
```
awvwgk, line 70: Should this still be version 3?

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -4808,19 +4808,16 @@ end subroutine readCustomisedHubbards
 
 
   !> Reads the electron dynamics block
-  subroutine readElecDynamics(node, input, geo, masses)
+  subroutine readElecDynamics(node, ctrl, geom)
```
awvwgk, line 4811: Passing `TElecDynamicsInp` should be preferred over the complete `TControl` as it makes the parser more modular.

test/prog/dftb+/timeprop/CH3_dftbu_singlet/dftb_in.hsd
bhourahine: Unfortunately not, as the finite difference methods changed after that parser version (this could either be requested with an extra keyword, or the test recomputed with the more accurate new default).

prog/dftb+/lib_dftbplus/parser.F90
bhourahine: OK, will revert (requires two parts of the structure to be passed  separately, as mass is in there as well).

test/prog/dftb+/timeprop/CH3_dftbu_singlet/dftb_in.hsd
bhourahine: Replaced it with a recomputed 20.1 input.

PR #554, opened 2020-07-19
Reks refactor update

bhourahine:
Updated #513 to state of main repository plus some bug fixes.

- [x] openMP memory issue with external charges
- [ ] Parser not requiring filling and mixer blocks for REKS
- [ ] Use current infrastructure for which type of energy is being written


inseonglee:
@bhourahine In current code, filling and mixer are not used for REKS case, but parser read these two blocks in Hamiltonian. So, these parts should be modified as your comment in #513. Also, the mermin and E0 energies are not necessary in REKS.

inseonglee:
I want to test this current code, but it fails building for cmake process, please see the issue #555.

bhourahine:
The changes in commit [4a2bf2f4](https://github.com/dftbplus/dftbplus/commit/4a2bf2f46b6f79ef9fc653c119814c4714265328) and the update the INSTALL.rst in [3097e34a](https://github.com/dftbplus/dftbplus/commit/3097e34a33b692c91ac34f52adc91877fb3e54c4) switched cmake to the more standard way of accepting extra arguments, so to build for gnu compilers:

env FC=gfortran CC=gcc cmake ..

which reads most of its settings from sys/gnu.cmake which is selected based on the compiler choice.

bhourahine:
I've now rebased the update branch against your pull request, so will add the changes to this branch and withdraw the other pull request.

Then I will make a pull request into this branch to make the parser and energy printing changes.

PR #556, opened 2020-07-19
Set release optimisation to O2 for intel/gnu

bhourahine:
Changes default behaviour to match previous build system.


bhourahine:
Note, this is already included in #513

PR #561, opened 2020-07-19
Intel O3 fix

bhourahine:
Removes the default -O3 from intel release builds, as should be -O2 (as is set in intel.cmake, which was giving -O3 -O2 in compile lines).

This change is already included within #513 but is here for separate review.


inseonglee:
Here, the commit 5c763ef is not included in #513 , only fb98c44 is included in that PR and merged.

bhourahine:
Commit 5c763ef is now merged to the main branch.

PR #571, opened 2020-07-21
Mpi config

aradi:
* Minor changes in the build system to allow for arbitrary linker options for external libraries
* Add GNU+MKL example to INSTALL.rst


CMakeLists.txt
```diff
@@ -30,20 +30,14 @@ if(WITH_API)
...
-else()
-  option(WITH_MPI "Whether DFTB+ should support MPI-parallelism" FALSE)
-endif()
 if(WITH_MPI AND NOT MPI_FORTRAN_FOUND)
   message(FATAL_ERROR "Compiler ${CMAKE_Fortran_COMPILER} is not MPI capable but is specified for "
     "a WITH_MPI=TRUE build")
 elseif(NOT WITH_MPI AND MPI_FORTRAN_FOUND)
   message(WARNING "MPI enabled compiler ${CMAKE_Fortran_COMPILER} is specified for a build where "
-    "WITH_MPI=FALSE")
+    "WITH_MPI=FALSE. Your build will NOT be MPI-parallelised.")
```
bhourahine, line 38: Might be worth giving a hint as to how to build with MPI. Something like:
MPI capable compiler found but your build will NOT be MPI-parallelised. To enable this set WITH_MPI to TRUE

aradi: :+1:  :heavy_check_mark:

INSTALL.rst
```diff
@@ -89,7 +89,11 @@ PLUMED 2.5 libraries.
 
 Notes:
 
-[1] Only serial version tested.
+[1] The timedep/C60_OscWindow test fails with recent versions (>= 0.3.8) of the
+OpenBlas library. If possible, use older versions or link against the MKL
```
bhourahine, line 93:
```suggestion
OpenBlas library. If possible, use an older version or link against either MKL or another blas
```

INSTALL.rst
```diff
@@ -181,30 +185,35 @@ In order to build DFTB+ carry out the following steps:
...
   You may adjust any variables defined in `config.make` or in the toolchain file
   by either modifying the files directly or by overriding the definitions via
-  the ``-D`` command line option. For example, in order to change the name of
-  the LAPACK library, you can override the ``LAPACK_LIBRARIES`` variable with::
+  the ``-D`` command line option. For example, in order to use the MKL-library
+  with the GNU-compiler, you would have to override the ``LAPACK_LIBRARIES``
+  variable::
+
+    FC=gfortran CC=gcc cmake -DLAPACK_LIBRARIES="mkl_gf_lp64;mkl_gnu_thread;mkl_core"  ..
+
```
bhourahine, line 193:
```suggestion

You can can also specify the location of other libraries, for example an MPI enabled build using ScaLAPACK::
  
  FC=mpiifort CC=icc cmake -DWITH_MPI=TRUE -DLAPACK_LIBRARY_DIRS=$MKLROOT/lib/intel64 ..

```

PR #574, opened 2020-07-24
Changes for ELSI 2.6.2 release

bhourahine:


prog/dftb+/lib_elecsolvers/elsisolver.F90
```diff
@@ -522,8 +520,8 @@ subroutine supportedVersionNumber(this, dateStamp)
       isSupported = .false.
     elseif (this%major == 2 .and. this%minor < 6) then
       isSupported = .false.
-    elseif (this%major == 2 .and. this%minor == 6 .and. all(this%patch /= [0,1])) then
-      ! library must be 2.6.{0,1}
+    elseif (this%major == 2 .and. this%minor == 6 .and. all(this%patch /= [0,1,2])) then
```
awvwgk, line 523: If they go on patching their release this list will be growing quickly, maybe make a relative comparison like:
```suggestion
    elseif (this%major == 2 .and. this%minor == 6 .and. this%patch >= 0 .and. this%patch <= 2) then
```

PR #579, opened 2020-07-25
Rename files and types for clarity over energy

bhourahine:
Also directly expose energy_init instead of requiring signature
detection. Additionally fix a dftb_ module name for dftbp_.

### DIFF
--- a/prog/dftb+/lib_dftb/energytypes.F90
+++ b/prog/dftb+/lib_dftb/energytypes.F90
@@ -8,13 +8,13 @@
 #:include 'common.fypp'
 
 !> Module to wrap around the different energy components in the DFTB total energy expression
-module dftbp_energies
+module dftbp_energyTypes
   use dftbp_assert
   use dftbp_accuracy
   implicit none
   private
 
-  public :: TEnergies, init
+  public :: TEnergies, energies_init
 
 
   !> Data type to store components of the energy as named variables instead of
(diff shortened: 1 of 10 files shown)


prog/dftb+/lib_dftb/energytypes.F90
```diff
@@ -8,13 +8,13 @@
...
 !> Module to wrap around the different energy components in the DFTB total energy expression
-module dftbp_energies
+module dftbp_energyTypes
   use dftbp_assert
   use dftbp_accuracy
   implicit none
   private
 
-  public :: TEnergies, init
+  public :: TEnergies, energies_init
```
awvwgk, line 17: Didn't we aim for a `TEnergies_init` naming convention with those initializers?

prog/dftb+/lib_dftb/energytypes.F90
```diff
@@ -8,13 +8,13 @@
 #:include 'common.fypp'
 
 !> Module to wrap around the different energy components in the DFTB total energy expression
-module dftbp_energies
+module dftbp_energyTypes
```
awvwgk, line 11: The naming convention for modules prefers snakecase here, please avoid capital letters in module names.

prog/dftb+/lib_timedep/dynamics.F90
```diff
@@ -42,8 +42,8 @@ module dftbp_timeprop
   use dftbp_periodic
   use dftbp_velocityverlet
   use dftbp_nonscc
-  use dftbp_energies, only: TEnergies, init
-  use dftb_evaluateenergies
+  use dftbp_energytypes
+  use dftbp_getenergies
```
awvwgk, line 46: Could you make the imports here explicit?

bhourahine: Imports all of the public routines, but sure.

PR #580, opened 2020-07-26
Fix version number to 20.1

bhourahine:
Might want to consider a backport into the 20.1 release for this one as CMake exposes wrong version id.


bhourahine:
Succeeded by #588

Issue #582, opened 2020-07-27
Electric field unit modifier inconsistent with force units


bhourahine:
/a and /aa for per Ångström

bhourahine:
Resolved by #583

PR #583, opened 2020-07-27
Added consistent AA unit conversions

bhourahine:
All units using Angstroms now have all three variants. Addresses #582


bhourahine:
Just fixed the relevant table in the manual.

PR #587, opened 2020-07-30
Allow overwrite of atomic numbers in dispersion input

### LINKED ISSUE #586: Pseudo-atom conflict with dispersion correction
Dear developers,

Currently we are performing test including pseudo-atoms in the dftb+ calculation still using the 19.1 version. We tailored specific skf-files to deal with dummy carbon atoms (labeled CX) and our initial tests proved quite positive. The overall aim of this work is to provide a QM/MM setup for carbon materials including dummy carbon atoms close to the QM/MM boundary. 

However, as soon as we invoke dispersion correction the element type CX is of course not recognised and although dftbplus terminates correctly, the associated energy and force output is broken - only some ******** are included in the output.

Is the use of dummy/pseudo atoms possible in DFTBPLUS (maybe just using a different letter such as X or Q) or, alternatively, is it possible to provide extra data in the dispersion input section for our new "element" CX?

Of course we could always turn of dispersion correction, but we believe it is inherent contribution in modern dftb applications.

Best wishes and many thanks,
Thomas S. Hofer

awvwgk:
Allows overwriting of atomic number information of for dispersion corrections relying on atomic numbers rather than species names (like DFT-D3 and DFT-D4).

- [x] Add `AtomicNumbers` method to `DFTD3`
- [x] Add `AtomicNumbers` method to `DFTD4` and child methods
- [x] Document the atomic number overwrite
- [x] Include regression tests

#### New Input

```cson
Dispersion = DFTD3 {
  AtomicNumbers { CX = 6 }
}
```

Closes #586


doc/dftb+/manual/dftbp.tex
```diff
@@ -2756,6 +2757,9 @@ \subsubsection{DftD3 optional settings}
 \item[\is{Threebody}] Whether three-body contributions should be included in the
   dispersion interactions.
 
+\item[\is{AtomicNumbers}]
+  Allows overwrite of atomic numbers associated with a species name.
```
bhourahine, line 2761: Probably needs a syntax use example as a verbatim block

awvwgk: Added an example, but I think it is more an expert feature one learns from the provided tests in the testsuite.

PR #588, opened 2020-07-31
Fixes stray 19.x version numbers

bhourahine:
Cherry-Pick of #580 on top of 20.1 tag

Back-port comment from #580 [review](https://github.com/dftbplus/dftbplus/pull/580#pullrequestreview-458567366) still relevant.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.5)
 cmake_policy(VERSION 3.5)
 
-set(DFTBPLUS_VERSION "19.2")
+set(DFTBPLUS_VERSION "20.1")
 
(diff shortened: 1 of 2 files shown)


awvwgk:
Is any of those 19.1 version strings actually preserved in the generated binary distribution (either to GH release or the cf package)? It seems like the binary is emitting the correct version and it should be available in all API based access as well.

So the only problem would be direct inclusion of DFTB+ in a CMake project, wouldn't it? Or are the `cmake` package files also affected?

bhourahine:
The generated ${CMAKE_INSTALL_PREFIX}/lib/pkgconfig/dftbplus.pc is effected, so anything picking up from that, but otherwise yes the direct CMake inclusion.

On the topic of dftbplus.pc, probably we should switch to [configure_package_config_file](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html) instead of configure_file.

PR #589, opened 2020-07-31
Scc refactor

bhourahine:
Refactor of the scc and coulomb modules to decouple dependencies

- [x] New naming convention for init routines
- [x] Charge constraint module adapted
- [ ] Coulomb routines as procedures
- [ ] REKS to new structure
  - [x] REKS using Coulomb container
  - [ ] Update Ewald parameters on lattice changes
- [ ] Remove SCC use of Coulomb routines
- [ ] Poisson solver integrated with Coulomb container


aradi:
Should we close this PR? Changes on master make it probably obsolate?

Issue #592, opened 2020-08-01
InputVersion in dftb_in converted to ParserVersion in dftb_pin


bhourahine:
Setting the input from the code release as InputVersion converts into the (internal) parser number in the processed output. Probably clearer to leave as code version and possibly to convert to release number if possible.

awvwgk:
I don't think this is desirable, on updating from an older `InputVersion` we will create input for the latest `ParserVersion`, we might have multiple `InputVersion` referring to the same `ParserVersion` or might be in between releases were the `ParserVersion` is already bumped but the `InputVersion` not.

Replacing the `InputVersion` by a `ParserVersion` is in my opinion the safest and least ambiguous choice.

bhourahine:
A look-up table back from ParserVersion to InputVersion then needs to be available somewhere, unless the manual always carries both instead of just the release label (== InputVersion, and the original motivation for what became InputVersion).

awvwgk:
I created the mapping from `InputVersion` to `ParserVersion` by looking up the respective parameters in the source at the release tags, as I was not able to find a place where this was documented. Having a list in the manual would be an improvement.

PR #593, opened 2020-08-02
Restart ehrenfest skip

bhourahine:
Skips over the ground state calculation in the case of restarted electron dynamics where there are no required results from an initial SCC calculation.


bhourahine:
Also correctly enables the ascii C6H6 restart test and recomputes the binary restart case to omit non-printed tags from the ground state.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -7371,6 +7389,8 @@ subroutine getParserVersion(node, versionString, parserVersion)
     integer, intent(out) :: parserVersion
 
     select case(trim(versionString))
+    case("20.2")
+      parserVersion = 9
```
awvwgk, line 7393: This should be done with the release not with this PR, I think. Goes back to the argument in https://github.com/dftbplus/dftbplus/issues/592#issuecomment-667562199, DFTB+ will still print 20.1 as version, I think the input file should not bypass this.

awvwgk (changes requested):
Seems fine, except for the bump of the InputVersion. This change should be done exclusively in the release process.

prog/dftb+/lib_dftbplus/parser.F90
bhourahine: @awvwgk hopefully resolved

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -7389,10 +7389,13 @@ subroutine getParserVersion(node, versionString, parserVersion)
...
-    case("20.2")
-      parserVersion = 9
+    ! upcomming release
+    !case("20.2")
+      !parserVersion = 9
     case("20.1")
-      parserVersion = 8
+      ! parserVersion = 8
+      ! temporarily until 20.2 release:
+      parserVersion = 9
```
awvwgk, line 7398: I would leave this as version 8, since this is what it is, otherwise input labelled with InputVersion 20.1 in the test suite will not get converted correctly.

awvwgk (approved):
Ready to go from my side now.

PR #600, opened 2020-08-05
Try to infer libdftbplus location automatically in ctypes API

### LINKED ISSUE #599: ctypes API should find libdftbplus itself
When building DFTB+ and the ctypes API locally it is easy to always work with the absolute path, but on distributions like the conda-forge package the absolute path has to found by inspecting the environment in the conda prefix.

The structure in the prefix is usually the same with

```
$PREFIX/lib/libdftbplus.*
$PREFIX/lib/python3.X/site-packages/dftbplus.py
```

Therefore, it might be safe to assume the relative location of `libdftbplus.*` to `dftbplus.py` is always `../../libdftbplus.*`, so something like

```python
libdftbplus = os.path.join(os.path.dirname(__file__), "..", "..", "libdftbplus")
```

might turn out to be a more reliable default than `"./libdftbplus"`.

awvwgk:
closes #599

### DIFF
--- a/tools/pythonapi/src/dftbplus.py
+++ b/tools/pythonapi/src/dftbplus.py
@@ -31,7 +31,7 @@ class DftbPlus:
     '''
 
 
-    def __init__(self, libpath='./libdftbplus', hsdpath='./dftb_in.hsd',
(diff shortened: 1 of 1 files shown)


bhourahine (approved):
Looks good

PR #601, opened 2020-08-06
MPI parallelisation of dispersion corrections

awvwgk:
- [x] UFF dispersion correction
  - [x] enforce style guide
  - [x] include OMP parallelisation
- [x] Slater–Kirkwood type dispersion
  - [x] enforce style guide
  - [x] include OMP parallelisation
  - [x] implement stress for Slater–Kirkwood type dispersion (manual update!)
- [x] Reciprocal lattice summation
  - [x] include OMP parallelisation
- [x] DFT-D4
- [x] Include in changelog


awvwgk:
I would tend to call this variant D3, but we can of course take parts of my #602 implementation for the Slater–Kirkwood case.

PR #602, opened 2020-08-06
Reimplement DFT-D3 in a simple form

awvwgk:
A version of my https://github.com/awvwgk/simple-dftd3 tailored for DFTB+ which does not carry the GPL-1+ license like the original implementation. Allows to work as drop-in replacement for the dftd3-lib.


bhourahine (approved):
Looks ready to merge

PR #604, opened 2020-08-06
Change indentation style of Python-API and related files

vanderhe:
Switch Python-API and related files to a less laborious indentation style, that is more convenient to maintain. In addition, continuation lines - indented by 4 white spaces - are more inline with the Fortran related style guide, while being also suitable for Python.

### DIFF
--- a/test/api/pyapi/testcases/fileinit/test_fileinit.py
+++ b/test/api/pyapi/testcases/fileinit/test_fileinit.py
@@ -31,8 +31,8 @@ NITER = 3
 
 
 def init_collective_variables():
-    '''Initialize collective variables, which will
-       hold summed up results of multiple test runs.
+    '''Initialize collective variables, which will hold summed up results of
+       multiple test runs.
     '''
 
     merminentot = 0.0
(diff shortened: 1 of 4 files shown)


vanderhe:
This PR only makes changes to docstrings and the documentation of the function arguments. At the time, I decided on a supposedly easier-to-read indentation style, which, however, hardly makes sense in terms of expandability and diff-behaviour. The code itself should already be inline with Pylint's style or a slightly modified config of it (at least as far as it makes sense to match that style).

Thank you for the suggested code formatters, I hadn't heard of them and will definitely take a look. It would be nice if all the Python-codes follow the same style. Of course, this assumes that the corresponding code formatters also handle docstrings properly.

awvwgk:
We should setup a linter workflow for the CI, like I did here: https://github.com/grimme-lab/xtb-python/blob/master/.github/workflows/python-linter.yml, the workflow can be limited to Python source in this repository and will not run on modifications on the Fortran source.

vanderhe:
What would your preferred indentation style look like? Can you provide a brief example?

awvwgk:
Usually trying to use NumPy style for docstring: https://numpydoc.readthedocs.io/en/latest/format.html, but this is more a matter of taste.

Something that slightly bothers we with the current style is the three space indentation that can result in cases where three space indentation and four space indentation clashes:

https://github.com/dftbplus/dftbplus/blob/0995fb6e2a23fd105d6b832e89a768660aa5e78a/tools/pythonapi/src/dftbplus.py#L274-L279

vanderhe:
I see, thanks. I like the idea of the automatic generation of reasonable documentation via Sphinx (this would certainly be worthwhile, especially for the somewhat more extensive dptools). I'll keep that in mind as a future project.

PR #605, opened 2020-08-07
Allow error handling in dispersion interface

awvwgk:
- allows passing a status indicator to the dispersion interface
- only require error handling from routines performing work, like `updateCoords`, since most other routines should use cached values or just update and invalidate internal caches
- `handleLatticeChanges` can propagate an error from a dispersion interface back to `processGeometry`
- DFT-D4 can propagate an error from the EEQ-container back to the caller
- the EEQ-container can return an error from the LAPACK wrapper back
- error handling is done by `HANDLE_ERROR` macro, which might later produce a stack trace on back propagation
- status indicator is currently not passed to `processGeometry`, therefore the LAPACK handler will still perform the kill on error (something for a later change)

Requires #601 to avoid merge conflicts later
Useful for #545 to implement error handling

### DIFF
--- a/prog/dftb+/include/error.fypp
+++ b/prog/dftb+/include/error.fypp
@@ -50,3 +50,13 @@
     end if
   end block
 #:enddef FORMATTED_ERROR_HANDLING
+
+
+#! Propagation of error handling, for now it just returns when in error
+#:def HANDLE_ERROR(errVar)
+  if (present(${errVar}$)) then
+    if (${errVar}$ /= 0) then
+      return
+    end if
+  end if
+#:enddef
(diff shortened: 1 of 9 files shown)


awvwgk:
Always happy to rebase.

@bhourahine Regarding your question, not an issue at all for git.

<details>
It is actually a more common situation in merge commit based git repos. We have here a situation like this:

```
– O – … – B – … – M
   \     /       /
    … – A – … – C
```

After you merged A into master (B) the commits in my branch become part of the master branch, therefore any branch starting from my branch will not start from the original branch point O, but from A when comparing to master again for another merge commit (M).

That's actually on of the reasons why I try to avoid merge commits in my projects at all cost.
</details>

PR #609, opened 2020-08-08
Add regression tests and restructure dptools

vanderhe:
I have just rediscovered an old project of mine that I have completely lost sight of over the last few years and I think that there is still a need for it. The PR supplements the dptools regression tests with missing elements and also restructures a bit (for some reason at the time I thought that these tasks belong together). The restructuring mainly relates to the introduction of command-line wrappers, which were not present for all tools.

I will leave the whole thing as a draft for now, as I would like to hear your opinion on it beforehand.


awvwgk:
I don't have much experience with dptools, but this could be a useful addition. Somebody of the @dftbplus/maintainers should have a look.

bhourahine (commented):
Still need to check for changes in the moved routines after the branch point (i.e. things that were not copied over by git as the code chunk had changed file). But otherwise about ready.

vanderhe:
Thanks for revising some outdated lines of code that I missed. If anything else needs to be fixed, that I can help with, just let me know.

bhourahine:
Its really now just a case of confirming (probably by checking line by line) whether there were any changes on the main trunk version where git would not map onto the moved routine locations in the branch.

vanderhe:
Since I was able to rebase on current master, I assume that no changes have been made. But I'm not sure if this is a guarantee for that...

To be on the safe side, I will have a look at the `git diff ...` of the affected files between the old (before my changes) and current HEAD as soon as possible (during this week). Honestly, I don't get why a line-by-line check is necessary.

bhourahine:
> Honestly, I don't get why a line-by-line check is necessary.

Unfortunately if you branch and then (as git would track it) delete part of a file and add a new unrelated file, there is nothing to link that new file to any changes happening on the originating branch in the original code block. If you are lucky there is a merge conflict, but that doesn't always happen, depending on branch histories and intermediate merges.

vanderhe:
I see, that's unfortunate.

I just checked if there have been any changes to _dptools_ since then; couldn't find any in the corresponding files. So I don't see a problem in this regard. Still, if it doesn't involve too much effort, a second person should possibly confirm.

bhourahine (approved):
Ready to go (haven't found any regressions)

PR #610, opened 2020-08-08
Add Latex and test intermediate files to .gitignore files

jhrmnn:


test/prog/dftb+/.gitignore
```diff
@@ -0,0 +1,5 @@
+/**/autotest.tag
+/**/band.out
+/**/charges.bin
```
bhourahine, line 3: There are various other *.bin output files that probably should be ignored (probably should be done in a later PR, as there also number of other text files that don't fit a particular pattern so needs discussion before any rationalisation/clean up and ignore).

jhrmnn: No reason whatsoever. I started with `/test/**/<filename>` in the top-level gitignore, but then decided to move it to lower directories, since, as you point out, one might anyway need to thing this through more thoroughly, and perhaps be relatively fine-grained in the patterns.

Changed it to just the filename now.

PR #616, opened 2020-08-12
Bug fix for unallocated write attempt

bhourahine:
qInput is not allocated for REKS, so should not be printed

### DIFF
--- a/prog/dftb+/lib_dftbplus/main.F90
+++ b/prog/dftb+/lib_dftbplus/main.F90
@@ -188,7 +188,7 @@ contains
       end if
     #:endif
       tWriteCharges = tWriteRestart .and. tMulliken .and. tSccCalc .and. .not. tDerivs&
-          & .and. maxSccIter > 1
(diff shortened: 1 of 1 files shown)


prog/dftb+/lib_dftbplus/main.F90
```diff
@@ -188,7 +188,7 @@ subroutine runDftbPlus(env)
       end if
     #:endif
       tWriteCharges = tWriteRestart .and. tMulliken .and. tSccCalc .and. .not. tDerivs&
-          & .and. maxSccIter > 1
+          & .and. maxSccIter > 1 .and. .not.allocated(reks)
```
awvwgk, line 191: Wouldn't it be easier to check the allocation status of `qInput` instead?

aradi: I agree, let's look for `allocated(qInput)` instead, and maybe as the first one in the condition list.

awvwgk (approved):

PR #618, opened 2020-08-14
Allow disabling installation of Python components

awvwgk:
- useful to avoid Python runtime dependencies of the final distribution (conda-forge)


bhourahine (approved):
Ready to merge

PR #623, opened 2020-08-17
Module to wrap targets with read only pointers

bhourahine:
Add routines to wrap large arrays as pointers such that they remain writable for routines with the target data structure but read only for where only the pointer structure is available.

- [x] Basic data structures
- [x] Simple operations (get shapes and slices)
- [ ] Use with arrays that need update status in modules outside the core (coords, neighbours, ...)
- [ ] Stubs for abstracting dense matrix models if distributed ( / co-arrays ? ) in future

### DIFF
--- a/prog/dftb+/lib_common/CMakeLists.txt
+++ b/prog/dftb+/lib_common/CMakeLists.txt
@@ -15,7 +15,8 @@ set(sources-fpp
   ${curdir}/schedule.F90
   ${curdir}/timer.F90
   ${curdir}/timerarray.F90
-  ${curdir}/unitconversion.F90)
+  ${curdir}/unitconversion.F90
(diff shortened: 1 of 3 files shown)


aradi:
It may add a lot of complexity and deteriorate performance if we wrap various different array operations for the sake of read-only access... Probably, it is easier to expose them directly (and wait for my proposal of the `protected` attribute in derived types to be accepted by the standard committee :wink:)

bhourahine:
Something to discuss, as having different modules all holding coordinates (and worse, neighbour maps) is a mess.
So protected will be in the F2030 standard then?

bhourahine:
As per discussion, will be superseded by refactoring of types after 20.2 release.

PR #625, opened 2020-08-18
Read eigenvectors as initial guess for REKS

inseonglee:
Add normalization of eigenvectors by using unitary transformation.
Currently, the option `ReadEigenvectors` can be used when REKS block is used.

It seems to be good to move the parser (`ReadEigenvectors`) routine from `REKS` to `Hamiltonian` block.
e.g. Hamiltonian = DFTB { ReadEigenvectors = Yes }
Then, this option can be generally used for several methods.


bhourahine:
This does not seem to build with MPI enabled as there is a call to gemm but it is not imported. Probably there needs to either be some more preprocessor wraps around this routine, or gemm should always available.

inseonglee:
Now gemm is always available regardless of MPI.

prog/dftb+/lib_dftbplus/main.F90
```diff
@@ -6247,6 +6247,80 @@ subroutine getReksInitialSettings(env, denseDesc, h0, over, neighbourList, &
...
+    do ii = 1, nOrb
+      tmpS(ii,ii) = 1.0_dp
+    end do
+
+    ! Diagonalize CSC to obtain a unitary matrix, U = CSC^(-1/2)
+
+    tmpEigen(:) = 0.0_dp
+    call env%globalTimer%startTimer(globalTimers%diagonalization)
+    ! tmpC becomes eigenvectors (X) of CSC
+    call diagDenseMtx(electronicSolver, 'V', tmpC, tmpS, tmpEigen)
```
bhourahine, line 6300: Any reason to use a generalized eigensolver if tmpS = I ?

prog/dftb+/lib_dftbplus/main.F90
```diff
@@ -6247,6 +6247,80 @@ subroutine getReksInitialSettings(env, denseDesc, h0, over, neighbourList, &
...
+    call env%globalTimer%stopTimer(globalTimers%diagonalization)
+
+    ! Make inverse square root matrix consisting of eigenvalues (s) of CSC
+    tmpS(:,:) = 0.0_dp
+    do ii = 1, nOrb
+      tmpS(ii,ii) = 1.0_dp / sqrt(tmpEigen(ii))
```
bhourahine, line 6306: What happens in the case of of eigenvalues being ~0 ? Presumably if S_new is well conditioned, this should not be a problem, but it might be worth trapping for this.

prog/dftb+/lib_dftbplus/main.F90
inseonglee: Is there a specific eigensolver to diagonalize certain matrix without tmpS? If then, `solveSecularEqn` routine also can be modified.

bhourahine: In the dftbp_eigensolver module, heev wraps the ?syev and ?heev symetric/hermitian solvers.

prog/dftb+/lib_dftbplus/main.F90
inseonglee: Then, how about to print warning message (or error) when the eigenvalues are close to zero? In test case (ethylene molecule), the eigenvalues are almost ~1.0 at many geometry steps and its maximum (minimum) values are ~1.5 (~0.6) at earlier step. Probably, 1.9 (0.1) would be proper standard? If you agree this, then I will add this check line.

prog/dftb+/lib_dftbplus/main.F90
inseonglee: Now REKS uses heev routine for diagonalization without an identity matrix. Please see the commit 050577b.

prog/dftb+/lib_dftbplus/main.F90
inseonglee: From additional test (commit 88af4af), the eigenvalues of ~0.3 or ~3.9 give correct convergence, so it is ambiguous to find a proper standard for eigenvalues in this case.

inseonglee:
Except checking of zero eigenvalues, other tasks are finished. (change of charge convergence in REKS, automatic use of eigenvectors of previous steps in geometry optimisation, use of heev routines, ...)

Any recommendations to check zero eigenvalues?

prog/dftb+/lib_dftbplus/main.F90
bhourahine: Numerically it is only if the magnitude is below epsilon that there could be a problem. This is possible for badly behaved overlap, but probably just testing for this and stopping or just mildly regularize 1.0_dp/(max(sqrt(tmpEigen(ii)),epsilon(0.0_dp))

bhourahine:
Generally looks ready to merge, but unfortunately the test example binary read causes some problems. As this is using a binary read, it is not cross platform and even on x86_64 at least one compiler fails to read the initial integer in the file. Could you either make
 
- a 2 step test (as is done for other tests cases), where a first calculation writes the eigenvecs.bin in the native format and the second reads that file. An example is in the test/prog/dftb+/scc/GaAs_2_restart where the shell script testrun.sh is called first by the test system.

- the code read an ascii file instead of unformatted binary (its less precise, but will be cross platform). The complication here is that the ascii eigenvector data is not written to high precision at the moment.

inseonglee:
@bhourahine Now the test of reading mo is fixed, please see new commits.

bhourahine:
There are some problems with the record length as the first value in the file is an integer, not real(dp), so the read in still fails with particular compilers as in that case the record lengths are  different.

Also, for the test example, only the integer and then 12 real(dp) values are being saved into the .bin file. Is this expected behaviour? For usual DFTB, there would the 12 * 12 * nSpin values for the single particle states.

bhourahine:
> There are some problems with the record length as the first value in the file is an integer, not real(dp), so the read in still fails with particular compilers as in that case the record lengths are different.

Possibly now fixed.

inseonglee:
@bhourahine Thanks for fixing the record problem, now it works normally in my cluster too. It seems to be good.

bhourahine (approved):
Ready to merge.At some point waveplot needs refactoring to use the read routines (and to plot REKS states correctly).

PR #631, opened 2020-08-25
Self-consistent DFT-D4

awvwgk:
- [x] split DFT-D4 references from DFT-D4 calculator
- [x] reduce memory usage by reusing local arrays
- [x] remodel ATM and two-body dispersion to use same interface
- [x] adjust CN container interfaces to match EEQ container interfaces
- [x] add updateCharges procedure to abstract dispersion interface
- [x] add updateCharges call to getChargesPotential in Hamiltonian
- [x] expand gemv interface to handle 4x2->2 contraction
- [x] add self-consistent D4 container
- [x] add missing contribution to gradient
- [x] add regression tests
- [x] document the new feature
- [x] include a note in the changelog


bhourahine (commented):
Probably needs to print to stdout that this is self consistent d4, (at the moment the message is just "Using DFT-D4 dispersion corrections").

prog/dftb+/lib_dftb/dftd4param.F90
```diff
@@ -17,59 +17,67 @@ module dftbp_dftd4param
...
   interface getSqrtZr4r2
     module procedure :: getSqrtZr4r2Symbol
     module procedure :: getSqrtZr4r2Number
   end interface getSqrtZr4r2
 
 
+  interface init
+    module procedure :: TDftD4Calculator_init
+    module procedure :: TDftD4Ref_init
+  end interface
```
aradi, line 77: I'd suggest not to overload, but use the explicit `T*_init()` routines when initializing the objects. (Usually, error messages due to argument mismatch on an overloaded routine are not very informative.)

prog/dftb+/lib_dftb/dftd4param.F90
```diff
@@ -83,48 +91,51 @@ module dftbp_dftd4param
...
-    !> Maximum charge scaling height for partial charge extrapolation.
+    !> Maximum charge scaling height for partial charge extrapolation
     real(dp) :: chargeScale = 3.0_dp
 
-    !> Charge scaling steepness for partial charge extrapolation.
+    !> Charge scaling steepness for partial charge extrapolation
     real(dp) :: chargeSteepness = 2.0_dp
 
+    !> Dispersion is evaluated self-consistent
+    logical :: tSelfConsistent = .false.
```
aradi, line 135: You don't have to prefix with `t` for logicals, could also use the more natural `selfConsistent` (or `isSelfConsistent`) instead.

prog/dftb+/lib_dftb/dispdftd4.F90
```diff
@@ -136,55 +183,113 @@ subroutine DispDftD4_init(this, inp, nAtom, speciesNames, latVecs)
...
+
+    !> Instance of the type to describe
+    class(TDispDftD4), intent(in) :: this
+
+    if (allocated(this%sc)) then
+      write(unit, "(A)") "Using self-consistent DFT-D4 dispersion corrections"
+    else
+      write(unit, "(A)") "Using DFT-D4 dispersion corrections"
+    end if
+  end subroutine writeDftD4Info
```
aradi, line 239: What about making this a type bound procedure, returning `.true.` or `.false.` and letting the caller decide about the best way to render this information. Alternatively, returning an allocatable character with `"self-consistent"` or `"non-self-consistent"` would make also sense to me, but still, with rendering on the caller side. (So, that in a far future we can write a good unit test for this functionality as well :wink:)

aradi (approved):
Apart of the minor issues OK from my side. (We really need to come up with a better handling / passing around of geometry information...)

prog/dftb+/lib_dftb/dispdftd4.F90
awvwgk: The solvation model uses a similar strategy, the containers know how to display their information correctly, the main program only knows that the containers can display this information. This strategy is necessary in case we have to access module private information for display. Returning a single string is not really a solution if the container becomes more complex.

aradi: OK, I see. Than just leave this for the moment as it is.

PR #632, opened 2020-09-03
Add missing patterns to set_version script

aradi:
I hope, I included them all.


bhourahine:
Misses ./doc/api/doxygen/Doxyfile
But also should probably do something (print a reminder?) about the need to update getParserVersion in parser.F90.

aradi:
Valid points, thanks. Should be fixed now.

bhourahine (approved):
Ready to go

PR #633, opened 2020-09-07
Restore testing for atoms on top of each other

bhourahine:
Also changed some use statements to be more specific and name convention for initialisation of types.


awvwgk (commented):
Can't you just set a logical when this warning is generated or turn it directly into an error? Going through the complete list again shouldn't be necessary.

https://github.com/dftbplus/dftbplus/blob/67de54b5af73bc88613214e6bc121f05f4607615/prog/dftb%2B/lib_dftb/periodic.F90#L536-L538

PR #638, opened 2020-09-18
WIP: Implement coverage analysis

aradi:
Enables local coverage analysis with lcov/genhtml. Coverage info uploading to codecov.io will be done separately.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -376,10 +376,13 @@ CMake configuration file, as it is declared as dependency in the DFTB+ Cmake
 config file.
 
 
-Generating developer documentation
-==================================
+Additional information for developers
+=====================================
 
-Developer documentation can be generated using the FORD source code
+Source code documentation
+-------------------------
+
+Source code documentation can be generated using the FORD source code
 documentation generator by issuing ::
 
   cd doc/dftb+/ford && ford dftbplus-project-file.md
@@ -388,8 +391,8 @@ in the main source directory. The documentation will be created in the
 `doc/dftb+/ford/doc` folder.
 
 
-Developer build instructions
-============================
+Customized developer builds
+---------------------------
 
 You should avoid to customize the build by changing the variables in the CMake
 config files directly as your changes may accidently be checked in into the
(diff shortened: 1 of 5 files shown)


INSTALL.rst
```diff
@@ -406,3 +409,36 @@ your config file contains toolchain dependent options, consider to define the
 See this `CMake customization file
 <https://gist.github.com/aradi/39ab88acfbacc3b2f44d1e41e4da15e7>`_ for a
 template.
+
+
+Coverage analysis
+-----------------
+
+You generate coverage reports locally using the ``lcov`` and the ``genhtml``
```
bhourahine, line 417: 'You can generate'

INSTALL.rst
```diff
@@ -406,3 +409,36 @@ your config file contains toolchain dependent options, consider to define the
...
+tools, provided they are installed on your system (and CMake can find them). The
+workflow (assuming ``make`` as build backend) is then as follows:
```
bhourahine, line 419: 'as the build backend'

bhourahine (approved):
Seems ready to go (see comments for some minor documentation suggestions).

awvwgk:
Upload seems to work fine:

- [upload](https://github.com/awvwgk/dftbplus/commit/cdff062778a5a82c822c9de5e02a767e34643553)
- [build](https://travis-ci.com/github/awvwgk/dftbplus/builds/185234015)
- [report](https://codecov.io/gh/awvwgk/dftbplus/commit/cdff062778a5a82c822c9de5e02a767e34643553)

The change required is minimal, I suggest we include it this PR already.
I also made a PR against Bálint's branch to see if the codecov commenter works as expected (https://github.com/aradi/dftbplus/pull/19).

awvwgk (commented):
The coverage report produced is targeting the preprocessed files by default, which makes this analysis almost useless for the user, unless the mapping back to the original source file is trivial (no preprocessor used, which is hardly ever the case).

Setting the `-n` flag for `fypp` should solve the issue, it might be worth to do this automatically when the coverage target is requested.

awvwgk:
Another thing I noticed, setting the `-n` preprocessor flag for the coverage build will increase the CI time by a factor of two, means generating coverage reports is not for free.

aradi:
> Another thing I noticed, setting the `-n` preprocessor flag for the coverage build will increase the CI time by a factor of two, means generating coverage reports is not for free.

Wow, really? This is somewhat scaring...

Otherwise, for sure, we should add the `-n` as default when doing coverage tests.

awvwgk:
> Wow, really? This is somewhat scaring...

Check [build with coverage](https://travis-ci.com/github/awvwgk/dftbplus/builds/185234015) (24 min) and [build with coverage and -n](https://travis-ci.com/github/awvwgk/dftbplus/builds/185238021) (55 min). Not sure what is going on there, but I ran both CIs in quick succession, so it shouldn't be the machine. Maybe we can check this locally or on dpdevel first.

PR #639, opened 2020-09-19
Minor typo and print clean up

bhourahine:


prog/dftb+/lib_dftbplus/mainio.F90
```diff
@@ -4765,12 +4765,16 @@ subroutine writeReksDetailedOut1(fd, nGeoSteps, iGeoStep, tMD, tDerivs, &
...
+        write(fd, "(1X,A)") "SA-REKS optimizes the averaged state, not individual states."
+        write(fd, "(1X,A)") "These charges are not from individual states."
+        write(fd, "(1X,A)") "Similarly, the values in band.out file indicate"
+        write(fd, "(1X,A)") "the band energies and occupations for the averaged state."
+        if (.not.reks%tRD) then
+          write(fd, "(1X,A)") "If you want to compute the relaxed density,"
+          write(fd, "(1X,A)") "please, set 'RelaxedDensity = Yes' option"
+        else
+          write(fd, "(1X,A)") "These results are for a relaxed density."
+        end if
```
inseonglee, line 4777: Here the charges and populations given in detailed.out file do not correspond to the those for relaxed properties. The relaxed charges are only given in relaxed_charges.dat file when reks%tRD is true. The data (charges, populations) given in detailed.out and occupations in band.out are always from the averaged state.

PR #644, opened 2020-09-26
Return printing of alpha parameter

### LINKED ISSUE #641: Output no longer contains Ewald alpha parameter
**Describe the bug**
The DFTB+ manuals (v17.1 through v20.1) say that the Ewald alpha parameter is written to the (standard) output (for SCC = Yes and Periodic = Yes). However, this seems only true for the 17.1 release, not the more recent ones.

**To Reproduce**
1. Run DFTB+ >= v18 on any periodic structure, with SCC
2. Search in vain for the Ewald parameter in the standard output (or any of the other output files)

Am I missing something? Looks like this section of the manual simply wasn't updated.
If possible, I would prefer DFTB+ to again print out the Ewald parameter :)

bhourahine:
Prints the alpha parameter for the initial geometry.

Closes #641


awvwgk (approved):
Good to merge from my side.

PR #646, opened 2020-09-30
Fix 2 contacts

pecchia:
I have enabled an unused feature of libNEGF, namely fast transmission for 2 contacts. This restricts the GF calculation to 1 block and also saves memory. However PDOS cannot be computed in this case.
Since TunnelingAndDos{} by defaults computes the PDOS on the whole central region I added a tag computeLDOS = Yes|No. 
If set to No, and there are 2 contacts, the code falls into the fast routine for transmission.

I modified the test graphene_x to check this new feature. Other tests are already checking the DOS.


bhourahine (commented):
Looks like there is an undefined variable, which intel19 picks up in debug builds:

forrtl: severe (194): Run-Time Check Failure. The variable 'iterative_MP_calculate_transmissions_$NCONT' is being used in '/mnt/scratch/hourahine/dftbplus/build/external/libnegf/origin/src/iterative.f90(3269,15)' without being defined

bhourahine (approved):
Ready to merge

Issue #212, opened 2019-01-29
MPI externals missing from release


mstud:
Hi, the release tar.gz unfortunately does not include `mpifx` and `scalapackfx` but instead only an empty directory `origin` because you just linked the external repos in your git tree probably.
This leads to a compile error when setting `WITH_MPI := 1` because the source files for those external dependencies won't be found.
The `get_opt_externals` script also does not seem to take care of this, nor is it documented anywhere that one has to download those sources manually before the MPI version can be built.
Is this intentional?

mstud:
Okay, thanks, guess this issue can be closed then.

PR #651, opened 2020-10-22
Submodules updated automatically using CMake

william-dawson:
DFTB+ Developers,

As you know, DFTB+ depends on various git submodules. As a result, when you clone the repository you have to remember to type `git submodule update --init --recursive`. To me, it seems reasonable to just have cmake do this automatically during the configure step. I have inserted some code into the CMakeLists to do just that in this pull request. Unfortunately, this solution doesn't automatically pull the submodules in the case of the github release (#212), but it will only print a warning about the submodules and not stop compilation.

Of course, if you prefer to keep this a manual process I understand. Whatever you think!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,6 +62,15 @@ set(EXPORTED_EXTERNAL_LIBRARY_DIRS)
 set(EXPORTED_EXTERNAL_PACKAGES)
 set(EXPORTED_EXPORTED_COMPILED_LIBRARIES)
 
+# Pull git submodules into the repository.
+find_package(Git)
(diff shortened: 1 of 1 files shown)


william-dawson:
Sorry, I didn't see #636!

aradi:
Just to let yo know: inspired by your idea, I have extended #636 now with the possibility to automatically retrieve the submodules via `git submodule`. (I just had to add some tests to make sure, it does not overwrite developers submodules during development...). I think, we can now handle all different scenarios in a very robust way. So, thanks anyway.

william-dawson:
That sounds great, thanks @aradi!

PR #656, opened 2020-10-26
Use Travis for OSX testing

aradi:


aradi:
This seems to work now:

* It disables GitHub's test and uses Travis for OSX testing.
* It uses Python 3.8 for testing as Numpy is broken for Python 3.9 on OSX.

.travis.yml
```diff
@@ -18,20 +23,28 @@ addons:
       - libscalapack-openmpi-dev
       - libarpack2-dev
       - pkg-config
+  homebrew:
+    packages:
+      - arpack
+      - python@3.8
```
awvwgk, line 29: Maybe explicitly install GCC with ``gcc@9`` or ``gcc@10`` as well?

aradi: Valid point, indeed. I just tried it. :-1:  Now I really tend to leave away all homebrew package installation and reduce testing instead. (Adding `gcc@10` upgrades to from gcc 10.2 to gcc 10.2_1. Adding `arpack` installs among others latest python 3.9 (why?!) colliding with python 3.8 already present, causing homebrew exiting with an error...) I don't see an easy way to establish a reliable testing framework if homebrew is invoked...

PR #661, opened 2020-10-31
Fire

bhourahine:
FIRE geometry optimizer, can sometimes be competitive with lBFGS in number of steps.


doc/dftb+/manual/dftbp.tex
```diff
@@ -495,6 +497,24 @@ \subsection{gDIIS\cb}
   \is{Memory} should between 3 and 20~\cite{NoceWrig06}.
 \end{description}
 
+\htcbsubsection{FIRE}
+\label{sec:dftbp.FIRE}
+
+The {\it fast inertial relaxation engine} of Bitzek {\it et
+  al.}~\cite{Bitzek-PRL-97-170201}, using the default values from their paper
+and a Leap-frog integrator. This can be faster than conjugate gradient and
+competitive with LBFGS in some cases.
```
awvwgk, line 506: I would note here that it might be a cheap alternative for large scale calculations, since the optimization step is not costing much. Also, I would add a warning, that the user should carefully evaluate if FIRE is working correctly for their systems before using it over other optimizers.

bhourahine: Not sure if the step cost is relevant due to the linear algebra dominant costs in solving the hamiltonian, but have added text about testing first.

awvwgk (approved):

PR #665, opened 2020-11-03
Simplify ELSI version check

aradi:
- Adding functions for comparing semantic versions
- Removing date-check (as we would have to add for every ELSI release a date-check explicitely) and relying on returned version number only.


awvwgk (commented):
I would recommend to implement one operator (like `<`) and implement all other operators based on this. Have a look at: https://github.com/fortran-lang/fpm/blob/master/fpm/src/fpm/versioning.f90

aradi:
@awvwgk Makes absolutely sense, the example is really nice. I've implemented something similar now.

prog/dftb+/lib_common/version.F90
```diff
@@ -0,0 +1,198 @@
...
+
+    !> Reference version (RHS of the comparison)
+    class(TVersion), intent(in) :: refVersion
+
+    !> Result of the comparison
+    logical :: compResult
+
+    compResult = version > refVersion
+    if (.not. compResult) then
+      compResult = version < refVersion
```
awvwgk, line 192:
```suggestion
      compResult = refVersion > version
```

prog/dftb+/lib_common/version.F90
```diff
@@ -0,0 +1,198 @@
...
+    end if
+    if (present(patch)) then
+      this%numbers(3) = patch
+    end if
+
+  end function TVersion_create
+
+
+  !> Checks whether two versions are greater.
+  pure function greater(version, refVersion) result(compResult)
```
awvwgk, line 83: Could be `elemental` as well (implying `pure`).

awvwgk (approved):
Much cleaner without the preprocessor :wink:. Some minor suggestions but otherwise looks good.

PR #666, opened 2020-11-04
Fix parser segfault

aradi:


bhourahine:
Is it worth also adding the input case that triggered this to the tests?

aradi:
I'm not so sure. The chance that we make exactly the same error (wrong child deleting on root) is very small. If it happens somewhere else (on a lower level), a test with the original input won't catch it. So, I'd recommend to merge it without test.

I have also changed the logic slightly, by always removing the `InputVersion` tag, even if it is an alias for the current parser version. That way, the behaviour is more consistent.

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -264,6 +261,29 @@ subroutine parseHsdTree(hsdTree, input, parserFlags)
...
+    integer, allocatable, intent(out) :: implicitParserVersion
+
+    type(fnode), pointer :: child, dummy
+    type(string) :: versionString
+
+    call getChild(root, "InputVersion", child, requested=.false.)
+    if (associated(child)) then
+      call getChildValue(child, "", versionString)
+      implicitParserVersion = parserVersionFromInputVersion(unquote(char(versionString)), child)
+      dummy => removeChild(root, child)
```
awvwgk, line 280: If I understand `removeChild` correctly, `dummy` will be pointing to `child` after the child value is disconnected from the root. Wouldn't it be more intuitive to destroy the `dummy` node afterwards (even if `dummy` and `child` are pointing to the same thing)?

awvwgk (commented):
~~This PR changes how `InputVersion` is handled in the processed input. On master `InputVersion` is currently preserved when it matches the current release version and only converted to `ParserVersion` if it points to an older release version, while with this PR `InputVersion` is always converted to `ParserVersion`.~~ Already mentioned above, took me to long to review this...

Both is fine with me.

bhourahine:
> I'm not so sure. The chance that we make exactly the same error (wrong child deleting on root) is very small. If it happens somewhere else (on a lower level), a test with the original input won't catch it. So, I'd recommend to merge it without test.

Its more to have an absolutely minimal input example which runs but relies on defaults for almost everything.

prog/dftb+/lib_dftbplus/parser.F90
aradi: I agree, that is didactically indeed more consistent.

PR #667, opened 2020-11-05
Enclose global variables of dftbp_initprogram in a derived type

tomaskubar:
A new type TGlobalData is introduced, as is its instance globalData.

There are 3 issues, hopefully easy to resolve:
o) I am afraid this also includes some changes we had made to the API :-/ The largest is that the list of atoms and species is passed to DFTB+ through the API, and consequently no Geometry block is required in the input. I think I should remove this change, or just let me know what to do with it.
o) The array *shiftPerLUp* is only allocated if WITH_TRANSPORT (in initTransportArrays), but it seems it may be needed even in other cases (in addChargePotentials, hamiltonian.F90). Consequently, when compiled without transport, the test dftb+_poisson/CH4-poisson fails.
o) All tests pass on a local computer, but a few fail on Github. Not sure what the problem is.


aradi:
I've rebased your changes and also squashed them (partially, still keeping your contributions personalized). Do not work in this PR any more. If you want to add something or base further work on your changes, see #678 . Note, the code here is still buggy, it fails for MPI-builds...

PR #668, opened 2020-11-05
lBFGS line search option to disable

bhourahine:
Doesn't fix the line search algorithm, but gives the option of performing a quasi-Newton step instead.


awvwgk (changes requested):
The LBFGS Tests should enable the linesearch in the test input for reproducibility:

- dftb+_geoopt/H2O_lbfgs
- dftb+_halogen/3492
- dftb+_halogen/3501
- dftb+_geoopt/GaAs_8_latconst_lbfgs

awvwgk (approved):
Looks like a good initial step.

PR #670, opened 2020-11-09
Optional linemin routine for lbfgs plus oldcompat

bhourahine:
Also fixes 447c6726 to use old compatibility instead of changing tests.

Finally, fixes broken test label for H2O_lbfgsStep in tests (note, this was broken but not detected).


prog/dftb+/lib_geoopt/lbfgs.F90
```diff
@@ -139,7 +140,13 @@ module dftbp_lbfgs
...
     logical :: isLineSearch
 
+    !> Is the original line search used
+    logical :: isOldLSUsed
+
     !> line minimizer
+    type(TLineMin) :: lineMin
+
+    !> original internal line search
     type(TLineSearch) :: lineSearch
```
aradi, line 150: It would be more appropriate to make them `allocatable` and allocate only the one which is needed.

aradi (approved):

PR #676, opened 2020-11-17
Update the set version script

bhourahine:


aradi:
I've rebased and extended and merged it...

PR #678, opened 2020-11-24
Eliminate globals

aradi:
Packaging all globals into derived types to enable multiple instances


aradi:
As far as I can see, it is fairly complete now. The only places with remaining globals should be the ARPACK-interface and the Poisson/MUDPACK interface. Former is optional, latter will be made optional in a followup PR.

aradi:
Main work done. Further refinements should be made in followup PRs.

aradi:
Please note, that this PR does not contain the libNEGF fix from stage-20.2.1, so that the failing OSX tests are with high probabilty due to the bug in libNEGF...

prog/dftb+/api/mm/mmapi.F90
```diff
@@ -165,11 +254,11 @@ subroutine TDftbPlus_init(this, outputUnit, mpiComm)
       stdOut = output_unit
     end if
 
-    if (nDftbPlusCalc /= 0) then
-      write(stdOut, "(A)") "Error: only one DFTB+ instance is currently allowed"
-      stop
-    end if
-    nDftbPlusCalc = 1
+!   if (nDftbPlusCalc /= 0) then
```
bhourahine, line 257: Why not fully remove instead of commenting out?

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -120,933 +117,995 @@ module dftbp_initprogram
...
+  type :: TNegfInt
+  end type TNegfInt
+
+#:endif  
+
+! private
 
   !> Tagged output files (machine readable)
-  character(*), parameter :: autotestTag = "autotest.tag"
+  character(*), parameter :: autotestTag = "autotest.tag" ! parameter
```
bhourahine, line 131: Why comment that an obvious parameter is a parameter?

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -120,933 +117,995 @@ module dftbp_initprogram
...
 
-  !> Is the check-sum for charges read externally to be used?
-  logical :: tSkipChrgChecksum
+    !> MD integrator
+    type(TMDIntegrator), allocatable :: pMDIntegrator
 
-  !> reference neutral atomic occupations
-  real(dp), allocatable :: q0(:, :, :)
+    !> Temperature profile driver in MD
+    type(TTempProfile), allocatable :: temperatureProfile ! target
```
bhourahine, line 560: Why the target comment?

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -120,933 +117,995 @@ module dftbp_initprogram
...
+    !> data type for REKS
+    type(TReksCalc), allocatable :: reks
+
+    !> atomic charge contribution in excited state
+    real(dp), allocatable :: dQAtomEx(:)
+
+  contains
+
+    procedure :: initProgramVariables
+!   procedure :: inputCoherenceCheck
```
bhourahine, line 1060: Why leave this as a comment?

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -4085,44 +4064,53 @@ end subroutine checkTransportRanges
...
     end if
 
-    @:SAFE_DEALLOC(sccCalc, img2CentCell, species, species0, coord, coord0)
-    @:SAFE_DEALLOC(latVec, origin, recVec, invLatVec, cellVec, rCellVec, iCellVec)
-    @:SAFE_DEALLOC(neighbourList, nNeighbourSk, nNeighbourRep, iSparseStart)
-    @:SAFE_DEALLOC(hubbU, atomEigVal, referenceN0, mass, speciesMass)
-    @:SAFE_DEALLOC(ham, iHam, chargePerShell, chargePerAtom, over, kPoint, kWeight)
-    @:SAFE_DEALLOC(nEl, spinW, xi, UJ, nUJ, niUJ, iUJ, Ef, esp)
-    @:SAFE_DEALLOC(indMovedAtom, conAtom, conVec, pipekMezey)
+    @:SAFE_DEALLOC(this%sccCalc, this%img2CentCell, this%species, this%species0, this%coord)
```
bhourahine, line 4081: why not deallocate (this) instead of its components?

prog/dftb+/lib_dftbplus/parser.F90
```diff
@@ -135,7 +135,7 @@ subroutine parseHsdTree(hsdTree, input, parserFlags)
     type(fnode), pointer :: hsdTree
 
     !> Returns initialised input variables on exit
-    type(TInputData), intent(out) :: input
+    type(TInputData), intent(inout) :: input
```
bhourahine, line 138: What is the reason for requiring (inout) instead of (out)?

prog/dftb+/lib_io/hsdparser.F90
```diff
@@ -131,28 +131,17 @@ module dftbp_hsdparser
   !> list label
   character(len=*), parameter :: attrList = "l"
 
-
-  !> Length of a parsed line
+  !> Length of a parsed line (adjust lineFormat below, if you change this number)
   integer, parameter :: lc = 1024
 
+  !> Format string to read in one parse line (field length must correspond to lc above)
+  character(len=*), parameter :: lineFormat = "(A1024)"
```
bhourahine, line 138: Strong case for using preprocessing, to set 1024 only once for its use in 3 places (lc, lineFormat and MAXRECL).

prog/dftb+/lib_math/eigensolver.F90
```diff
@@ -29,10 +29,6 @@ module dftbp_eigensolver
 #:endif
 
 
-  !> Used to return runtime diagnostics
-  character(len=100) :: error_string
```
bhourahine, line 33: Not clear why this was replaced with copies in each subroutine.

prog/dftb+/lib_math/lapackroutines.F90
```diff
@@ -22,10 +22,6 @@ module dftbp_lapackroutines
   private
 
 
-  !> Used to return runtime diagnostics
-  character(len=100) :: error_string
```
bhourahine, line 26: same comment as eigensolver.F90

prog/dftb+/lib_timedep/dynamics.F90
```diff
@@ -293,9 +293,6 @@ module dftbp_timeprop
   !> Prefix for dump files for restart
   character(*), parameter :: restartFileName = 'tddump'
 
-  !> Used to return runtime diagnostics
-  character(len=120) :: error_string
```
bhourahine, line 297: same comment as eigensolver.F90

bhourahine (commented):
See various comments (including comments about redundant comments), but the main thing that is missing is a tester that runs different instances simultaneously  in different threads/OMP parallel region.

bhourahine:
Intel18 can't compile the current state of the branch:

negf_int.f90(1106): error #7496: A non-pointer actual argument shall have a TARGET attribute when associated with a pointer dummy argument.   [CSRHAM]
      call negf_density(this%negf, iSCCIter, iS, iK, this%csrHam, this%csrOver, mu(:,iS),&
----------------------------------------------------------^

prog/dftb+/lib_math/eigensolver.F90
aradi: To really eliminate all global variables (although I agree, that collision here is unlikely, but still). On the module level, only constants should occur.

aradi:
I'll comment on / fix the issues you raise. As for the tester: We can only add this, once we can require a build with reentrant components only. That should be covered in an additional PR.

prog/dftb+/lib_dftbplus/initprogram.F90
aradi: You're absolutely right, it came from 'mechanical transformation'. I've changed the logic now.

prog/dftb+/lib_dftbplus/parser.F90
aradi: Must have been a remainder of some intermediate state. Corrected.

prog/dftb+/lib_io/hsdparser.F90
aradi: Makes sense, done. But not for `MAXRECL`, which contains 1024 only by accident.

aradi:
The problem with Intel 18 is a compiler bug. I've added a workaround. (Intel 19 handles it already correctly).

aradi:
I've checked in the revised version. Although it does not have a proper test for the multiple instance scenario, I'd suggest to merge it and we handle the tests in the PR which introduces the build option ensuring thread-safe build.

bhourahine (approved):
Ready to go

PR #683, opened 2020-11-27
Warn about the slowness of TS gradient calculation

aradi:
Put some warnings about the slowness of the TS gradient calculation (as it uses finite differences).


awvwgk (approved):
Looks good to me.

aradi:
I'll rebase it on 20.2 and merge it against `stage-20.2.1` so that it gets part of the patch-release.

PR #684, opened 2020-11-30
Fix zero lattice derivative in detailed.out

aradi:
Hot fix for zero lattice derivatives in detailed.out. To be directly merged into preparation branch for 20.2.1.


bhourahine (approved):
Ready to merge

PR #686, opened 2020-11-30
Implement TNegfEnv

aradi:
Superceeds #682 with a cleaned up history and based on current master.


utils/test/check_submodule_commits
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python3.5
```
bhourahine, line 1: Any reason to pin this to a specific python release?

bhourahine (approved):
Generally looks ready, see the comment about the script though.

aradi:
I have a few issues with the original implementation. As it just simply overrides the central `TMpiEnv` type currently, I think, it is more logical to integrate the functionality there and not to create a separate type and type instance which are not used at all. See a69921d for details. @bhourahine @pecchia let me know, whether you agree to it.

aradi:
Also, I've removed the `20.2.1` milestone, as this is not hotfix for 20.1. It shoud be merged IMO normally to master.

utils/test/check_submodule_commits
aradi: Comes from the original unsquashed code. I've removed that now.

pecchia:
I created a TNegfEnv exactly as BlacsEnv. If that can exist, why not negfEnv. There I could store in principles things needed for initialization. Also I think it is a good idea  in general to keep  negfMpi somewhere.

pecchia:
Conerning the script  there is an issue with  subprocess.run
This is supported starting from python 3.5. 
When I run the script as is, I get an error because python3 runs v 3.4.x
As usual my local problems, but is there a way to force a specific version ? I forgot to revert back the script before ci.

aradi:
The question here is, how much the main program must know about extension specific descriptors. `TMpiEnv` and `TBlacsEnv` are both used by the main code, to communicate between processes and to distribute matrices. `TNegfEnv` was effectively a copy of `TMpiEnv`, and created a communicator, which was then used do override the one in `env%mpi` anyway. So, there is not much point IMO to keep a separate type in the main code, if the main code uses `env%mpi` anyway. Let' s discuss about it at the next devel meeting more in detail.

As for the `python3.5` issue. You're right, I was not aware that it was only working from Python 3.5 and above. The easiest thing would be to set up your `PATH` environment variable, so that `python3.5` is found first, if `python3` is required. Currently, we do not have a way to set the python-version centrally for all scripts.

prog/dftb+/lib_extlibs/negf_int.F90
```diff
@@ -41,61 +41,53 @@ module negf_int
...
-  !> compressed sparse row overlap
-  type(z_CSR), target :: csrOver
-  type(Z_CSR), pointer :: pCsrOver => null()
+  end type TNegfInt
 
-  !> non wrapped direct calls
-  private :: negf_density, negf_current, negf_ldos
+  #:if WITH_MPI
+    !> Initialize MPI groups according to libNEGF needs
+    public :: negf_setup_mpi_communicators
```
bhourahine, line 78: Will need to become a procedure, or have its functionality migrated into mpienv.F90

bhourahine (approved):
Looks about ready to merge, but see comment.

prog/dftb+/lib_extlibs/negf_int.F90
aradi: The wrapper is indeed unnecessary, I've removed it.

bhourahine:
@aradi slight problem, the libnegf submodule doesn't point to the main libnegf repository (this has now been updated, so can be fixed).

bhourahine:
> @aradi slight problem, the libnegf submodule doesn't point to the main libnegf repository (this has now been updated, so can be fixed).

#697 addresses this

PR #687, opened 2020-11-30
Migrate CI to GH actions

awvwgk:
We cannot build on Travis CI anymore due to a change of mind regarding their Open Source policy. So we have to migrate to another provider now.

- [x] Create a GH actions workflow to test on Ubuntu and OSX
- [x] `git rm .travis.yml`


bhourahine (approved):
Looks ready to go

PR #688, opened 2020-12-02
Stage 20.2.1

aradi:
Proposal for hot-fix release. To be merged to `release` and tagged as `20.2.1`

PR #689, opened 2020-12-03
Fix detailed.out write case

bhourahine:
Fixes if logic around printing detailed.out files, which were not happening without dispersion combinations.


aradi:
I've applyied the same patch in PR #688. This wayit would enter 20.2.1. See 69dd06ec.

PR #690, opened 2020-12-03
Change to calculate dispersion once

bhourahine:
(unless self-consistent dispersion is used)


prog/dftb+/lib_dftbplus/main.F90
```diff
@@ -941,7 +941,7 @@ subroutine processGeometry(env, iGeoStep, iLatGeoStep, tWriteRestart, tStopScc,
           end if
         end if
 
-        if (allocated(dispersion)) then
+        if (allocated(dispersion) .and. .not. tConverged) then
```
aradi, line 944: Don't we need the same in the other branch of the reks-if?

bhourahine: Just changed this as well (and probably fixed energy summation for REKS in this case during SCC cycles).

awvwgk (changes requested):
The commit history seems to be polluted with some unrelated commits, could you rebase to clean this up?

bhourahine:
> The commit history seems to be polluted with some unrelated commits, could you rebase to clean this up?

This was a consequence of a rebase (I think colliding with that forced push to master the other week). Will try again.

bhourahine:
This should now be the cherry picked version, not sure whether the history is cleaner though.

awvwgk:
History is still polluted. Which commits are relevant for this PR? Is it only 2df521d349bef044504328caafe93fe041bb017a?

awvwgk:
@bhourahine Could you have a look at cc9b4a8ba832cec5f0176a318c42b4ca541b6316 and a590ec719e0ea363d920fa8d43ca0356a4725619?

To apply this commit to this PR use:

```
git branch -D singleDispCalc
git fetch https://github.com/dftbplus/dftbplus a590ec719e0ea363d920fa8d43ca0356a4725619
git checkout FETCH_HEAD
git switch -c singleDispCalc
git push --force git@github.com:bhourahine/dftbplus singleDispCalc
```

bhourahine:
Changes are in clean PR #718

PR #691, opened 2020-12-08
Add masses to MM API

fbonafe:
Add a function to retrieve the atomic masses array from initprogram.F90 and test for it within the extcharges test.


prog/dftb+/api/mm/mmapi.F90
```diff
@@ -438,6 +440,26 @@ function TDftbPlus_nrOfAtoms(this) result(nAtom)
   end function TDftbPlus_nrOfAtoms
 
 
+  !> Returns the atomic masses for each atom in the system.
+  function TDftbPlus_getAtomicMasses(this) result(mass)
```
awvwgk, line 444: Most of the getter procedures for array-valued quantities are implemented as subroutines rather than functions. the `getAtomicMasses` should follow this scheme as well.

awvwgk (changes requested):

awvwgk (approved):

Issue #692, opened 2020-12-21
Build fails for MSYS2 MinGW toolchain


awvwgk:
**Describe the bug**

While trying to apply the MinGW64 toolchain from [MSYS2](https://msys2.org) to a few projects I noticed DFTB+ does not work correctly in the MSYS2 bash shell.

**To Reproduce**

**Expected behaviour**

**Additional context**

awvwgk:
The shipped fypp doesn't work on a clean Windows build due to missing Python installation. For Windows builds it is therefore always required to specify fypp by `-DFYPP=$(which fypp)` (for MSYS2 bash, command might differ for CMD.exe or PowerShell).

awvwgk:
Found out that the `CustomLapack` module fails for default settings, but works fine if I explicitly provide the `-DLAPACK_LIBRARY=...` in the configuration step.

husakm:
I tested Cygwin  insetad of MSYS2. It worked out of box ...

PR #698, opened 2021-01-08
Add MinGW compilation to CI workflow

awvwgk:
Related to #692.

Does not test the resulting binaries due to the limitations of the DFTB+ testsuite (mainly due to symlink issues). Not sure if this makes a useful addition for the CI, therefore opening this for discussion.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,7 +87,7 @@ jobs:
 
     - name: Configure build
       run: >-
-        cmake -B _build -G Ninja
+        cmake -B ${BUILD_DIR} -G Ninja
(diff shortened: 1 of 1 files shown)


aradi:
I'd very much favor to have a minGW test/port. But testing the building process only is not really satisfactory IMO.

Actually, if it is just the symlinks for the SK-files, we could circumvent that by providing a special file (e.g. `SKFILES`) in each test directory, which lists the sk-files needed. (Had the actual additional benefit of not having to resymlink things, in case the location of the SK-files changes.)

husakm:
Cygwin based windows build works OK including running all tests and OMP support.

PR #702, opened 2021-01-16
Gamma point periodic range separation

bhourahine:
Ground state range separated calculation for periodic systems. Not fully checked and tested, but seems to show conditional convergence for diamond unit cell. Uses [1/r cutoff](https://dx.doi.org/10.1021/ct900494g) approach.

- [x] Gamma point
- [ ] Check energy and forces
- [ ] Stress
- [ ] Logic to safely set cutoff
- [ ] Regression tests

### DIFF
--- a/prog/dftb+/lib_dftb/rangeseparated.F90
+++ b/prog/dftb+/lib_dftb/rangeseparated.F90
@@ -111,7 +111,9 @@ module dftbp_rangeseparated
 
   contains
 
-    procedure :: updateCoords
+    procedure :: updateCoordsClust
+    procedure :: updateCoordsNeigh
+    generic :: updateCoords => updateCoordsClust, updateCoordsNeigh
     procedure :: addLrHamiltonian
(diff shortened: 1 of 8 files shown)


prog/dftb+/lib_dftb/rangeseparated.F90
```diff
@@ -293,7 +293,11 @@ subroutine updateCoordsNeigh(this, coords, nNeigh, iNeigh, img2CentCell)
 
     nAt = size(this%species)
     @:ASSERT(nAt == size(nNeigh))
-    this%coords(:,:) = coords(:, :nAt)
+    if (size(this%coords, dim=2) /= size(coords, dim=2)) then
+      deallocate(this%coords)
+      allocate(this%coords(3,size(coords, dim=2)))
+    end if
+    this%coords(:,:) = coords
```
awvwgk, line 300: Why not just use the automatic reallocation of the LHS directly rather than implementing it here?

bhourahine:
Closed by #947

PR #705, opened 2021-01-30
Print norm of gradient and lattice gradient in optimization

awvwgk:
I'm was missing a good way to track the optimization progress in DFTB+ so far since the used maximum force component is not always representative for the complete system. This PR adds a printout of a norm of the gradient while performing optimizations.

### DIFF
--- a/prog/dftb+/lib_dftbplus/main.F90
+++ b/prog/dftb+/lib_dftbplus/main.F90
@@ -6453,20 +6453,36 @@ contains
 
   end subroutine calcPipekMezeyLocalisation
 
+  
+  !> Prints information about maximal forces in the system.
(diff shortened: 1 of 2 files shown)


bhourahine (approved):
Fine to merge, but probably worth collapsing printMax* and print*Norm into a single routine in the force and lattice cases, where the norm2() and maxval(abs()) are applied to the same input array inside of the routine, as they are called together.

awvwgk:
I wasn't sure about the intent behind having fine grained separation of the printing routines in `mainio.F90` and just followed the existing scheme here.

aradi:
I've pushed a commit which averages the norms by the number of atoms / lattice vectors to make it size independent as we discussed. @awvwgk if it serves your purpose still well, it could be merged.

PR #706, opened 2021-01-31
Fixed atoms

bhourahine:
Addresses FixedAtoms request on mailing list (also fixes Hessian ranges if contacts are present).

- [x] keywords
- [x] parser restructure
- [x] documentation
- [ ] regression test cases


bhourahine:
Redundant due to #719

PR #709, opened 2021-02-05
Make imports explicit

helmutwecke:


aradi:
@helmutwecke That looks very good. Please apply it wherever it makes sense.

helmutwecke:
I already made all imports consistent. Now I'm going to run the buildbot tests. @aradi

helmutwecke:
I have already merged the actual master branch, solved the conflicts and added the explicit module imports where needed. I have also ran the tests, and all seems to be good. I think this branch is ready to be merged into master. @aradi

aradi:
I've updated it to current master and resolved conflicts. It does not compile yet, I'll check in the fix in a few minutes (hopefully).

aradi:
@helmutwecke Now, it seems to compile. Please continue from here. There are still a lot of  implicit imports all over the place. You could find them by
```sh
find . -name '*.F90' | xargs grep "^[ ]*use" | grep -v only | grep -v external | grep -v dftbp_assert
```
or something similar.

helmutwecke:
@aradi . I added the missing explicit imports, but there are some imports that do not give an error when they are commented out. I am going to run the buildbot tests to see if, the imports should be deleted or not

awvwgk:
There is a conflict for `prog/dftb+/lib_derivs/staticperturb.F90`, which has to be resolved first.

helmutwecke:
@aradi @awvwgk . I solved the conflict on staticperturb.F90. There should be no more problems

PR #710, opened 2021-02-08
Added script for elastic constants

bhourahine:


utils/numcheck/calcelastic
```diff
@@ -0,0 +1,215 @@
...
+    """Prints calculated elastic constants."""
+
+    for ii in range(6):
+        for jj in range(6):
+            elastic[ii][jj] /= (PRESURE_AU * 1.0E9) # Convert to GPa
+    print("\nElastic constants (GPa):")
+    for ii in range(6):
+        print (("%12.4f"*6) % tuple(elastic[ii]))
+
+
```
aradi, line 128: I guess, the routines below are just copy-paste versions of the one in `calcderivs`. Why not putting them into a Python module in this directory (e.g. `numchekhelpers`) and letting both programs import it?

utils/numcheck/calcelastic
```diff
@@ -0,0 +1,215 @@
...
+def readgen(fname):
+    """Reads in the content of a gen file."""
```
aradi, line 130: I guess, the routines below are just copy-paste versions of the one in `calcderivs`. Why not putting them into a Python module in this directory (e.g. `numchekhelpers`) and letting both programs import it?

bhourahine: Doesn't that in turn suggest requiring a proper setup.py and install (or absorbing into dp_tools)?

aradi (approved):

PR #712, opened 2021-02-10
Fix circular deps

aradi:


awvwgk (approved):
This looks good to me.

PR #713, opened 2021-02-10
Fix wrong content of serial binary eigenvector files

aradi:
Due to wrong indexing, the eigenvectors of the first k-point were written for all k-points.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,9 @@ Changed
 Fixed
 -----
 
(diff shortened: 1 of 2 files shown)


bhourahine (approved):
Probably should be a change log mention of this bugfix (unlikely to affect many users, but significant if it does).

awvwgk (approved):
Quite straight-forward fix, still should be added to the changelog before it is merged.

aradi:
Thanks, I added a comment to the Change Log.

PR #715, opened 2021-02-16
Enable surface area model in combination with COSMO

awvwgk:
This patch simply connects the SASA container with the COSMO container, hurray for a simple modular design.


prog/dftb+/lib_solvation/cosmo.F90
```diff
@@ -397,7 +437,11 @@ function getRCutoff(this) result(cutoff)
     !> Resulting cutoff
     real(dp) :: cutoff
 
-    cutoff = 0.0_dp
+    if (allocated(this%sasaCont)) then
+      cutoff = max(cutoff, this%sasaCont%getRCutoff())
+    else
+      cutoff = 0.0_dp
```
bhourahine, line 443: This should probably be before the if test, as the max branch can get an undefined value of cutoff, i.e.:
cutoff = 0.0_dp
if (allocated(allocated(this%sasaCont)) then
cutoff = max(cutoff, this%sasaCont%getRCutoff())
end if

bhourahine (approved):
Ready to merge

PR #720, opened 2021-02-21
Add missing error call

bhourahine:
Missing error call in an eigensolver routine


awvwgk (approved):
Easy fix. Ready to go.

PR #719, opened 2021-02-21
Implement index selection expression parsing

aradi:
As a side effect, a robust error passing mechanism is also introduced.

* [x] Implement index selection parser
* [x] Use parser for all relevant keywords
* [x] Document index selection syntax.
* [x] Create entensive testing

PR #721, opened 2021-02-26
Extended transQ

thomas-niehaus:
Phase I of III bringing the TD-LC branch of code back to the general. Here I generalized the getia index array to cover also occ-occ and vir-vir transitions. This allows all transition charges to be cached, which is important for the LC part. The reverse array iatrans was also extended. 
Phase II: Using cached charges everywhere in the TD part
Phase III: Bring the LC part over


prog/dftb+/lib_timedep/linrespgrad.F90
```diff
@@ -318,19 +348,19 @@ subroutine LinRespGrad_old(tSpin, this, iAtomStart, grndEigVecs, grndEigVal, scc
       ALLOCATE(symmetries(1))
       symmetries(:) = [ " " ]
     end if
-
+
```
bhourahine, line 351:
```suggestion

```

prog/dftb+/lib_timedep/linrespgrad.F90
```diff
@@ -346,8 +376,8 @@ subroutine LinRespGrad_old(tSpin, this, iAtomStart, grndEigVecs, grndEigVal, scc
     ALLOCATE(osz(this%nExc))
 
     ! Find all single particle transitions and KS energy differences for cases that go from filled
-    ! to empty states
-    call getSPExcitations(grndEigVal, filling, wij, getij)
+    ! to empty states, create index arrays for ov,oo,vv
```
bhourahine, line 379:
```suggestion
    ! to empty states, create index arrays for ov,oo,vv
```

prog/dftb+/lib_timedep/linrespgrad.F90
```diff
@@ -583,7 +593,15 @@ subroutine LinRespGrad_old(tSpin, this, iAtomStart, grndEigVecs, grndEigVal, scc
       end if
 
       ! set up transition indexing
-      call rindxov_array(win, mLUMO, nxov, getij, iatrans)
+      call rindxov_array(win, nxov, nxoo, nxvv, getia, getij, getab, iatrans)
+
+   #:block DEBUG_CODE
```
bhourahine, line 598: Good! (like this)
PR #719 added a unit test framework, so this could also migrate there eventually.

prog/dftb+/lib_timedep/rs_linresp.F90
```diff
@@ -3184,4 +3229,54 @@ subroutine getHooT(nOcc, nXov, nXoo, nOrb, homo, nAtom, nMatUp, iaTrans, getIJ,
 
   end subroutine getHooT
 
+  subroutine chargeTestRS(iAtomStart, sTimesGrndEigVecs, grndEigVecs, win, nVir, nOcc, &
```
bhourahine, line 3232: Is there an easy way to have one routine for both chargeTest and chargeTestRS ?

bhourahine (commented):
General things:
- would be good to document the i,j; a,b; A,B labelling convention somewhere
- is there a performance cost for the extra indexing with the DFTB2 case? The indexing and evaluation of transitions within the blocks are now always generated, but might be good if these can be skipped if not needed.

prog/dftb+/lib_timedep/linrespgrad.F90
thomas-niehaus: - That's something that should appear top of the main linresp routine, and will be done if the RS stuff is merged. But generally i,j is occ, ab is virtual, rs are general MO.  
- There is some overhead to store the occ-occ and vir-vir charges in case one does just a spectrum and no gradients. So yes, I should add a flag for doing just occ-virt or all.

prog/dftb+/lib_timedep/rs_linresp.F90
thomas-niehaus: That could be done, but since the test routines will go away soon and we anyway merge the LC and non-LC code (at least that is my plan) it is maybe not worth the effort !?

bhourahine: That's why I said 'easy'.

prog/dftb+/lib_timedep/linrespgrad.F90
bhourahine: Do you want to make the occ-virt flag change in this branch, or as a separate pull request afterwards?

thomas-niehaus: I will do it in this branch, no need for a seperate pull request. How does it work: I simply push my changes to the current branch and the pull request is updated with the most recent files automatically?

aradi: Yes, this is exactly, how it should be done. (Pull requests are always pinned to branches, not to specific commits)

prog/dftb+/lib_timedep/linrespcommon.F90
```diff
@@ -227,36 +227,66 @@ end subroutine rindxvv
...
     !> resulting index array from orbital pairs to compound index
-    integer, intent(out) :: iatrans(:,lumo:,:)
+    integer, intent(out) :: iatrans(:,:,:)
 
-    integer :: ia
+    integer :: ia, ij, ab
 
+    @:ASSERT(size(getia,dim=2) == 3)
     @:ASSERT(size(getij,dim=2) == 3)
+    @:ASSERT(size(getab,dim=2) == 3)
```
aradi, line 260: I think, we can leave away those asserts. (The compilers bounday checks should catch this error in debug mode automatically.)

aradi (approved):
Generally, in a future PR, I'd welcome a derived type, which contains all the indices `getia`, `getij`, `getab`, `win`, `iatrans` (and maybe also the size variables `nxov`, etc.). This could be generated at the beginning (with options to generate it only partially in case not all of those is needed) and then passed around as one (`intent(in)`) argument to all routines which have to deal with indexing.

prog/dftb+/lib_timedep/linrespgrad.F90
thomas-niehaus: ok!

PR #722, opened 2021-02-26
Decrease nr. of continuation lines in dftd3param

aradi:
Reduce their number to a safe (and standard conforming) limit, which most Fortran build tools can handle.


prog/dftb+/lib_dftb/dftd3param.F90
```diff
@@ -469,7 +472,10 @@ pure subroutine copyC6(c6ab)
       &    0.0000_dp,    0.0000_dp,    0.0000_dp,    0.0000_dp,    0.0000_dp, &
       &    0.0000_dp,    0.0000_dp,    0.0000_dp,    0.0000_dp,    0.0000_dp, &
       &    0.0000_dp,    0.0000_dp,    0.0000_dp,    0.0000_dp,    0.0000_dp, &
-      &    0.0000_dp,    0.0000_dp,    0.0000_dp,    0.0000_dp,    0.0000_dp, &
+      &    0.0000_dp,    0.0000_dp,    0.0000_dp,    0.0000_dp,    0.0000_dp],&
+      &    [maxRef, maxRef, blockSize])
+
+    c6ab(:, :, 39:57) = reshape([ &
```
bhourahine, line 478: Not sure if its clearer, but the sizing could be done explicitly with multiples of blockSize in each chunk
```suggestion
    c6ab(:, :, 2*blockSize+1:3*blockSize) = reshape([ &
```

bhourahine (commented):
Seems harmless, is there a compiler with problems on this number of continuation lines?

awvwgk (approved):
Looks good to me.

PR #723, opened 2021-02-26
Mpi issues

aradi:
* Add the possibility of launching MPI-calculation via the C-interface
* Release MPI and BLACS resources when calculator is finalised to prevent MPI-framework running out of communicator if calculator is set-up and teared-down repeatedly.


prog/dftb+/api/mm/dftbplus.h
```diff
@@ -110,6 +110,21 @@ void dftbp_api(int *major, int *minor, int *patch);
 void dftbp_init(DftbPlus *instance, const char *outputfilename);
 
 
+/**
+ * Initializes a DFTB+ calculator (MPI-version).
+ *
+ * \param[inout] instance  Handler of DFTB+ instance.
+ *
+ * \param[in] outputfilename  Name of the file, where the DFTB+ screen output should be written.
+ *     If you pass NULL, it will be written to standard output. If you pass any other file, it will
```
bhourahine, line 119:
```suggestion
 *     If you pass NULL, it will be written to standard output. If you pass any other file name as a string, it will
```

prog/dftb+/api/mm/dftbplus.h
```diff
@@ -110,6 +110,21 @@ void dftbp_api(int *major, int *minor, int *patch);
...
+ *     be open, and the file will be written there. Pass "/dev/null" to suppress output.
```
bhourahine, line 120:
```suggestion
 *     be opened, and the output will be written there. Pass "/dev/null" to suppress output.
```

prog/dftb+/api/mm/capi.F90
```diff
@@ -475,4 +486,26 @@ function fortranChar(cstring, maxlen)
...
+    integer, intent(out) :: outputUnit
+    logical, intent(out) :: tOutputOpened
+
+    character(c_char), pointer :: pOutputFileName
+    character(:), allocatable :: fortranFileName
+
+    if (c_associated(outputFileName)) then
+      call c_f_pointer(outputFileName, pOutputFileName)
+      fortranFileName = fortranChar(pOutputFileName)
+      open(newunit=outputUnit, file=fortranFileName, action="write")
```
bhourahine, line 501: Is it worth trapping the error return from this (as the filename might be garbage)?

prog/dftb+/api/mm/dftbplus.h
```diff
@@ -110,6 +110,21 @@ void dftbp_api(int *major, int *minor, int *patch);
...
+ *
+ * \param[in] mpiComm  MPI-communicator id. You must convert your MPI-communicator with the
```
bhourahine, line 122:
```suggestion
 * \param[in] mpiComm  MPI-communicator id. Before calling, you must convert your MPI-communicator with the
```

prog/dftb+/lib_common/mpienv.F90
```diff
@@ -119,6 +120,18 @@ subroutine TMpiEnv_init(this, nGroup)
   end subroutine TMpiEnv_init
 
 
+  !> Finalizes the communicators allocated here
```
bhourahine, line 123:
```suggestion
  !> Finalises the communicators in the structure supplied here
```

bhourahine (commented):
Definitely needs at least one regression tester for this.

bhourahine:
The mpifx changes have been merged.

prog/dftb+/api/mm/capi.F90
aradi: Not at the moment (currently we do not pass back any errors anway...)

bhourahine (approved):
Ready to go.
Probably needs to be later PRs to add a C tester and also error checking on file opening.

PR #724, opened 2021-02-26
Fix results.tag output

aradi:
Orbital populations are written for both spins, gross charge is calculated correctly for spin polarized case, spin polarizations are also written.


bhourahine:
We really need to regression test the contents of this file (could to this as a hack with a shell script in the test system to re-names to autotest.tag for some specific example).

awvwgk:
The tester could just enable such check in case a `_results.tag` file is found.

aradi:
Sure, we can "hack" this in into (the already quite horrible) autotest-er script. :wink: Still, let's merge this first, if you agree and make that as a separate PR.

awvwgk (approved):

bhourahine:
#731

PR #726, opened 2021-02-27
Contact file now only storing coulomb shift

bhourahine:
Minor variable re-naming and update of the contact potential to only contain the 1/r shift part (with a new contact file format version as a result).


awvwgk (commented):
What is the reasoning to only retain the second order Coulomb shift here? How is the Coulomb shift for a contact calculation different from a solvation shift or dispersion shift?

bhourahine:
@awvwgk at the moment, neither of those work with transport calculations, since the contact presently has to be a DFTB2 periodic calculation. Eventually, this will have to be revised, but the current contact format does contains the gross charges, from which both of those types corrections could be evaluated in a way consistent with the boundary conditions.

awvwgk:
I understand this point, but which of the other shift contributions, if they work with transport in the future, should be included in the restart file? Adding a new term to transport might result in a small surprise if one forgets to modify the `coulombShell` generation accordingly.

bhourahine:
@awvwgk the file format version for the contact shifts should also be increased at that point if there are extra contributions.
At the moment, the reason for the Coulomb shift is to avoid re-evaluation of the electrostatics for the contact from the stored self-consistent charges (the Coulomb re-evaluation would probably not be a significant cost, but of course needs to match the boundary conditions of the periodic contact). Other (local so far) potentials are presently re-evaluated from the charges (spin, +U).
Probably there is a discussion to be had as to what should be simply re-evaluated from the charges vs. what should be stored as potential shift(s). In principle the contact file could instead store information about the contact calculation and the resulting self-consistent charges.

awvwgk (approved):

PR #727, opened 2021-02-28
Variational ground state option

bhourahine:
Optional use of output potential in energy expression, skipping a re-evaluation. This would be relevant for expensive electrostatics (q dependent external media, self-consistent dispersion models, ...)
- [x] ground state case
- [x] regression test
- [x] document option
- [x] Intersection with Poisson solver RecomputeAfterDensity

### DIFF
--- a/test/prog/dftb+/scc/2H2O_nonvariational/dftb_in.hsd
+++ b/test/prog/dftb+/scc/2H2O_nonvariational/dftb_in.hsd
@@ -1,0 +1,44 @@
+Geometry = GenFormat {
+6 C
+ O H
+   1   2     -.499735     -.006429      .000000
+   2   1    -1.467047     -.147444      .000000
+   3   1     1.392547      .133948      .000000
+   4   2    -1.876914      .729873      .000000
+   5   2     1.737135     -.325150     -.779513
+   6   2     1.737135     -.325150      .779513
+}
+
+Driver = {}
+
+Hamiltonian = DFTB {
+  SCC = Yes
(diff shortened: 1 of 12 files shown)


test/prog/dftb+/scc/2H2O_nonvariational/dftb_in.hsd
```diff
@@ -0,0 +1,44 @@
...
+   4   2    -1.876914      .729873      .000000
+   5   2     1.737135     -.325150     -.779513
+   6   2     1.737135     -.325150      .779513
+}
+
+Driver = {}
+
+Hamiltonian = DFTB {
+  SCC = Yes
+  VariationalEnergy = No
```
aradi, line 16: This seems not to match the feature as implemented above (althoug I like this approach more as it would made the logic much easier...) and the test fails, if executed. Due to a typo in `tests` (and a bug in `autotest2`, which apparently causes non-existent tests to succeed) it was not reported as failure.

test/prog/dftb+/tests
```diff
@@ -144,6 +144,7 @@ analysis/graphene_localise             #? not WITH_MPI
 onsite/H2O                             #? MPI_PROCS <= 1
 dftb+u/CH3                             #? MPI_PROCS <= 1
 dispersion/2H2O                        #? MPI_PROCS <= 1
+dispersion/2H2O_nonvariational         #? MPI_PROCS <= 1
```
aradi, line 147: That should read `scc/2H2O_nonvariational`.

PR #728, opened 2021-02-28
Clean up and add transport compatibility checks

bhourahine:


bhourahine:
Fix for mac bug will be in a different branch

PR #729, opened 2021-03-01
Gcc 7 and 8 don't have findloc

bhourahine:
Workaround for partial F08 standard implementation.


awvwgk (approved):

bhourahine:
Looks like gcc7 is more broken than that, as the fix itself doesn't work due to an indexing problem with arrays of character(*) data. Add into the findloc hack:
do iSpecies = 1, size(selection%speciesNames)
   write(*,*)iSpecies, selection%speciesNames(iSpecies)
end do
prints
           1 h                                                 
           2 h
while for gcc9:
           1 c                                                 
           2 h

PR #732, opened 2021-03-02
Enable writing of cavity information as cosmofile

awvwgk:
Allows writing of the Cosmo cavity as cosmo file. The cosmo format uses a strange unit convention mixing Ångström with atomic units, but I tried to get all of those units in the printout as correct as possible.

#### New input

```cson
Analysis {
  WriteCosmoFile = Yes
}
```

### DIFF
--- a/doc/dftb+/manual/dftbp.tex
+++ b/doc/dftb+/manual/dftbp.tex
@@ -4174,6 +4174,7 @@ calculation and/or calculate properties.
 \begin{ptable}
   \kw{AtomResolvedEnergies} & l & & No & \\
   \kw{MullikenAnalysis} & l & & Yes & \\
+  \kw{WriteCosmoFile} & l & & No & \\
   \kw{CM5} & m & MullikenAnalysis = Yes & & \pref{sec:dftbp.CM5} \\
   \kw{WriteNetCharges} & l & MullikenAnalysis = Yes & No & \\
   \kw{ProjectStates} & m & & \cb & \\
@@ -4196,6 +4197,9 @@ calculation and/or calculate properties.
 \item[\is{WriteNetCharges}] If \is{Yes}, the net charges (component of the
   charge associated only with the on-site part of the atomic orbitals, not the
   off-site hybridisation).
+
+\item[\is{WriteCosmoFile}] If \is{Yes}, the cavitity information is written as
+  cosmo file to dftbp.cosmo. Only works if the solvation model is COSMO.
 
 \item[\is{CM5}] If present the charge model 5 (CM5)\cite{marenich2012} corrected
   atomic partial charges will be written.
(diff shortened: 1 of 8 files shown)


doc/dftb+/manual/dftbp.tex
```diff
@@ -4197,6 +4198,9 @@ \section{Analysis}
   charge associated only with the on-site part of the atomic orbitals, not the
   off-site hybridisation).
 
+\item[\is{WriteCosmoFile}] If \is{Yes}, the cavitity information is written as
+  cosmo file to dftbp.cosmo. Only works if the solvation model is COSMO.
```
bhourahine, line 4202:
```suggestion
  a cosmo file to dftbp.cosmo. Only works if the solvation model is COSMO.
```
Is there an available reference for what is in a cosmo file (I'm guessing this is for COSMOtherm/COSMO-RS or similar)?

awvwgk: Good question, we tried to reconstruct them from the Turbomole generated cosmo files. Of course this file format is far from being standardized in any way. Other implementations outside of the Turbomole package are rare, not sure if they use COSMO files:
- https://github.com/usnistgov/COSMOSAC
- https://www.scm.com/doc/COSMO-RS

bhourahine (approved):
Looks about ready to go, but any changes from that format document? Also, can you add this feature to the change log.

prog/dftb+/lib_dftbplus/mainio.F90
```diff
@@ -5294,4 +5301,33 @@ subroutine writeReksDetailedOut1(fd, nGeoSteps, iGeoStep, tMD, tDerivs, &
...
+    real(dp), intent(in) :: coords0(:,:)
+
+    !> Total energy
+    real(dp), intent(in) :: energy
+
+    integer :: unit
+
+    select type(solvation)
+    class is (TCosmo)
+      write(stdOut, '(*(a:, 1x))') "Cavity information written to", cosmoFile
```
aradi, line 5326: Cool, I am positively surprised too see, that unlimited format repetition finally works with all compilers we use!

prog/dftb+/lib_solvation/cosmo.F90
```diff
@@ -909,4 +913,131 @@ subroutine efld(nsrc, src, csrc, ntrg, ctrg, ef)
...
+    keps = 0.5_dp * (1.0_dp - 1.0_dp/this%dielectricConst)
+    dielEnergy = keps * dot_product(zeta, phi)
+
+    write(unit, '(a)') &
+        & "$info", &
+        & "prog.: dftb+"
+
+    write(unit, '(a)') &
+        & "$cosmo"
+    write(unit, '(2x, a:, "=", g0)') &
```
aradi, line 968: Why is the colon edit descriptor needed here? The I/O list has a fixed size which matches the format descriptors, so no format rewinding happens.

aradi (approved):

prog/dftb+/lib_solvation/cosmo.F90
awvwgk: cosmo files can provide a string only entry here like `cavity closed`.

PR #734, opened 2021-03-03
Code findloc() in index selection for GCC 7/8 compatibility

aradi:
Makes the findloc() fix less intrusive :wink:

### DIFF
--- a/prog/dftb+/lib_io/indexselection.F90
+++ b/prog/dftb+/lib_io/indexselection.F90
@@ -84,7 +84,8 @@ module dftbp_io_indexselection
     integer :: indexRange(2)
 
     !> Names of the species, in case it is an atom index selection. Shape: [nSpecies]
-    !> Workaround: gcc7 assignment broken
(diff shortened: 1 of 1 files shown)


awvwgk (approved):

prog/dftb+/lib_io/indexselection.F90
```diff
@@ -502,4 +492,18 @@ function getTokenType_(token) result(tokenType)
   end function getTokenType_
 
 
+  !> Replacement for findloc(speciesNames, species, dim=1) to ensure GCC 7/8 compatibility
+  function findloc_(speciesNames, speciesName) result(iSpecies)
```
bhourahine, line 496: At least make the function pure if you are [rearranging the deckchairs](https://en.wiktionary.org/wiki/rearrange_the_deck_chairs_on_the_Titanic) 
As soon as gcc7/8 pass out of common use, this should be ditched anyway.

PR #735, opened 2021-03-03
Set reshapes from blocksize

bhourahine:
Just in case the block size changes again.


awvwgk (commented):
While I certainly appreciate the effort, I want to note that there is a `writeCopyC6Code` routine in this module to automate such changes.

bhourahine:
It took a few minutes to work out the regex for emacs, so not so bad. The question is more whether to use this style of indexing (hence potentially modifying the writeCopyC6Code) or just leave it as numerical ranges.

awvwgk:
Not too sure about never touching it again. See my note in https://github.com/dftbplus/dftbplus/issues/738 on the increased compile time from 30sec to >10min with the last change on this file. Currently I'm patching the file to actually work on DFTB+.

PR #736, opened 2021-03-04
Fix usage of unallocated instance

aradi:


awvwgk (commented):
Seems to be a misconception on my side using `merge` as ternary operator. All compiler I'm aware of are actually short circuiting the evaluation of the arguments correctly to avoid accessing unallocated variables.

However, it's easier to solve the underlying problem and kick out the `parEwald` argument completely with #737 instead.

PR #737, opened 2021-03-04
Remove unused Ewald parameter from DFT-D4

awvwgk:
I think this is the simpler solution here.

Closes #736


aradi:
Perfect, that's even better.

aradi (approved):

PR #739, opened 2021-03-05
Allow selection of ideal conductor

awvwgk:
This allows to set the dielectric constant to infinity and approach the ideal conductor limit for both COSMO and GB (note that GB == ALPB for epsilon -> infinity).


test/prog/dftb+/tests
```diff
@@ -209,6 +210,7 @@ solvation/cosmo_g1                     #? MPI_PROCS <= 2
 solvation/cosmo_b33                    #? MPI_PROCS <= 2
 solvation/cosmo_i8e                    #? MPI_PROCS <= 2
 solvation/cosmo_444                    #? MPI_PROCS <= 2
+solvation/cosmo_inf                    #? MPI_PROCS <= 2
```
bhourahine, line 213: This case is too small for 2 processors

awvwgk:
```suggestion
solvation/cosmo_inf                    #? MPI_PROCS <= 1
```

bhourahine:
Unfortunately this is being hit by the problem #740 fixes (and which in turn ran into CI problems, which #741 might be resolving).

doc/dftb+/manual/dftbp.tex
```diff
@@ -3548,6 +3548,7 @@ \subsubsection{Conductor like screening model}
     solvent density in \kw{Density} \modif{\modtype{massdensity}}.
     \kw{MolecularMass} and \kw{Density} only affect the calculation if
     ``reference'' is chosen as state of the solution.
+    Setting \kw{Epsilon} to \is{Inf} results in the ideal conductor limit.
```
bhourahine, line 3551:
```suggestion
    Setting \kw{Epsilon} to equal \is{Inf} results in the ideal conductor limit.
    When \dftbp{} is built with an IEEE compliant compiler, this sets the dielectric constant to
    $\varepsilon_r = +\infty$, for other compilers it is instead set as the largest available real value.
```

bhourahine (approved):
Ready to merge (keps could have been added to the derived type, and set to 1 for the Inf case, even if not an IEEE compiler, but I guess the thinking is that this avoid the parser needing to understand anything about how COSMO actually works).

PR #740, opened 2021-03-07
Fix for rename on tContactCalc

bhourahine:
Fixes rename of variable that is currently breaking the main branch


bhourahine:
#741 includes this PR

PR #743, opened 2021-03-10
Adds container for repulsive contributions

aradi:
Hides the details, how the repulsive is generated. Allows for easier integration of other or additional repulsive contributions (e.g. ChIMES).

### DIFF
--- a/prog/dftb+/lib_dftb/CMakeLists.txt
+++ b/prog/dftb+/lib_dftb/CMakeLists.txt
@@ -54,6 +54,7 @@ set(sources-fpp
   ${curdir}/slakocont.F90
   ${curdir}/slakoeqgrid.F90
   ${curdir}/sparse2dense.F90
+  ${curdir}/splinepolyrep.F90
   ${curdir}/spin.F90
   ${curdir}/spinorbit.F90
   ${curdir}/stress.F90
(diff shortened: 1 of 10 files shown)


awvwgk (commented):
How do we create a new container based on the base class in practice? The interfaces are not particular uniform especially regarding the order of the dummy arguments, also the `updateCoords` is currently useless because it missing important quantities of the neighbour list.

A container implementing more expensive repulsive contributions might decide to calculate the energy, gradient and virial in the `updateCoords` entry point and use cached results in the getter functions, like the dispersion or solvation containers are currently doing.

aradi:
Good point. Actually, the current container should only collect (and hide) all repulsive contributions which can be active at the same time. So, it just centralizes the scattered routines/variables in one place.  The motivation is the ChIMES force-field, which in some cases may be parameterized as a correction on top of the traditional spline repulsive. Then we need two different repulsive contributions, and the current object could hide the details.

I agree, that a base class for deriving repulsive potentials would be also useful, indeed. But, that would be additional to `TRepulsive`. The actual instances could then be passed to `TRepulsive` at initialization, which then would invoke them.

awvwgk:
> Then we need two different repulsive contributions, and the current object could hide the details.

Have a look at the solvation containers which are both standalone and nestable at the same time to allow a modular construction of the solvation energy expression. The same principle can be used for repulsive contributions here.

aradi:
You mean, making it right? Already from the beginning?! :laughing: 
I've created a generic interface as a separate type now, trying to match the footsteps of `TSolvation`.

awvwgk (approved):
Thanks a lot for taking the extra mile ;). This will be helpful in the future for other repulsive potentials.

PR #744, opened 2021-03-10
Work around intel compiler problems for D3-params

### LINKED ISSUE #738: Compile time regression
**Describe the bug**

Compile time with `ifort` has significantly increased due to https://github.com/dftbplus/dftbplus/pull/722. Since the dftd3param was added in this way to remove the overhead from `ifort` compiling the old copyc6 routines I consider this a regression.

**To Reproduce**

Compile DFTB+ at c082c6770d1c1dfce8fc62fa1431a9cc489a3624 (dftd3param takes ~30 sec) and 3f52e21edb3aa9adb2f1ee612306ef4a96804a45 (dftd3param takes >10 min) with `ifort`.

**Expected behaviour**

Don't take >10 min to compile a file that could be compiled in 30 sec.

aradi:
Avoiding `reshape` seems to boost Intels performance. Also, array boundaries are formulated using the size constants.

Closes #738. Closes #735.


awvwgk (approved):
Thanks, this resolved the issue.

PR #748, opened 2021-03-15
Correct typo in various hsd file comments

bhourahine:
Clean up comments in hsd files with the old 'tollerance' typo.


awvwgk (approved):
LGTM

PR #750, opened 2021-03-18
Symmetrises the charges in contact calculations

bhourahine:
As the two principal layers are identical, their charges should
be symmetric. In the absence of code changes to just calculate
one PL and duplicate it, this instead symmetrises the charge
between equivalent atoms in the two layers.

### DIFF
--- a/prog/dftb+/lib_dftbplus/main.F90
+++ b/prog/dftb+/lib_dftbplus/main.F90
@@ -975,7 +975,8 @@ contains
                 & this%maxSccIter, this%sccTol, tStopScc, this%tMixBlockCharges, this%tReadChrg,&
                 & this%qInput, this%qInpRed, sccErrorQ, tConverged, this%dftbU, this%qBlockOut,&
                 & this%iEqBlockDftbU, this%qBlockIn, this%qiBlockOut, this%iEqBlockDftbULS,&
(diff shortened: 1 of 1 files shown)


bhourahine:
Would be obsoleted by #755

prog/dftb+/lib_dftbplus/main.F90
```diff
@@ -3623,6 +3629,27 @@ subroutine getNextInputCharges(env, pChrgMixer, qOutput, qOutRed, orb, nIneqOrb,
...
+              qBlockIn(:,:,iAt+nAtom/2,:) = qBlockIn(:,:,iAt,:)
+            end do
+          end if
+          if (allocated(qiBlockIn)) then
+            do iAt = 1, nAtom/2
+              qiBlockIn(:,:,iAt,:) = 0.5_dp*(qiBlockIn(:,:,iAt,:) + qiBlockIn(:,:,iAt+nAtom/2,:))
+              qiBlockIn(:,:,iAt+nAtom/2,:) = qiBlockIn(:,:,iAt,:)
+            end do
+          end if
+        end if
```
aradi, line 3651: You might consider to put this into a separate subroutine (e.g. `symmetrizeContactCharges()`) and call it from here.

aradi (approved):

bhourahine:
@aradi as mentioned, #755 would make this PR obsolete by using only 1 PL in the contact calculation.

PR #751, opened 2021-03-22
Charge file only containing central cell data

bhourahine:
Bump file version, but leave option to read previous file as well.


awvwgk (commented):
Is there a regression test for this change?

bhourahine:
> Is there a regression test for this change?

The existing ascii tests check the old format, while several tests generate and then re-read charge.bin files in the current format (and pass). Or are you asking for something like running a supercell and loading its charges into a transport calculation?

awvwgk (approved):

PR #752, opened 2021-03-23
Checks for use of external field and solvents

bhourahine:
As there is a missing electrostatic term for extra fields from the solvent bulk at the moment.
- [x] stop external fields for program and API if solvent modifies the electrostatic environment
- [x] stop external polarizable media calculations if solvent modifies the electrostatic environment
- [x]  print description/warnings about dipole an/or electrostatic potentials in the presence of field modifying solvents
- [x] Block time propagation being used in cases modified by the electrostatic environment of the solvent

### DIFF
--- a/prog/dftb+/api/mm/mmapi.F90
+++ b/prog/dftb+/api/mm/mmapi.F90
@@ -161,7 +161,7 @@ contains
 
     if (.not. associated(this%hsdTree)) then
       print *, 'ERROR: input has not been created yet!'
-      stop
+      error stop
     end if
     call getChild(this%hsdTree, rootTag, root)
 
@@ -231,7 +231,7 @@ contains
 
     if (geo%nSpecies /= maxval(geo%species) .or. minval(geo%species) /= 1) then
       write (*, *) "Nr. of species and nr. of specified elements do not match."
-      stop
+      error stop
     end if
 
   end subroutine TDftbPlusAtomList_addToInpData
@@ -394,7 +394,7 @@ contains
     if (allocated(this%main%solvation)) then
       if (this%main%solvation%isEFieldModified()) then
         write(stdOut, "(A)") "Error: External fields currently unsupported for this solvent model"
-        stop
+        error stop
       end if
     end if
 
@@ -423,7 +423,7 @@ contains
     if (allocated(this%main%solvation)) then
       if (this%main%solvation%isEFieldModified()) then
         write(stdOut, "(A)") "Error: External fields currently unsupported for this solvent model"
-        stop
+        error stop
       end if
     end if
 
@@ -449,7 +449,7 @@ contains
     if (allocated(this%main%solvation)) then
       if (this%main%solvation%isEFieldModified()) then
         write(stdOut, "(A)") "Error: External fields currently unsupported for this solvent model"
-        stop
+        error stop
       end if
     end if
 
@@ -582,7 +582,7 @@ contains
 
     if (.not. this%tInit) then
       write(stdOut, "(A)") "Error: Received uninitialized TDftbPlus instance"
-      stop
+      error stop
     end if
 
   end subroutine TDftbPlus_checkInit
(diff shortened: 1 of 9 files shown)


prog/dftb+/api/mm/mmapi.F90
```diff
@@ -387,6 +387,11 @@ subroutine TDftbPlus_setExternalPotential(this, atomPot, potGrad)
 
     call this%checkInit()
 
+    if (allocated(this%main%solvation)) then
+      write(stdOut, "(A)") "Error: External fields currently unsupported for solvent models"
+      stop
```
awvwgk, line 392: Should this really use `stop` here?

prog/dftb+/api/mm/mmapi.F90
```diff
@@ -409,6 +414,11 @@ subroutine TDftbPlus_setExternalCharges(this, chargeCoords, chargeQs, blurWidths
...
```
awvwgk, line 419: Same here, should this be a `stop`?

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -2100,6 +2100,13 @@ subroutine initProgramVariables(this, input, env)
           & spin orbit calculations")
     end if
 
+    if (allocated(this%solvation)) then
+      if (this%tExtChrg .or. this%tEField) then
+        write(*,*)'External fields are not currently compatible with implicit solvents.'
```
awvwgk, line 2105: There is no action associated with this write statement

awvwgk (changes requested):
I don't think we should block solvation and external fields. At least for Born models no additional terms arise here. The check should be more concrete and only block COSMO with external fields.

For solvation and external charges I don't see a good way to put them together at all. Especially because the cavity generation for an external charge is not clear. This is something we should create at least a warning for.

bhourahine:
> At least for Born models no additional terms arise here

Is that a correct assertion? For a spherical cavity inside a dielectric of eps_r and where there is an external homogeneous field E0, the electric field inside the 'bubble' is E0 * 3eps_r/(2eps_r +1) from Gauss' law. This should apply to a Born approximation equally as with a PCM, or am I missing something special about generalised Born? The external fields in the code are definitely not adapted for this.

prog/dftb+/lib_dftbplus/initprogram.F90
bhourahine: Yes, that was a stupid typo on my part...

awvwgk:
Oh right, I was only looking at the solute and cavity terms here.

aradi (approved):
It looks OK to me.

prog/dftb+/api/mm/mmapi.F90
```diff
@@ -391,6 +391,13 @@ subroutine TDftbPlus_setExternalPotential(this, atomPot, potGrad)
 
     call this%checkInit()
 
+    if (allocated(this%main%solvation)) then
+      if (this%main%solvation%isEFieldModified()) then
+        write(stdOut, "(A)") "Error: External fields currently unsupported for this solvent model"
+        stop
```
awvwgk, line 397: There is still the question if we really should stop here? I don't think it is a good idea to stop in the API and return with exit code 0. If we really have to kill the API user, can we at least return with a non-zero exit code, please?

awvwgk (changes requested):

prog/dftb+/api/mm/mmapi.F90
bhourahine: @awvwgk are you asking for just an "error stop", or should we discuss meaningful error codes?
This brings up the need to have proper error code return from the API.

awvwgk: Exactly, I'm just asking for some non-zero exit code on an error, `error stop` will do.

awvwgk (approved):

PR #753, opened 2021-03-23
DFTD4: Use OMP shared instead of OMP reduction

awvwgk:
- try to avoid OMP errors due to allocation of temps on stack
- *not* the best possible solution here, but avoids forcing uses to set `OMP_STACKSIZE` at runtime
- could lead to performance penalties due to blocking writes on shared arrays

### DIFF
--- a/prog/dftb+/lib_dftb/dispdftd4.F90
+++ b/prog/dftb+/lib_dftb/dispdftd4.F90
@@ -660,7 +660,7 @@ contains
     real(dp) :: f6, df6, f8, df8, f10, df10
 
     !$omp parallel do default(none) schedule(runtime) &
-    !$omp reduction(+:energies, gradients, sigma, dEdq, dEdcn) &
+    !$omp shared(energies, gradients, sigma, dEdq, dEdcn) &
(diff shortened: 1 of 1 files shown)


bhourahine (approved):
Seems OK to merge. Have you tried benchmarking it for locks?

prog/dftb+/lib_dftb/dispdftd4.F90
```diff
@@ -660,7 +660,7 @@ subroutine dispersionGradient(calc, iAtFirst, iAtLast, nNeighbour, neigh, &
     real(dp) :: f6, df6, f8, df8, f10, df10
 
     !$omp parallel do default(none) schedule(runtime) &
-    !$omp reduction(+:energies, gradients, sigma, dEdq, dEdcn) &
+    !$omp shared(energies, gradients, sigma, dEdq, dEdcn) &
```
aradi, line 663: I am not sure. Wouldn't we get possible race conditions, when `energies(iAt2f)` and `gradients(:, iAt2f)` is accessed below? I fear, that this would only work with a full (symmetric) neighbour list.

PR #756, opened 2021-03-27
Put keps into the ADT for cosmo

bhourahine:
Also minor comment typo

### DIFF
--- a/prog/dftb+/lib_solvation/born.F90
+++ b/prog/dftb+/lib_solvation/born.F90
@@ -1140,7 +1140,7 @@ contains
   end subroutine getBornMatrixStillCluster
 
 
-  !> compute Born matrix using Still interaction kernel
+  !> compute Born matrix using P16 interaction kernel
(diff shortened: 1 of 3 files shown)


awvwgk (changes requested):
In this case I would evaluate `keps` in the Parser rather than the initialization and drop the dielectric constant from both the input and the container as it can be easily recalculated from `keps`.

bhourahine:
Is it worth saving the couple of bytes of storage for the dielectric in the ADT in exchange for having to convert back again from keps in 2 different places?

awvwgk:
> Is it worth saving the couple of bytes of storage for the dielectric in the ADT in exchange for having to convert back again from keps in 2 different places?

The idea behind having only one was that they can't diverge by any mean and the solvation model doesn't really need the dielectric constant for anything other than calculating keps or displaying it. But let's keep both, won't hurt for now.

awvwgk (approved):

PR #758, opened 2021-04-01
Repulsive correction via ChIMES force field

aradi:
- [X] Build ChIMES (currently only as sub-module!)
- [X] Add ChIMES force-field to repulsive
- [X] Test example
- [ ] Checking/clarifying stress tensor sign convention
- [ ] Basic conflict testing based on user input (e.g. ChIMES only works with 3D periodic systems so far)
- [ ] Documentation
- [ ] Adding Fortran08 interface to the official ChIMES repository


bhourahine:
Looks like this is passing [some regressions](https://github.com/rk-lindsey/chimes_calculator/pull/3)

PR #759, opened 2021-04-01
Refactor ion dynamics in Ehrenfest code

fbonafe:
Change order of the propagation algorithm such that the ion dynamics can be done externally, passing the new coordinates and velocities at each timestep and getting the time-dependent forces from the API.


bhourahine (approved):
Ready to merge

PR #764, opened 2021-04-06
Charge-dependent dispersion with reks

inseonglee:
- [x] : Allow charge-dependent dispersions (MBD/TS, dftd4) with REKS method.
    - [x] : Move position of current dispersions to getHamiltonianLandEnergyL routine.
    - [x] : Modify getMullikenPopulation routine to adjust attribute of qNetAtom.
- [x] : Block use of MBD with REKS.
- [x] : Add energy%Eavg to avoid confusions.
- [x] : Add test files to be able to check dispersion with REKS.

Except adding test, fixing some REKS routines is finished. Please see the changes.
I will add some test files including dftd4 or MBD dispersions, but this may take a few days.


prog/dftb+/lib_reks/reksen.F90
```diff
@@ -144,13 +144,18 @@ subroutine calcSaReksEnergy(this, energy)
...
-    energy%Etotal = energy%Eelec + energy%Erep + energy%eDisp
+    energy%Etotal = energy%Eelec + energy%Erep + energy%Edisp
 
     ! Compute the total energy for SA-REKS states
     do ist = 1, this%nstates
       this%energy(ist) = sum(this%weightL(ist,:)*this%enLtot(:))
     end do
 
+!    if (abs(energy%Etotal - this%energy(this%rstate)) >= epsilon(1.0_dp)) then
+    if (abs(energy%Etotal - this%energy(this%rstate)) >= 1.0e-8_dp) then
```
inseonglee, line 155: When I use epsilon(1.0_dp), this part shows an error. But Etotal and energy(rstate) show identical energy. I do not know why this condition satisfies. Is there any good option for tolerance? Currently, I use 1.0E-8.

prog/dftb+/lib_dftbplus/main.F90
```diff
@@ -1063,7 +1063,7 @@ subroutine processGeometry(this, env, iGeoStep, iLatGeoStep, tWriteRestart, tSto
             & this%extPressure, this%cellVol, this%dftbEnergy(1)%TS, this%tAtomicEnergy,&
             & this%dispersion, this%tPeriodic, this%tSccCalc, this%invLatVec, this%kPoint,&
             & this%iAtInCentralRegion, this%electronicSolver, this%reks,&
-            & allocated(this%thirdOrd), this%isRangeSep)
+            & allocated(this%thirdOrd), this%isRangeSep, qNetAtom=this%qNetAtom)
```
inseonglee, line 1066: @bhourahine In this commit, I add printing of qNetAtom to detailed.out file. In this block (1041 ~ 1070), why was this part added? Is both blocks (~1060 and (~790 or ~1020)) needed?

bhourahine: @inseonglee if I understand the question correctly, this is a consequence of updating the self-consistent dispersion during the SCC loop, but not updating it for non-self-consistent dispersion until after the loop terminates (this leads to a more complicated logic to print the detailed.out data but is required as polarizability can only be evaluated from the output charges after at least one diagonalization).
Probably after the next release, this will be refactored to only print total energies after the loop terminates, in part to also refactor the electrostatic potential evaluation to only happen once in each cycle of the loop.This means the stationary, not the variational, SCC-DFTB energy would be available during the SCC loop.
The disadvantage of these changes is that in an aborted calculation, there will be no indication of the energy. It would only be once the SCC loop correctly terminates that all components of the total energy are available. The advantages are a somewhat faster calculation and a simpler code structure.

inseonglee:
@bhourahine @aradi Now, the works are finished, so this PR is ready for review.

Notice that non-self-consistent MBD cannot be calculated with REKS, thus I prevent this case. Since REKS needs relaxed density to obtain correct charges for target state after SCC iterations, thus correct contribution for MBD-type dispersion cannot be treated. After implementing self-consistent MBD/TS dispersions, then the commit 691bf53 will be changed.

Currently, self-consistent dftd4 with REKS is tested, please see commit 18475a5.

prog/dftb+/lib_reks/reksen.F90
bhourahine: Could this be noise in the dispersion energy? Some of the more complex dispersion models have more numerical components to their evaluation and so will depend on several tolerances being convereged.

bhourahine:
@inseonglee yes, the API of the MBD library does not yet expose all of the components needed to evaluate that model self-consistently. It can only be used with output charges from the SCC loop.

prog/dftb+/lib_reks/reksen.F90
inseonglee: Then, how to handle this error? The reason I add this error is just to check the total energy and energy contributions. This part is not mandatory. Do I remove this error?

prog/dftb+/lib_dftbplus/main.F90
inseonglee: OK, I see. Then, this part will be refactored after next release. I will not change this part.

inseonglee:
Additionally, should I add this work to CHANGELOG? If it is added, then which section (added, changed, fixed) is proper to this work?

bhourahine:
> Additionally, should I add this work to CHANGELOG? If it is added, then which section (added, changed, fixed) is proper to this work?

As this is similar to the log comment about self-consistent D4, I suggest in the changed section

prog/dftb+/lib_reks/reksen.F90
bhourahine: If the tolerance of 1E-8 is working, this can be refactored later if there is a change in the accuracies of the dispersion models. Are you seeing this problem for molecular or periodic geometries, as some of the dispersion methods sum the periodic case in real space, but should possibly be changed to used Ewald summation.

inseonglee: So far I didn't test the periodic systems. For molecular geometries, tolerance of 1E-8 is working, and I have never seen this error. Then, if you agree, I will remove this line since this part is not tested for periodic systems.

bhourahine (commented):
I'm waiting on various compiler combination tests to complete, but other than the variable name, everything seems about ready to merge.

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -459,6 +459,9 @@ module dftbp_initprogram
     !> Do we need to show Mulliken charges?
     logical :: tPrintMulliken
 
+    !> Logical to determine whether to calculate net charge per atom (qNetAtom)
+    logical :: tAllocate
```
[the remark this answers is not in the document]
inseonglee: OK, I will use 'is' than 't'. In 3512 line, the logical 'tNetAtomCharges' already exists to decide printing of qNetAtom. How about 'isDispMbd'? Actually, qNetAtom is used when the user want to use MBD (isDispMbd) or to print net atom charge (tNetAtomCharges) to detailed.out.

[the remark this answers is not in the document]
bhourahine: isQNetAllocated might still be a good use, as in principle this can be true even if MBD is not used.
It looks like the intel 19 compiler, when compiled with debug enabled is failing regression tests for dftb+_reks/CNH4+_dftd4* can you test that combination, or should I look at this?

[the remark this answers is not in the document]
inseonglee: I see, I will change to 'isQNetAllocated'.
For regression tests, I will test them.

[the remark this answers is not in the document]
inseonglee: Currently, with intel 2021.1 compilers, the compilation is failed with debug enabled showing following sentence.

"catastrophic error: **Internal compiler error: internal abort** Please report this error along with the circumstances in which it occurred in a Software Problem Report.  Note: File and line given may not be explicit cause of this error."

This error occurs in './build-debug/_deps/mbd-src/src/mbd_methods.F90'.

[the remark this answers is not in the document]
inseonglee: Now, the regression test will be passed, please test once again with fixed code. And please see commit 19d79ab.
The reason why the error occurs are explained in the commit message. With my intel 21 compiler, all tests pass.

PR #766, opened 2021-04-08
Change to call error() in api

bhourahine:
Instead of error stop.


awvwgk (approved):
Much simpler this way.

PR #773, opened 2021-04-22
Workarounds for the NVIDIA/PGI compiler

aradi:
Some fixes to enable compilation with PGI/NVIDIA compiler.

Status:
* Serial builds (both Debug and RelWithDebInfo) work and all tests pass with the following build configure options: `-DWITH_TRANSPORT=1 -DWITH_SOCKETS=1 -DWITH_UNITTESTS=1`. The API (`-DWITH_API=1`) only works partially, as a copy operation in the q-dependant external potential generator causes segfault (corresponding tests fail)...

* Parallel compilation could not be tested so far.

* Tested with nvhpc-21.3 on dpdevel.


sys/nvidia.cmake
```diff
@@ -0,0 +1,96 @@
...
+#
+# Notes:
+#
+#  * CAVE PGI: The NVIDIA/PGI Fortran compiler has a long-standing tradition of being horribly
+#    broken (as of 2021). Usually, there are always several (valid) Fortran constructs it can not
+#    cope with, and also those, which you could get compiled yesterday, may be broken with their
+#    actual compiler of today. Therefore, be prepared for false compiler errors, internal compiler
+#    errors and segfaulting executables. But, there is of course always the chance, that this day
+#    happens to be your PGI-lucky day. Challenge your luck! Start the build, cross your fingers,
+#    and hope for the best...
```
awvwgk, line 14: Well, it is hard to disagree with this statement.

awvwgk (approved):

aradi:
Since the NVidia compiled binary has still segfault issues, the branch should not be merged for the moment. It will be rebased time-to-time to keep up with the main branch. It will be only merged, if a stable NVidia compiled binary can be obtained with it.

Issue #774, opened 2021-04-24
Processor grid too big


Martinrainer:

bhourahine:
Hello Martin,

that run time error occurs for cases where the system does not contain
sufficient atomic orbitals for the parallel eigensolver. The default
value for the BLACS BlockSize is set as 32 (see the manual section about
Parallel{} if you want to change this), corresponding to 8 sp basis
atoms. If more than one processor is requested for this small system,
then the error is triggered, as the other processors would be idle.

If you have spin polarization or k-points, there are options to increase
the parallelism for small systems (see the Groups keyword in the
parallel section of the manual).

Regards

Ben

On 24/04/2021 10:02, Martinrainer wrote:
> CAUTION: This email originated outside the University. Check before
> clicking links or attachments.
>
>
>     Hi
>     After building DFTB+ with Intel MPI, MKL, and running the code
>     with mpirun -np 2 dftb+ I receive the following error:
>     Processed input in HSD format written to 'dftb_pin.hsd'
>     Starting initialization...
>
> ERROR!
> -> Processor grid (1 x 2) too big (> 1 x 1)
> Abort(1) on node 0 (rank 0 in comm 0): application called
> MPI_Abort(MPI_COMM_WORLD, 1) - process 0
> Abort(1) on node 1 (rank 1 in comm 0): application called
> MPI_Abort(MPI_COMM_WORLD, 1) - process 1
>
> I guess the code has not been built in parallel, do you have any idea
> how to solve this problem?
>
> *To Reproduce*
> config.cmake :
> WITH_MPI, WITH_TRANSPORT, WITH_API, WITH_PYTHON, TOOLCHAIN "intel"
> intel.cmake:
> set(LAPACK_LIBRARY "mkl_intel_ilp64;mkl_sequential;mkl_core"
> ,set(SCALAPACK_LIBRARY "mkl_scalapack_ilp64;mkl_blacs_intelmpi_ilp64"
> FXX=mpif90 FC=mpiifort CC=icc
>
> No error is produced during build with cmake. The shared object (so)
> MKL libraries are used by default during link. The build fails if
> static libraries are used.
>
> Thanks
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
[...]

PR #775, opened 2021-04-24
Propagate orbital grid size error up

bhourahine:
Also print slightly more information about what the error means (response to #774 )


aradi (approved):
LGTM

PR #777, opened 2021-04-25
Print gradients for SA-REKS states

inseonglee:
These gradients can be used as diabatic properties.

### DIFF
--- a/prog/dftb+/lib_reks/reksio.F90
+++ b/prog/dftb+/lib_reks/reksio.F90
@@ -180,18 +180,20 @@ module dftbp_reksio
           end if
         end do
 
-!        write(stdOut,*) "AVG state"
-!        write(stdOut,'(3(f15.8))') this%avgGrad(:,:)
-!        write(stdOut,'(3(f15.8))')
-!        do ist = 1, this%nstates
-!          write(stdOut,*) ist, "st state (SA-REKS)"
-!          write(stdOut,'(3(f15.8))') this%SAgrad(:,:,ist)
-!          if (ist == this%nstates) then
-!            write(stdOut,"(A)") repeat("-", 50)
-!          else
-!            write(stdOut,'(3(f15.8))')
-!          end if
-!        end do
+        if (this%Plevel >= 2) then
+          write(stdOut,*) "AVG state"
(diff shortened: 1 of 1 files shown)


prog/dftb+/lib_reks/reksio.F90
```diff
@@ -180,18 +180,20 @@ subroutine printReksGradInfo(this, derivs)
...
-!          write(stdOut,*) ist, "st state (SA-REKS)"
-!          write(stdOut,'(3(f15.8))') this%SAgrad(:,:,ist)
-!          if (ist == this%nstates) then
-!            write(stdOut,"(A)") repeat("-", 50)
-!          else
-!            write(stdOut,'(3(f15.8))')
-!          end if
-!        end do
+        if (this%Plevel >= 2) then
+          write(stdOut,"(12X,A)") "AVG state"
```
bhourahine, line 184: Any reason not to write AVG in full instead of as an acronym?

prog/dftb+/lib_reks/reksio.F90
```diff
@@ -180,18 +180,20 @@ subroutine printReksGradInfo(this, derivs)
...
+          write(stdOut,'(3(f15.8))') this%avgGrad(:,:)
+          write(stdOut,'(3(f15.8))')
+          do ist = 1, this%nstates
+            write(stdOut,*) ist, "st state (SA-REKS)"
```
bhourahine, line 188: Might be better to make an array of term names, something like
character(3), parameter :: ordinals(6) = ['1st', '2nd', '3rd', '4th', '5th', '6th']
and
write(stdOut, *)ordinals(ist) // " state (SA-REKS)"

bhourahine (approved):
Seems ready to merge, but see some cosmetic comments.

prog/dftb+/lib_reks/reksio.F90
inseonglee: Done.

prog/dftb+/lib_reks/reksio.F90
inseonglee: Done.

PR #778, opened 2021-04-26
Fixed a bug in the interface between dftbplus and i-pi

BingqingCheng:
Transpose the cell matrix at the interface.


bhourahine:
Thank you for the patch. Is it possible to remove the windows line endings (which is why all lines of the file are listed as changed)?

prog/dftb+/lib_io/ipisocket.F90
```diff
@@ -1,373 +1,373 @@
...
+    call writebuffer(this%socket, natom)
+    call this%logger%write('ipisocket%send: write to socket: natom', 3)
+
+    ! transmit forces
+    call writebuffer(this%socket, reshape(forces, [3 * natom]))
+    call this%logger%write('ipisocket%send: write to socket: forces', 3)
+    call this%logger%write(forces, 5, '(f12.6)')
+
+    ! transmit stress
+    call writebuffer(this%socket, reshape(stress, [9]))
```
bhourahine, line 346: To be consistent with the transpose of the returned lattice vectors, should this not also be transposed?

BingqingCheng: I have indeed checked the stress tensor, and it was okay (without transposing): in i-pi I printed out "stress_md" that has the format [xx, yy, zz, xy, xz, yz], and it was consistent with the "Total stress tensor" printed out in detailed.out by dftb+.
To comfirm, the "Total stress tensor" from dftb+ follows the format
xx xy xz
yx yy yz
zx zy zz?

bhourahine: The tensor is symmetric, so is probably hiding this. Its probably a good idea to either transpose (even though it is 'harmless'), or add a comment in line 345 that this is symmetric, so does not need transposition.

Looking at the (i-PI driver example)[https://github.com/i-pi/i-pi/blob/fae06f697c30436a38591320acffaf8213e85505/drivers/f90/driver.f90#L679] it looks like this _might_ be expecting the virial, not the stress tensor. Do you know anything about this?

BingqingCheng: The stress tensor is indeed symmetric so it doesn't matter. I'll add a comment to note this.

I think the "virial" in i-pi refers to the fact that it's only the virial component of the stress tensor (excl the kinetic part). If there's an opposite sign because of the different defination of the internal/external pressure, this should show up when I printed out the "stress_md" using i-pi.

prog/dftb+/lib_io/ipisocket.F90
```diff
@@ -1,6 +1,6 @@
 !--------------------------------------------------------------------------------------------------!
 !  DFTB+: general package for performing fast atomistic simulations                                !
-!  Copyright (C) 2006 - 2020  DFTB+ developers group                                               !
+!  Copyright (C) 2006 - 2019  DFTB+ developers group                                               !
```
bhourahine, line 3:
```suggestion
!  Copyright (C) 2006 - 2020  DFTB+ developers group                                               !
```

prog/dftb+/lib_io/ipisocket.F90
```diff
@@ -233,7 +233,7 @@ subroutine receive(this, coord, cell, tStop)
 
     ! lattice vector data
     call readbuffer(this%socket, commsBuffer2)
-    cell(:,:) = reshape(commsBuffer2, [3, 3])
+    cell(:,:) = transpose(reshape(commsBuffer2, [3, 3])) ! eae32
```
bhourahine, line 236: What does the comment denote? As this is versioned, the author information is available in the log.

BingqingCheng: When I was going through the very painful debugging process, Edgar Engel gave some help. 

Okay I'll remove the comment.

bhourahine (approved):
Ready to merge

PR #779, opened 2021-04-26
Block dftd4 gradient with reks

inseonglee:
From previous PR, I forgot to block calculation of gradient for DFTD4 with REKS.

### DIFF
--- a/prog/dftb+/lib_dftbplus/initprogram.F90
+++ b/prog/dftb+/lib_dftbplus/initprogram.F90
@@ -2021,7 +2021,7 @@ contains
         allocate(dftd4)
         if (allocated(this%reks)) then
           if (input%ctrl%dispInp%dftd4%selfConsistent .and. this%tForces) then
-            call error("Calculation of selfconsistent dftd4 is not compatible with force calculation in REKS")
+            call error("Calculation of self-consistent dftd4 is not currently compatible with force calculation in REKS")
           end if
         end if
         if (this%tPeriodic) then
@@ -2037,7 +2037,7 @@ contains
         else if (allocated(this%reks)) then
           call error("Selfconsistent MBD/TS dispersion is blocked from REKS")
           if (this%tForces) then
-            call error("Calculation of selfconsistent MBD/TS is not compatible with force calculation in REKS")
+            call error("Calculation of self-consistent MBD/TS is not currently compatible with force calculation in REKS")
           end if
         end if
         allocate (mbd)
--- a/prog/dftb+/lib_dftbplus/initprogram.F90
+++ b/prog/dftb+/lib_dftbplus/initprogram.F90
@@ -2019,6 +2019,11 @@ contains
         call move_alloc(sdftd3, this%dispersion)
       else if (allocated(input%ctrl%dispInp%dftd4)) then
         allocate(dftd4)
+        if (allocated(this%reks)) then
+          if (input%ctrl%dispInp%dftd4%selfConsistent .and. this%tForces) then
+            call error("Calculation of selfconsistent dftd4 is not compatible with force calculation in REKS")
+          end if
+        end if
         if (this%tPeriodic) then
           call init(dftd4, input%ctrl%dispInp%dftd4, this%nAtom, this%speciesName, this%latVec)
         else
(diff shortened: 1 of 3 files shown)


prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -2019,6 +2019,11 @@ subroutine initProgramVariables(this, input, env)
         call move_alloc(sdftd3, this%dispersion)
       else if (allocated(input%ctrl%dispInp%dftd4)) then
         allocate(dftd4)
+        if (allocated(this%reks)) then
+          if (input%ctrl%dispInp%dftd4%selfConsistent .and. this%tForces) then
+            call error("Calculation of selfconsistent dftd4 is not compatible with force calculation in REKS")
```
bhourahine, line 2024:
```suggestion
            call error("Calculation of self-consistent dftd4 is not currently compatible with force calculation in REKS")
```

prog/dftb+/lib_dftbplus/initprogram.F90
```diff
@@ -2029,6 +2034,11 @@ subroutine initProgramVariables(this, input, env)
       else if (allocated(input%ctrl%dispInp%mbd)) then
         if (this%isLinResp) then
           call error("MBD model not currently supported for Casida linear response")
+        else if (allocated(this%reks)) then
+          call error("Selfconsistent MBD/TS dispersion is blocked from REKS")
+          if (this%tForces) then
+            call error("Calculation of selfconsistent MBD/TS is not compatible with force calculation in REKS")
```
bhourahine, line 2040:
```suggestion
            call error("Calculation of self-consistent MBD/TS is not currently compatible with force calculation in REKS")
```

bhourahine (approved):
Ready to merge, but see the suggestions. The forces are more an issue for MBD/TS (which is missing a term if used self-consistently) than DFT-D4 (which is implemented but seems to have more noise than would normally be expected).

prog/dftb+/lib_dftbplus/initprogram.F90
inseonglee: I changed the sentence based on your suggestions.

awvwgk:
What is the reason to block DFT-D4 in combination with REKS?

inseonglee:
> What is the reason to block DFT-D4 in combination with REKS forces?

Exactly, self-consistent DFT-D4 should be blocked with REKS forces. To calculate REKS forces, we have to solve CP-REKS equations and the orbital gradient is needed to evaluate this term. Thus, to compute correct contribution, additional routines should be implemented such as calculation of H-XC kernel or T-derivatives.

awvwgk:
> To calculate REKS forces, we have to solve CP-REKS equations and the orbital gradient is needed to evaluate this term

This shouldn't be an issue with the self-consistent DFT-D4 implementation since the it enters the SCF / REKS equations in a fully self-consistent way, unless I'm missing something that is different in REKS than in the usual SCF.

prog/dftb+/lib_dftbplus/initprogram.F90
awvwgk: This is a bit confusing, because we currently don't have a self-consistent MBD/TS implementation.

inseonglee:
> > To calculate REKS forces, we have to solve CP-REKS equations and the orbital gradient is needed to evaluate this term
> 
> This shouldn't be an issue with the self-consistent DFT-D4 implementation since the it enters the SCF / REKS equations in a fully self-consistent way, unless I'm missing something that is different in REKS than in the usual SCF.

In REKS paper (https://pubs.acs.org/doi/10.1021/acs.jctc.9b00132), charge-dependent contributions should be evaluated in gradients, please see eqs (14) and (26) in the paper. Due to this point, direct derivation for such contributions are required as in SCC, spin, and LC terms. This part is currently implemented in reksgrad.F90 file, you can find the routines in 600 and 2300 lines.
In this file, several contributions are calculated to obtain correct gradients.
Thus, this is an issue of REKS, not DFT-D4. Moreover, if a new contribution shows charge-dependency, then the charge-gradient (coupled perturbed equations) should be obtained.

prog/dftb+/lib_dftbplus/initprogram.F90
inseonglee: Then, which sentence do you prefer? How about "MBD/TS model is not currently compatible with force calculation in REKS"?

awvwgk:
Thanks for the clarification, than solvation should probably blocked with REKS gradients as well (not sure if this is already done).

inseonglee:
> Thanks for the clarification, than solvation should probably blocked with REKS gradients as well (not sure if this is already done).

Yes, you are correct, and you can check that part in 5285 line in initprogram.F90.

awvwgk (approved):

PR #776, opened 2021-04-24
Move perturb resp

bhourahine:
Shifts routines into their own module and moves derivative driver into a separate routine from the perturbation potential.

PR #782, opened 2021-05-10
Response kernel

bhourahine:
Contains #776 
- [x] move perturbation routines into own module
- [x] k-points, helical boundary conditions
- [x] RPA option for response kernel
- [ ] external on-site potentials
- [x] external atomic potential
- [ ] test  cases for all combinations of hamilonian with/without metallic filling
- [ ] rationalise output of response kernel and frontier orbital derivatives
- [ ] spin polarised analogues for Fukui output


bhourahine:
Most of this is covered by #782

PR #783, opened 2021-05-12
Add config file for stale bot

aradi:


.github/stale.yml
```diff
@@ -0,0 +1,62 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 180
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 14
```
bhourahine, line 8: Seems a little short, based on the stale test being ~6 months, how about 28 days (~1 month)?

aradi: Makes absolutely sense, indeed.

PR #788, opened 2021-05-31
Increase size of long strings

aradi:
As `character(lc)` is often used as container for file paths, its size should be big enough to contain really long path names.


bhourahine (approved):
Fine to merge, but probably adding a helper to do non-advancing reads and return an allocated  string would be  a better long term [solution](https://comp.lang.fortran.narkive.com/Fbavc1zc/limitations-of-dynamic-strings)

aradi:
You're absolutely right. Unfortunately, our parser has currently a line limit of 1024 anyway, so that would have to be changed first.

PR #784, opened 2021-05-13
Module names2

helmutwecke:
I changed the module names on all the necesary files. Now I am going to make the buildbot tests to assure everything is fine. @aradi

PR #789, opened 2021-06-01
Omit the explicit import of dftbp_assert

aradi:


prog/dftb+/include/common.fypp
```diff
@@ -40,9 +40,12 @@
 #! Check a condition if WITH_ASSERT is True and call assertError if condition is False.
 #:def ASSERT(cond)
   #:if WITH_ASSERT
-  if (.not. (${cond}$)) then
-    call assertError("${_FILE_}$", ${_LINE_}$)
-  end if
+  block
+    use dftbp_assert, only : assertError
+    if (.not. (${cond}$)) then
+      call assertError("${_FILE_}$", ${_LINE_}$)
```
awvwgk, line 46: While are on it, can we pass the actual condition tested as string to the assertion error as well?

awvwgk (approved):
Looks good to me. Only one completely unrelated suggestion to the `assertError` routine.

Merging this one will probably require rebasing for most feature branches.

prog/dftb+/include/common.fypp
aradi: Yes, that would make sense, indeed. I'll add it in a separate PR.

aradi:
Solved in #784 .

PR #790, opened 2021-06-01
Fix the running out of mpi-communicators bug

aradi:
Even a single little can make much. Hopefully, this fixes the running out of mpi-communicators bug when DFTB+ is called via the API.


bhourahine (approved):
GTM

aradi:
Thx! To be on the safe side, I will wait for confirmation from the original reporter and merge it then.

bhourahine:
@aradi after 3 weeks, worth just merging it?

bhourahine:
#800 is this fix on top of the new source tree structure.

PR #792, opened 2021-06-07
Support xTB calculations via external library

awvwgk:
- allow to conditionally compile with tblite library
- integrate wrapper data type in input data and parser
- start initialization of DFTB data structures from library
- integrate tblite interface handler with SCC procedure
- implement gradients
- add tests for GFN1-xTB and IPEA1-xTB
- documentation and changelog adjusted

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -231,7 +231,7 @@ if(WITH_TBLITE)
   #list(APPEND PKG_CONFIG_REQUIRES dftd4)
 
   set(TBLITE_GIT_REPOSITORY "https://github.com/awvwgk/tblite.git")
-  set(TBLITE_GIT_TAG "4c80d8b5ef39f6debf82938cdcb6e323743c0788")  # do not change manually!
+  set(TBLITE_GIT_TAG "778f1c1623f52eca752f4fdeb1cbcb641e33ec76")  # do not change manually!
   dftbp_config_hybrid_dependency(tblite tblite::tblite "${HYBRID_CONFIG_METHODS}" "QUIET"
     external/tblite "${exclude}" "${TBLITE_GIT_REPOSITORY}" "${TBLITE_GIT_TAG}")
   #list(APPEND PKG_CONFIG_REQUIRES tblite)
--- a/prog/dftb+/lib_dftbplus/main.F90
+++ b/prog/dftb+/lib_dftbplus/main.F90
@@ -5845,6 +5845,7 @@ contains
     logical :: tImHam
 
     tImHam = allocated(iRhoPrim)
+    totalStress(:, :) = 0.0_dp
 
     if (allocated(sccCalc)) then
       if (tImHam) then
--- a/prog/dftb+/lib_extlibs/tblite.F90
+++ b/prog/dftb+/lib_extlibs/tblite.F90
@@ -133,6 +133,9 @@ module dftbp_tblite
     !> Repulsion energy
     real(dp) :: erep
 
+    !> Halogen bonding energy
+    real(dp) :: ehal
+
     !> Non-self consistent dispersion energy
     real(dp) :: edisp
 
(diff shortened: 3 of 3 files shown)


aradi (commented):
On the programming/interfacing side, I am very happy with it as it is. There are currently a few building issues, though:

* Works only with CMake 3.19 or above. (Due to `BLAS::BLAS` target, that's a problem on DFTB+'s side, I can fix it (hopefully)).
* Makes ICE with gfortran 10 and below due to multiple variable names in the OMP reduction clause (e.g. in dtd3's zero and mzero files). As OMP-is hardwired in those libraries CMake, passing `-DWITH_OMP=False` when configuring DFTB+ does not help.
* NAG does not implement the F2018 feature of dealing with optionals in `bind(c)` routines. (As in dftd3/api.f90).

We should discuss, which of those we want to try to fix and which of those we can leave with...

aradi:
I've added a fix in this [PR](https://github.com/awvwgk/dftbplus/pull/210), you may might merge it into your branch, if you wish.

awvwgk:
> Works only with CMake 3.19 or above. (Due to `BLAS::BLAS` target, that's a problem on DFTB+'s side, I can fix it (hopefully)).

Means I might have to increase the required CMake version in my projects a bit *or* add a lot of boilerplate to backport this target to older CMake versions myself. Hard decisions...



> Makes ICE with gfortran 10 and below due to multiple variable names in the OMP reduction clause (e.g. in dtd3's zero and mzero files)

Fixed upstream, with https://github.com/awvwgk/simple-dftd3/commit/a6475bae522252078da46cc238429e124a037635 and https://github.com/dftd4/dftd4/commit/776f8af9b975c7bf7adcf01633d4400fd2298a48. 

> As OMP-is hardwired in those libraries CMake, passing `-DWITH_OMP=False` when configuring DFTB+ does not help.

OpenMP isn't hardwired, it's just using a different name for the option. I transferred the `WITH_OMP` value to the specific option for the subprojects now.


> NAG does not implement the F2018 feature of dealing with optionals in `bind(c)` routines. (As in dftd3/api.f90).

Didn't realized it was Fortran 2018, but okay. C-API is optional now in those projects and will always be disabled when using them as subprojects in DFTB+ to avoid reexporting the their C-API symbols by accident.

---

This leaves the `BLAS::BLAS` issue to be addressed.

aradi:
> > Works only with CMake 3.19 or above. (Due to `BLAS::BLAS` target, that's a problem on DFTB+'s side, I can fix it (hopefully)).
> 
> Means I might have to increase the required CMake version in my projects a bit _or_ add a lot of boilerplate to backport this target to older CMake versions myself. Hard decisions...

No, leave the CMake required versions for the moment. I'll try to address it within our custom LAPACK/BLAS finders...

aradi (commented):
The newest build of the NAG compiler is now able to deal with all dependencies, provided dtd4 contains the [include fix](https://github.com/dftd4/dftd4/pull/110).

aradi:
I get `NaN` for the stress in the `gfn1_succinic` example when using the NAG compiler in Debug mode `-DCMAKE_BUILD_TYPE=Debug -DWITH_OMP=0 -DWITH_TBLITE=1`. As it autoinitializes all floats to NaN, I guess, it is an uninitialized float somewhere...

awvwgk:
> As it autoinitializes all floats to NaN, I guess, it is an uninitialized float somewhere...

Sorry, I missed that `getStress` in `lib_dftbplus/main.F90` is returning the stress tensor as `intent(out)` but accumulating it via `getStress(tmpStress); totalStress = totalStress + tmpStress` for each call except for the `slakoCont` contributions which are directly written to `totalStress` instead.

aradi:
Thanks, seems to pass the NAG-test now :wink: . I'll merge it to the main branch.

aradi (approved):

PR #798, opened 2021-07-01
Add integral container to handle overlap related integrals

awvwgk:
Create integral container to store overlap related integrals

### DIFF
--- a/src/dftbp/type/integral.F90
+++ b/src/dftbp/type/integral.F90
@@ -1,0 +1,25 @@
+!--------------------------------------------------------------------------------------------------!
+!  DFTB+: general package for performing fast atomistic simulations                                !
+!  Copyright (C) 2006 - 2021  DFTB+ developers group                                               !
+!                                                                                                  !
+!  See the LICENSE file for terms of usage and distribution.                                       !
+!--------------------------------------------------------------------------------------------------!
+
+!> Data types to handle overlap related integrals
+module dftbp_type_integral
+  use dftbp_common_accuracy, only : dp
+  implicit none
+
+  private
+  public :: TIntegral
+
+  !> Container to store overlap related integrals
+  type :: TIntegral
+
+    !> Overlap integrals in atomic block sparse form
+    real(dp), allocatable :: overlap(:)
+
+  end type TIntegral
+
+contains
+end module dftbp_type_integral
--- a/src/dftbp/type/integral.F90
+++ b/src/dftbp/type/integral.F90
@@ -19,6 +19,12 @@ module dftbp_type_integral
     !> Overlap integrals in atomic block sparse form
     real(dp), allocatable :: overlap(:)
 
+    !> Real Hamiltonian integrals in atomic block sparse form
+    real(dp), allocatable :: hamiltonian(:, :)
+
+    !> Imaginary Hamiltonian integrals in atomic block sparse form
+    real(dp), allocatable :: iHamiltonian(:, :)
+
   end type TIntegral
 
 contains
--- a/src/dftbp/type/integral.F90
+++ b/src/dftbp/type/integral.F90
@@ -11,7 +11,8 @@ module dftbp_type_integral
   implicit none
 
   private
-  public :: TIntegral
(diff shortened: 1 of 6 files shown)


bhourahine:
@awvwgk we should probably discuss (but not necessarily make any changes at the moment) the case where there can be a complex basis, so the sparse overlap also has an imaginary component. This would happen for example with gauge including basis functions (magnetic fields or magnetic property evaluation).

awvwgk:
Introducing the container will allow to add complex overlap integrals easily without requiring major changes. I already thought about just adding an `iOverlap` as well, but we can still do so later. I will make a separate PR based on this branch, which will add the multipole integrals to the container, but I want to keep this refactor focused on just introducing the container type for now.

bhourahine:
@awvwgk its more that the integral build would produce real overlaps, which then get processed to become complex. So should we a) store both the original real and the Re + Im transformed complex overlap, b) over-write the real part with the transformed real part and store the imaginary separately, or c) do the transformation dynamically (requires carrying around boundary conditions and vector potentials into sparse/dense conversion routines).

awvwgk:
Overwriting the original overlap with the real part sounds most straight-forward. I don't think we need the original overlap again in such a scenario. It probably worth to consider bundling together the ~~partial charges~~ populations to make the main code more agnostic about the quantities needed before introducing an imaginary overlap component.

src/dftbp/type/integral.F90
```diff
@@ -0,0 +1,61 @@
...
+module dftbp_type_integral
+  use dftbp_common_accuracy, only : dp
+  implicit none
+
+  private
+  public :: TIntegral, TIntegral_init
+
+
+  !> Container to store overlap related integrals
+  type :: TIntegral
```
bhourahine, line 18: For multi-determinantal cases (Delta SCF, REKS, CI, ...), the overlap would be the same for different instances of the hamiltonian. Presumably OK if they are processed sequentially in serial, as uses only one copy of this structure (and parallel doesn't care too much). Suggests either a) live with the duplication b) pointers :( or c) put the overlap in its own structure.
Also, is the reason not to include H0 in the structure to avoid making this model specific?

bhourahine (approved):
Ready to merge, but can we discuss at next dev meeting even if merged?

bhourahine:
> Overwriting the original overlap with the real part sounds most straight-forward. I don't think we need the original overlap again in such a scenario. It probably worth to consider bundling together the ~partial charges~ populations to make the main code more agnostic about the quantities needed before introducing an imaginary overlap component.

Would be good if we can do things like Mulliken analysis in the original real basis (so back transform the sparse density matrix) as this keeps the complex weirdness more localised in code.

src/dftbp/type/integral.F90
awvwgk: Delta SCF and REKS are bypassing the issue with multiple Hamiltonians either by doing serial processing (Delta SCF) or by allocating their own sparse arrays (REKS). Splitting the Hamiltonian to a separate data structure, maybe together with the density matrix would help to create a more general CI framework in DFTB+. Still, I don't see a way to make such change immediately useful for Delta SCF or REKS.

The other issue is that the core Hamiltonian can currently not be included in this data structure since the real time propagation is using a local copy and I haven't looked through the whole code to check whether this is actually necessary.

bhourahine: The RT propagation is updating H0 when it changes geometry, but for convenience just re-used the initial ground state H0 when initially implemented. As the code flow won't go back to a ground state calc after running propagation (?) the ground state H0 can be removed and only one structure used.

awvwgk: Sounds good to me, I can check if I'm able to reduce it a single instance without breaking anything. Maybe I can give it a stab later today, but first multipoles.

bhourahine: Yes, both Delta SCF and REKS are sequential in working with determinants. They should get refactored to run them in parallel for the MPI build, but probably kept sequential for the openMP/serial case (which is where saving the memory for multiple overlap copies would matter).

PR #799, opened 2021-07-01
Error propagation and restart file module

bhourahine:
First stage towards tool for editing restart of TD propagation.
Also uses error propagation upwards for TD routines.


bhourahine:
Interesting that the mac OS regression failed with a cancel on first commit, but worked fine on re-running.

awvwgk:
Sometimes you get a bad machine for OSX on Azure / GH actions, which runs by a factor of 10 slower than usual, quite annoying.

awvwgk (approved):
Looks good to me

PR #800, opened 2021-07-01
Fix the running out of mpi-descriptors bug

bhourahine:
Replicates change from before source tree reorganisation.

Co-authored-by: Bálint Aradi <aradi@uni-bremen.de>

Replaces #790


awvwgk (approved):
Looks like a straight-forward change to me.

PR #803, opened 2021-07-06
Fix and tests for atom ext potentials

bhourahine:
Fixes an atom indexing bug and adds two quick regression tests that were tripped off by it.


awvwgk (approved):
Thanks, looks good to me.

PR #804, opened 2021-07-07
Bugfix: self-consistent DFT-D4 uses charges instead of populations

awvwgk:
Sign error in self-consistent DFT-D4 as charges in DFTB+ are actually populations, while DFT-D4 expects actual partial charges.


bhourahine (approved):
LGTM

PR #794, opened 2021-06-21
Extend workflow to include coverage testing

awvwgk:
Adds a workflow to perform a code coverage test and upload the report to codecov

PR #806, opened 2021-07-07
Coverage testing workflow

aradi:
Rebased version of #794, with directory selection and manually prepared gcov reports.

Closes #794

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,2 @@
+fixes:
+  - "/home/runner/work/dftbplus/dftbplus::"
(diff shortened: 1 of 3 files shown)


bhourahine:
@aradi Is the codecov/patch just lines not covered by the test, or is there something else breaking?

aradi:
@bhourahine As far as I can see, this is only because of not covered lines. If you look at the list, those are all files which were probably not present at the time of the last upload.

bhourahine (approved):
OK, lets see what happens on the main repository.

PR #809, opened 2021-07-08
Add tblite to relevant github actions

bhourahine:


bhourahine:
The integration test for tblite might be coming up as a fail due to needing either an append to CMAKE_MODULE_PATH  or CMAKE_PREFIX_PATH in its CMakeLists.txt
@aradi any ideas?

awvwgk:
This is an issue in the whole stack of tblite's dependencies, they don't play nicely with CMake package files yet. Will fix this at some point and test integrations upstream, but for now this is not working.

bhourahine:
@awvwgk as the ifort case is running the xtb cases, how about including this for the moment?

aradi:
Note, this is the CMake integration test, which fails, not any of the regression tests. The ifort case does not run the integration test, therefore, it does not trigger the error. I can probably have a look at the tb-lite stack, whether it can be fixed easily enough to export the proper targets...

awvwgk:
This will be fixed with #825

PR #810, opened 2021-07-09
Changed variable from keyword

bhourahine:
Status => errStatus (as keyword on some file operations).
Also fixed some line lengths in source.

### DIFF
--- a/src/dftbp/common/blacsenv.F90
+++ b/src/dftbp/common/blacsenv.F90
@@ -44,7 +44,7 @@ contains
 
 
   !> Initializes BLACS grids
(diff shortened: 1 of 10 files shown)


aradi:
I am wondering, why the rename is necessary? A variable named `status` will never collide with a `status=` keyword argument of a command/function. (Actually, even in the Fortran 2018 standard description they use variables called `status` in some of the examples...

bhourahine:
Its more that the habit of using keywords (whether a collision can happen or not) should be discouraged. Also, as the status only carries error flags, its a more informative name for 3 extra characters.

PR #811, opened 2021-07-09
Bring RangeSep routines back into main TD-DFTB routines

aradi:
- Removal of calls to indxoo and indxvv where possible

- Removal of direct calls to transQ

- Stratmann diagonalizer as alternative to Arpack for standard
  TD-DFTB.  Works for S/T, spin pol, temperature, windowing. Not yet
  for onsite corrections and not yet OpenMP

Stratmann is significantly faster for small nExc (factor 2-10), but
Arpack is faster for nExc around 100. Stratmann has accuracy issues
for osc strength (deviation 0.01 %) if initial SubSpace is too small.

Co-authored-by: Bálint Aradi <aradi@uni-bremen.de>

This is the rebased version of #781 and closes #781


aradi:
@thomas-niehaus This is the rebased version of #781 . Could you have a quick look over the "Files changed" section (which a diff between current master and this branch) to check, whether every change you wanted to make was carried over correctly. (As far as I can see, it was). I've also fixed the `old_compat` routine.

thomas-niehaus (approved):

PR #815, opened 2021-07-13
Deactivate getFileId

helmutwecke:


app/misc/skderivs/skderivs.F90
```diff
@@ -82,13 +82,13 @@ subroutine main(inp)
     write(stdout, "(A)") "Following files will be created:"
     call resize_string(buffer, 1024)
     do ii = 1, size(inp%iHam)
-      fpHam(ii) = getFileId()
+      ! fpHam(ii) = getFileId()
       strTmp = trim(inp%output) // ".ham." // i2c(ii)
       open(fpHam(ii), file=strTmp, status="replace", position="rewind")
```
aradi, line 87: Note, that won't work this way. You have to use `newunit=fpHam(ii)`, if `fpHam(ii)` is not initialized when the `open()` statement is executed!

app/misc/skderivs/skderivs.F90
```diff
@@ -14,7 +14,7 @@ program skderivs
   use dftbp_common_globalenv, only : stdOut
   use dftbp_dftb_slakoeqgrid
   use dftbp_io_charmanip
-  use dftbp_io_fileid
+  !use dftbp_io_fileid
```
aradi, line 17: Please delete unnecessary lines instead of commenting them out.

aradi:
@helmutwecke  Please note, that the code in its current form does not even compile. It would be desirable, that the commits you push are at least correct enough to be compiled. In case we have to bisect history to find a bug, broken commits are quite a pain. (We'll squash those commits at the end. Just make sure, that you do  **not** merge master into your feature branch before we did so.)

helmutwecke:
I have made some changes that solve the main problem that my previous commit had. The main obstacle I have now is at src/dftbp/timedep/linresp.F90, here the dependence on getFileId is still present and I do not understand why they are there. This is the only place where getFileId() is still present at this commit. @aradi Could you please discuss with me how should I handle the situation at linresp.F90?

helmutwecke:
@aradi I implemented the discussed changes regarding fdExc and also passed the logicals instead of file identification numbers to the respective functions. I will do the Buildbot tests. Afterwards I will merge Master and resolve conflicts.

PR #821, opened 2021-07-18
neighbour based range separated Hamiltonian

inseonglee:
When we use 'NeighbourBased' screening method, only upper-triangular LC matrix is generated.
Since REKS calculation requires fully-symmetrized LC hamiltonian, I simply add single line.

### DIFF
--- a/src/dftbp/dftb/rangeseparated.F90
+++ b/src/dftbp/dftb/rangeseparated.F90
@@ -549,6 +549,7 @@ contains
 
     call allocateAndInit(tmpHH, tmpDRho)
     call evaluateHamiltonian()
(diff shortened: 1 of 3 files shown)


thomas-niehaus:
Looks reasonable! Don't know if the codecov warnings are relevant or not, or what codecov even is. I could not attend the devel meeting today, where I would have wanted to ask this.

bhourahine:
@thomas-niehaus the codecov warnings are harmless (its measuring how much of the code the regression testing checks, but there are some issues with the Fortran testing that mean the number can sometimes change even when there is no effect).
I will merge this now, but at some point we need to find out why the neighbour based algorithm is producing non-symmetric matrices.

PR #823, opened 2021-07-21
Simplify spin polarized shifts in hamiltonian

bhourahine:


thomas-niehaus (approved):

src/dftbp/dftb/spin.F90
```diff
@@ -70,7 +71,7 @@ subroutine getSpinShift(shift, chargePerShell, species, orb, spinW)
     @:ASSERT(size(shift,dim=2)==nAtom)
     @:ASSERT(all(shape(chargePerShell)==shape(shift)))
     nSpin = size(chargePerShell, dim=3)
-    @:ASSERT(nSpin == 2 .or. nSpin == 4)
+    @:ASSERT(nSpin == 1 .or. nSpin == 3)
 
     shift(:,:,:) = 0.0_dp
     do iSpin = 2, nSpin
```
awvwgk, line 77: This change would be required to fix the regression testing
```suggestion
    do iSpin = 1, nSpin
```

awvwgk (changes requested):
Not sure if this actually improves the situation. The current routine seems quite clear to me, now that the hack if the zero based index is gone.

PR #824, opened 2021-07-22
Allow finding Slater–Koster files in DFTBPLUS_PARAM_DIR

awvwgk:
Adds environment variable `DFTBPLUS_PARAM_DIR` to search for Slater–Koster files. Variable is set automatically in testsuite to avoid using a user-specified variable while testing.

- [x] document in manual
- [x] add to changelog

### DIFF
--- a/src/dftbp/common/filesystem.F90
+++ b/src/dftbp/common/filesystem.F90
@@ -9,17 +9,28 @@
 
 !> File system utilities
 module dftbp_common_filesystem
+  use dftbp_extlibs_xmlf90, only : string, char, assignment(=)
   implicit none
   private
 
-  public :: getEnvVar, isAbsolutePath
+  public :: getEnvVar, isAbsolutePath, pathJoin, findFile, getParamSearchPath
+
+
+  !> Whether the operating system we are using is UNIX (FIXME: use preprocessor here instead)
+  logical, parameter :: is_unix = .true.
+
+  !> Path separator for this platform
+  character(len=*), parameter :: separator = merge("/", "\", is_unix)
+
+  !> Environment variable containing parameter directory for DFTB+
+  character(len=*), parameter :: paramEnv = "DFTBPLUS_PARAM_DIR"
 
 
 contains
 
 
   !> Check whether a file path is absolute
-  function isAbsolutePath(path) result(absolute)
+  pure function isAbsolutePath(path) result(absolute)
 
     !> Name of variable
     character(len=*), intent(in) :: path
@@ -27,9 +38,13 @@ contains
     !> Absolute file path
     logical :: absolute
 
-    character(len=*), parameter :: chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+    character(len=*), parameter :: chars = &
+        & "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
 
-    absolute = index(path, "/") == 1 .or. index(path, ":\") == 2 .and. scan(path, chars) == 1
+    absolute = index(path, separator) == 1
+    if (.not.(is_unix.or.absolute)) then
+      absolute = scan(path, chars) == 1 .and. index(path, ":") == 2 .and. scan(path, "\/") == 3
+    end if
   end function isAbsolutePath
 
 
@@ -47,12 +62,12 @@ contains
     call get_environment_variable(var, length=length, status=stat)
     if (stat /= 0) then
       return
-    endif
+    end if
 
     allocate(character(len=length) :: val, stat=stat)
     if (stat /= 0) then
       return
-    endif
+    end if
 
     if (length > 0) then
       call get_environment_variable(var, val, status=stat)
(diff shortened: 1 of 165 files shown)


awvwgk:
@aradi @bhourahine Let me know what you think about this solution.

src/dftbp/dftbplus/parser.F90
```diff
@@ -7624,4 +7632,55 @@ function parserVersionFromInputVersion(versionString, node) result(parserVersion
 
   end function parserVersionFromInputVersion
 
+
+  !> Get the environment variable describing the search path for DFTB+
```
bhourahine, line 7636:
```suggestion
  !> Get the environment variable describing the search path for DFTB model SK files
```

bhourahine:
You probably want to be able turn this code path off in dftb-in if not wanted. Things like OS not setting the variable, user space [thread collisions](http://fortranwiki.org/fortran/show/get_environment_variable), etc.

aradi (commented):
I like the approach a lot. I think, it is robust enough to avoid big surprises. There are two issues I see:
* We definitely have to print the names of the SK-files, which had been read, so that it can be identified, which one had been read.
* There is currently no way to tell the program to take the SK-files from the current folder and fail, if the SK-file names were not there (as it would start to look for alternatives in SK_PATH. What about skipping path-substitution not just when file path is absolute, but also, if it starts with `./`?

src/dftbp/dftbplus/parser.F90
```diff
@@ -7624,4 +7632,55 @@ function parserVersionFromInputVersion(versionString, node) result(parserVersion
...
+      outName = char(path(ip))//"/"//inName
+      inquire(file=outName, exist=exists)
+      if (exists) exit
+
+      outName = char(path(ip))//"\"//inName
+      inquire(file=outName, exist=exists)
+      if (exists) exit
+    end do
+
+    if (.not.exists .and. allocated(outName)) deallocate(outName)
```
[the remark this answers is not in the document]
awvwgk: But this will always deallocate the string we want to return. This branch should only be taken if the search was unsuccessful, *e.g.* we fall through the loop after exceeding all paths we can search

src/dftbp/dftbplus/parser.F90
```diff
@@ -7624,4 +7632,55 @@ function parserVersionFromInputVersion(versionString, node) result(parserVersion
...
+      return
+    end if
+
+    do ip = 1, size(path)
+      outName = char(path(ip))//"/"//inName
+      inquire(file=outName, exist=exists)
+      if (exists) exit
+
+      outName = char(path(ip))//"\"//inName
+      inquire(file=outName, exist=exists)
```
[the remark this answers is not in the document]
awvwgk: I can detect this accurately in Fortran as well, but just implemented this shortcut to have something for demonstration, before we commit to the idea.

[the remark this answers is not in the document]
aradi: You are right, I thought we have `return`s up there, not `exit`s.

src/dftbp/common/filesystem.F90
```diff
@@ -0,0 +1,155 @@
...
+  end function isAbsolutePath
+
+
+  !> Obtain the value of an environment variable
+  subroutine getEnvVar(var, val)
+
+    !> Name of variable
+    character(len=*), intent(in) :: var
+
+    !> Value of variable
```
aradi, line 57:
```suggestion
    !> Value of variable. Unallocated, if variable could not be retrieved.
```

src/dftbp/common/filesystem.F90
```diff
@@ -0,0 +1,155 @@
...
+      if (stat /= 0) then
+        deallocate(val)
+        return
+      end if
+    end if
+  end subroutine getEnvVar
+
+
+  !> Join two paths together
+  pure function pathJoin(prefix, suffix) result(path)
```
aradi, line 83: I'd suggest to call it `joinPaths()`.

src/dftbp/dftbplus/parser.F90
```diff
@@ -1314,6 +1318,8 @@ subroutine readDFTBHam(node, ctrl, geo, slako, poisson)
           end if
           strTmp = trim(prefix) // trim(elem1) // trim(separator) &
               &// trim(elem2) // trim(suffix)
+          call findFile(searchPath, strTmp, strOut)
+          if (allocated(strOut)) strTmp = strOut
```
aradi, line 1322:
```suggestion
          if (allocated(strOut)) call move_alloc(strOut, strTmp)
```

aradi (approved):
Apart of minimal esthetic suggestions, ready to go. I guess we would have to delete all SK-related symlinks then consistently. I've pushed a commit, which prints also the SK-names (this got abandoned on master for some reasons, this was a good place to introduce it again).

src/dftbp/dftbplus/parser.F90
awvwgk: Doesn't work because strTmp is not allocatable.

PR #825, opened 2021-07-26
Allow usage of parameter files with xTB library implementation

awvwgk:
This updates the `tblite` dependency to support parameter files as well. The format is described [here](https://tblite.readthedocs.io/en/latest/spec/parameter.html).

Also enables testing in GH actions workflow, therefore closes https://github.com/dftbplus/dftbplus/pull/809


bhourahine (approved):
LGTM

PR #829, opened 2021-08-01
Temporary disable lm xtb population print

bhourahine:
As orbital labels and m values in printing to detailed.out are incorrect for the (different to dftb) ordering.


awvwgk:
We should fix the underlying issue instead. Resorting the orbitals should be that much of an issue.

bhourahine:
@awvwgk agreed, but until fixed this make it safe (as there are already early adopters using this feature).

PR #830, opened 2021-08-02
Fixed number of elec with fixed/common Fermi

bhourahine:
As nEl was not updated for fixed Fermi energy / common Fermi level across spins.


awvwgk (approved):
Looks good to me

bhourahine:
@awvwgk Looks like there will have to be a follow up for transport as well. Should I do this as another PR, or include it here?

awvwgk:
I think this can be added here. Except you think @pecchia should have a look on this one first.

bhourahine:
@awvwgk the changes are in the main code only.

PR #832, opened 2021-08-04
Time propagation with range separation

bhourahine:
Possibly working
- [x] singlet case
- [x] triplet (but disagrees with Casida)
- [x] forces (blocked, as untested)
- [x] documentation
- [x] regression tests


thomas-niehaus:
I have tried this for Benzene and Butadiene. Benzene gives just noise, while Butadiene even returns NaN for the dipole moment. Out of the box, the TD+LC does not seem to work. (Which is a pity, because this way the TD Triplets could be tested.)

bhourahine:
Just added a C6H6 test case.

bhourahine:
@charlymedrano7 This should now include all of your PR changes and also re-bases the code on top of the changes for xTB time propagation. Can you try testing things out? It looks like the spin=0 polarised case is matching the spin free. If you want to make further changes, please base of this, not your previous branch (as there will be conflicts).
@fbonafe @thomas-niehaus could you both have a look at testing as well?

fbonafe:
Did you also compare RT and Casida without LC for the same systems @thomas-niehaus ? Do you observe the same differences in the excitation energies? I would like to know if there is a bug in the LC implementation or if this is a natural inaccuracy e.g. by the frequency discretization that we have from using the FFT.

src/dftbp/dftbplus/initprogram.F90
```diff
@@ -3644,14 +3644,22 @@ subroutine initProgramVariables(this, input, env)
...
-        call error("Electron dynamics does not work with range separated calculations yet.")
-      end if
-
       if (.not. this%tRealHS .and. input%ctrl%elecDynInp%tBondE) then
         call error("Bond energies during electron dynamics currently requires a real hamiltonian.")
       end if
 
+      if (this%isRangeSep) then
+        if (input%ctrl%elecDynInp%spType == tdSpinTypes%triplet) then
+          call error("Triplet excitation kick currently disabled for range-separated functionals")
```
charlymedrano7, line 3653: Just 'kicks'? Other types of perturbations may be blocked as well (probably they are already blocked but just to be sure)

charlymedrano7, line 3653: Just 'kicks'? Other types of perturbations may be blocked as well (probably they are already blocked but just to be sure)

test/app/dftb+/tests
```diff
@@ -18,6 +18,9 @@ deltadftb/C2H4_unpurified              #? OMP_THREADS <= 2 and not WITH_MPI
 deltadftb/indigo                       #? OMP_THREADS <= 4 and MPI_PROCS <= 2
 deltadftb/H2O                          #? OMP_THREADS <= 2 and not WITH_MPI
 
+timeprop/C4H6_rs                       #? OMP_THREADS <= 4 and not WITH_MPI
+timeprop/C4H6_rs_Singlet               #? OMP_THREADS <= 4 and not WITH_MPI
+#!timeprop/C4H6_rs_Triplet               #? OMP_THREADS <= 4 and not WITH_MPI
```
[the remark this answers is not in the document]
bhourahine: The test is commented out, so should be harmless. Looking at the commit history it might be somewhat messy to remove it from the commits, but could be taken out with another commit on top of this (but remains in the logs).

src/dftbp/dftbplus/initprogram.F90
bhourahine: Will fix the message (other types are also blocked).

src/dftbp/timedep/timeprop.F90
```diff
@@ -550,9 +550,6 @@ subroutine TElecDynamics_init(this, inp, species, speciesName, tWriteAutotest, a
       if (.not. this%tRealHS) then
         @:RAISE_ERROR(errStatus, -1, "Ion dynamics is not implemented yet for imaginary&
             & Hamiltonians.")
-      elseif (isRangeSep) then
-        @:RAISE_ERROR(errStatus, -1, "Ion dynamics is not implemented yet for range separated&
```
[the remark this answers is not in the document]
bhourahine: Harmless to leave them there until this is checked.

PR #833, opened 2021-08-04
Trim return value from unquote for ParameterFile entry

awvwgk:
```cson
Hamiltonian = xTB {
   ParameterFile = gfn2-xtb.toml  # works
   ParameterFile = "gfn2-xtb.toml"  # failed due to two trailing spaces produced by unquote
}
```


bhourahine (approved):
RTM

PR #835, opened 2021-08-09
Limit Python API installation to API enabled shared library builds

awvwgk:
Mainly to provide a useful error message instead of building a non-functional Python API.


bhourahine (approved):
RTM

PR #839, opened 2021-08-18
Allow DFTB+ to find hybrid dependencies by pkg-config

awvwgk:
This will be very helpful when building against the *tblite* stack (usually compiled with meson, therefore no CMake config files).


bhourahine (approved):
Seems OK, is there any issue for packagers downstream who want specific versions?

awvwgk:
I didn't add the `pkgconf` option to the ones active by default, meaning you have to opt-in. Similar to the CMake config files the search path is determined by an environment variable (`PKG_CONFIG_PATH`), which gives you some control to prefer one installation in case multiple are available.

PR #840, opened 2021-08-20
Simple dftd3 doc

### LINKED ISSUE #791: SimpleDftD3 undocumented
**Describe the bug**
SimpleDftD3 options does not appear in the manual.

aradi:
Documents the `SimpleDftD3` option.

Closes #791

### DIFF
--- a/doc/dftb+/manual/dftbp.tex
+++ b/doc/dftb+/manual/dftbp.tex
@@ -2714,7 +2714,9 @@ description of the methods see the following subsections):
 \item \is{DftD3}: Dispersion is calculated as in the dftd3
   code~\cite{grimme-jcp-132-154104,grimme-jcp-32-1456-1465} (see
   section~\ref{sec:dftbp.DftD3}). Modification hydrogen bond interaction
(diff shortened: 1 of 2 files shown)


doc/dftb+/manual/dftbp.tex
```diff
@@ -2878,13 +2880,16 @@ \subsubsection{SlaterKirkwood}
 \subsubsection{DftD3}
 \label{sec:dftbp.DftD3}
 
+\textbf{Note:} the \dftbp{} binary must be compiled with the DFT-D3 library
+enabled to use this feature. It is recommended to use the \is{SimpleDftD3}
+dispersion model instead, which offers similar functionality without any library
```
bhourahine, line 2885:
```suggestion
dispersion model instead, which offers similar functionality but without any library
```

bhourahine (approved):
Seems fine, but has the utf8 side-effect (fine if we are adopting this as standard, but should be in the git log somewhere in case of any future issue).

bhourahine:
Just noticed the utf-8 is its own commit, so don't squash...

awvwgk (approved):

PR #841, opened 2021-08-20
Throw error for non-existent tests

### LINKED ISSUE #671: tests processing misses junk entries - should warn/stop
**Describe the bug**
The autotests 'tests' file will not catch malformed entries that don't correspond to actual test names

**To Reproduce**

Add random entry to 
  test/prog/dftb+/tests
which doesn't correspond to an actual test

**Expected behaviour**

Throwing a warning or error on running ctest

aradi:
Closes #671


bhourahine (approved):
RTG

PR #847, opened 2021-09-13
Enable electric fields for xTB Hamiltonian

awvwgk:
Allow external potentials to interact with charges in xTB Hamiltonian.

To discuss:

- dipole moments currently don't interact with the field, should they be included for GFN2-xTB and GFN1-xTB?
  Note that `tblite` always makes the multipole moments available, regardless of the usage of multipole electrostatics

Related https://github.com/grimme-lab/xtb/issues/446


bhourahine:
Yes, the intra-atomic dipole should also be included (not sure if its worth giving an option to turn off the term for user experimentation purposes). This will eventually matter for things like 2D structures and selection rules.

awvwgk:
> Shouldn't there also be an extra energy term that looks like \mu \cdot \nabla V(\bf r) ?

Now included in the energy as well. Getting the multipole moments into the `calcEnergies` function for the inline evaluation is a bit unpractical, we should think about using a container for the external field in the future as well.

aradi (approved):

PR #848, opened 2021-09-13
Include dipole populations in dipole moment

awvwgk:
- relevant for methods evaluated with `tblite` library

### DIFF
--- a/src/dftbp/dftbplus/main.F90
+++ b/src/dftbp/dftbplus/main.F90
@@ -1227,7 +1227,7 @@ contains
     end if
 
     if (this%tDipole .and. .not. allocated(this%reks) .and. .not. this%tRestartNoSC) then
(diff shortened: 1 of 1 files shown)


bhourahine (approved):
Looks good to merge (how did you confirm the sign convention of the dipole term?).

awvwgk:
> how did you confirm the sign convention of the dipole term?

I followed the sign convention of the charge populations in the term above, since we are dealing with a dipole population here the sign has to be switched in the same way.

bhourahine:
Shouldn't there also be an extra energy term that looks like \mu \cdot \nabla V(\bf r) ?

awvwgk:
> Shouldn't there also be an extra energy term that looks like \mu \cdot \nabla V(\bf r) ?

Not completely following you here, how does this affect the total dipole moment?

bhourahine:
Sorry, should have been a comment in #847

PR #850, opened 2021-09-13
Enable xTB Hamiltonian for real-time propagation

awvwgk:
- [x] include atomic dipoles and quadrupole correctly
- [x] tests
- [x] documentation
- [x] changelog


awvwgk:
The routine to look into for the multipole populations is probably `getChargeDipole`:

https://github.com/dftbplus/dftbplus/blob/ee255735ba8c1280d5072d8bac9ffab700599b95/src/dftbp/timedep/timeprop.F90#L1328-L1330

Anything else I should be aware of when making the time propagation routines aware of the multipolar contributions to the Hamiltonian?

bhourahine:
The following simple input segfaults:

>  Geometry = GenFormat {
>    2  C
>    H
>    1 1 0 0 0
>    2 1 0 0 .7
>}
>Hamiltonian = xTB {
>method = "GFN2-xTB"
>}
>ElectronDynamics = {
>  Steps = 20000
>  TimeStep [au] = 0.2
>}

awvwgk:
At the moment the dipole and quadrupole populations are not propagated and remain at the initial density while the charges are propagated. We have to calculate the dipole and quadrupole populations in `getChargeDipole` as well, this might require inverting the dipole and quadrupole integral matrix in the same way the overlap is inverted, but I haven't yet written down all the math for this.

bhourahine:
Would only propagating the density matrix (as already done) and re-evaluating them from this not work? For a static structure, no integral evaluations would be needed, only if the atoms also move.

awvwgk:
The charge populations are currently calculated from the propagated density matrix in

https://github.com/dftbplus/dftbplus/blob/f71e42e6e86eb969dedb191bc40328d67d6b63dc/src/dftbp/timedep/timeprop.F90#L1382-L1390

A similar mechanism has to be implemented for the dipole and quadrupole populations, for this purpose we probably have to calculate the equivalent `DsqrBra`/`DsqrKet` and `QsqrBra`/`QsqrKet` matrices as in

https://github.com/dftbplus/dftbplus/blob/f71e42e6e86eb969dedb191bc40328d67d6b63dc/src/dftbp/timedep/timeprop.F90#L1719-L1731

We might not have to invert the dipole and quadrupole matrices. Or am I missing something?

awvwgk:
I think I have now correctly implemented support for dipoles and quadrupoles with xTB.

bhourahine:
@fbonafe could you try the new changes?

awvwgk:
A quick calculation with GFN1-xTB for the example from the recipes:

gfn1-nm

awvwgk:
image

Got it working for GFN2-xTB now as well.

awvwgk:
@fbonafe Just checked the Ehrenfest dynamics with xTB, we still need to implement

https://github.com/dftbplus/dftbplus/blob/f4c2a87a4a87136337527bcdae2853f78926c456/src/dftbp/timedep/timeprop.F90#L1976-L1977

for the xTB Hamiltonian by creating a dense version of

https://github.com/dftbplus/dftbplus/blob/f4c2a87a4a87136337527bcdae2853f78926c456/src/dftbp/extlibs/tblite.F90#L1460-L1466

awvwgk:
Added tests and documented the feature. Should be ready to go now.

awvwgk:
@aradi @fbonafe Seems like there is already interest in RT-TD-xTB on the mailing list. Let's move this PR forward and get finally merged.

src/dftbp/timedep/timeprop.F90
```diff
@@ -2759,9 +2886,93 @@ subroutine updateH0S(this, Ssqr, Sinv, coord, orb, neighbourList, nNeighbourSK,
...
+    !> Square quadrupole matrix
+    complex(dp), intent(inout), optional :: Qsqr(:,:,:,:)
+
+    real(dp), allocatable :: M3(:, :, :)
+    integer :: iSpin, iKS, iK
+
+    if (present(Dsqr)) then
+      Dsqr(:,:,:,:) = 0.0_dp
+      if (this%tRealHS) then
+        if (allocated(M3)) deallocate(M3)
```
aradi, line 2930: Naive question: how can `M3` be ever allocated when reaching this point?

src/dftbp/timedep/timeprop.F90
```diff
@@ -2759,9 +2886,93 @@ subroutine updateH0S(this, Ssqr, Sinv, coord, orb, neighbourList, nNeighbourSK,
...
+              & this%kPoint(:,iK), iNeighbour, nNeighbourSK, this%iCellVec, this%cellVec,&
+              & iSquare, iSparseStart, img2CentCell)
+        end do
+      end if
+    end if
+
+    if (present(Qsqr)) then
+      Qsqr(:,:,:,:) = 0.0_dp
+      if (this%tRealHS) then
+        if (allocated(M3)) deallocate(M3)
```
aradi, line 2953: The same here.

src/dftbp/timedep/timeprop.F90
awvwgk: Shouldn't be needed here
```suggestion
```

src/dftbp/timedep/timeprop.F90
awvwgk: Here it can be still allocated from unpacking the dipole integrals, which uses the same structure due to symmetry ;)

fbonafe (approved):

awvwgk:
Thanks for reviews, @fbonafe and @aradi. The tests on buildbot just passed, therefore I'll go ahead and merge.

PR #853, opened 2021-09-14
Avoid timing output overflow after 28 hours

aradi:


bhourahine:
I guess longer than a 2 year run time is unlikely (vague recollection of Thomas Köhler getting close back in the day).

aradi:
@bhourahine :smile: Actually, with the current settings we should be fine until ~3 years running time (or the first power outage, depending which comes first...). But output would look decent (number separated by space from previous field) only until 115 days.

bhourahine:
OK, could you put one more figure on the format then? As runs for more than 4 months are plausibly happening.

aradi:
Then we would have to decrease the nr. of digits. Actually, this way it would still show the correct number up to 3.16 years (rather then showing "****"), only leading space would missing in front for runs longer than 115 days. But that won't be too disturbing (apart of some esthetics), as the numbers would be still easily recognizable (and parseable).

bhourahine (approved):
Ready to merge

PR #854, opened 2021-09-15
Enable OMP threads for ELSI solvers

### LINKED ISSUE #843: Enable Multi-Threading When Using ELSI
When using DFTB+ with the ELSI solver library, DFTB+ disables multi-threading. 

Remove the following piece of code:
```
      #:if WITH_OMP
        if (omp_get_max_threads() > 1) then 
          call error("The ELSI-solvers should not be run with multiple threads. Set the&
              & environment variable OMP_NUM_THREADS to 1 in order to disable multi-threading.")
        end if
      #:endif
```
Which is on line 1868 of `src/dftbp/dftbplus/initprogram.F90`

**Describe alternatives you've considered**
Of course, the user can modify this code quite easily themselves. 

**Additional context**
@vyu16 might know more about why this decision was made. See  #254 as well. If it really just is an issue of testing, it should be possible to address that... though maybe time consuming! 

NTPoly can freeze with certain MPI implementations if we force the number of threads to be one (https://github.com/william-dawson/NTPoly/issues/143). That's a bug in my code, not in DFTB+, and I have a fix for this bug on the way (https://github.com/william-dawson/NTPoly/pull/170). When I push this upstream to ELSI, the issue will be resolved. Nonetheless, NTPoly is faster with threading than pure MPI. I imagine some of the other ELSI solvers are too. So having a roadmap for all solvers to be used as you wish would be good.

aradi:
Note: User still has to enable UseOmpThreads in the Parallel block.

Closes #843

### DIFF
--- a/src/dftbp/dftbplus/initprogram.F90
+++ b/src/dftbp/dftbplus/initprogram.F90
@@ -1865,13 +1865,6 @@ contains
         call error(tmpStr)
       end if
 
-      #:if WITH_OMP
-        if (omp_get_max_threads() > 1) then
-          call error("The ELSI-solvers should not be run with multiple threads. Set the&
-              & environment variable OMP_NUM_THREADS to 1 in order to disable multi-threading.")
-        end if
-      #:endif
-
       ! Would be using the ELSI matrix writing mechanism, so set this as always false
         this%tWriteHS = .false.
 


bhourahine (approved):
Ready to merge (will confirm if there are ELSI solvers still with this requirement, possibly then needing another PR).

PR #857, opened 2021-09-22
Add check for minimal Fortran compiler versions

aradi:


cmake/DftbPlusUtils.cmake
```diff
@@ -700,3 +683,27 @@ function (dftbp_create_lcov_targets lcov genhtml lcov_output_dir lcov_base_dir l
...
+#       message is printed about the check.
+#
+function (dftbp_check_minimal_compiler_version lang compiler_versions)
+  while(compiler_versions)
+    list(POP_FRONT compiler_versions compiler version)
+    if("${CMAKE_${lang}_COMPILER_ID}" STREQUAL "${compiler}")
+      if (CMAKE_${lang}_COMPILER_VERSION VERSION_LESS "${version}")
+        message(FATAL_ERROR "${compiler} ${lang} compiler is too old (found \"${CMAKE_${lang}_COMPILER_VERSION}\" < \"${version}\")")
+      else()
+        message(STATUS "${compiler} ${lang} compiler is recent enough (found \"${CMAKE_${lang}_COMPILER_VERSION}\" >= \"${version}\")")
```
awvwgk, line 705: Is this message relevant? CMake will print out the compiler version in any case.

cmake/DftbPlusUtils.cmake
```diff
@@ -223,6 +223,10 @@ function (dftbp_ensure_config_consistency)
 
   endif()
 
+  # Check minimal compiler versions
+  set(fortran_minimal_versions "GNU;9.0" "Intel;19.0" "NAG;7.0")
```
awvwgk, line 227: Does this work reliably with CMake using the semicolon as default delimiter? Or does it just assume that we will later always remove the entries of the list in pairs?

cmake/DftbPlusUtils.cmake
aradi: We can leave it away, indeed.

cmake/DftbPlusUtils.cmake
aradi: Yes, the routine doing the check pops always 2 entries at a time. One could equivalently write it as `set(... "GNU" "9.0" "Intel" "19.0")`, but I've found the form above considerably easier to read.

awvwgk (approved):

PR #858, opened 2021-09-24
Sym links

bhourahine:
Removed a large number (but not all) of the symlinks from the regression tests.
Outstanding issues:
What to do about cases with explicit specification ( H-H = "./H-H.skf" ) as the prefix mechanism won't(?) work for this case.


awvwgk (approved):
Thanks a lot for taking this effort. This is a good step forward already.

PR #859, opened 2021-09-24
PR to integrate rangesep branch in Casida

thomas-niehaus:
- Separate routines for rangesep TD calculations were removed, this avoids the enormous code duplication  
- Code can now do TD-LC Triplets and spin-polarized calculations, tests were added
- Removal of intermediate helper routines


thomas-niehaus:
I know, this is a massive PR, several routines are involved. But I could not break it down further. Sorry for that.

thomas-niehaus:
There are still a large number of direct calls to BLAS/LAPACK.

bhourahine:
Do you have access to the Bremen buildbot set-up? There are several compilers failing tests there with CMAKE_BUILD_TYPE=Debug

thomas-niehaus:
Yes, I have access to dpdevel. I tried gfortran-9 and NAG there and had no issues. Can you tell me which test fails exactly?

bhourahine:
See build numbers 637 and 634 for dftb_reg_devel_* on the buildbot web interface (the gcc7.5 is fail is due to a, now fixed, bug in the main branch).

bhourahine:
Just pushed fixes for the cases showing up with the more sensitive compiler tests (will see what the others make of this).

bhourahine:
Testing on buildbot now, hopefully should pass all compiler combinations.

thomas-niehaus:
Can you tell me how I can access the buildbot web interface? Thx

bhourahine:
There might be a bug in the spin indexing, as 
ss = getIA(win(jbs), 3)
is used in several places, but 
subroutine getSPExcitations()
only initialises getIA(:, :nSpin), so case 3 is never set(?)
Or am I missing something obvious?
This is in part why the code coverage dropped, as there are several places which then test for ss==2

thomas-niehaus:
Initialization of getIA is done as follows (subroutine getSPExcitations)
  getIA(ind,:) = [ii, aa, iSpin]
So sth like getIA(:,3) should be defined and be equal to either 1 or 2 (depending on iSpin)
Is this answering your question?

bhourahine:
Not quite, the current set of test examples never execute parts of the code that seem to be required for spin polarized cases.
You can see this in the code coverage above ( src/dftbp/timedep/linrespcommon.F90	(-10.08%) ). If you click through into the report you can see that line 650 (for example) is never run over all of the tests.
Either 1) there is an indexing bug which stops it being called or 2) none of the code test cases include spin polarization where that type of transition is included.

bhourahine:
@thomas-niehaus That is probably less of a concern if the behaviour is portable between machines (i.e. the same strengths come out with the same settings on different platforms). It could then be documented and users can test effects of expanding the space if necessary. Presumably the discrepancy comes from states outside of the sub-space that have non-negligible dipole moments?

On the topic of documentation, could you add a description of the Stratmann keywords to the manual as part of this PR?

thomas-niehaus:
Ben, could you check whether the structure of the docu conforms?

bhourahine:
@thomas-niehaus have edited it a bit to match the parser behaviour and Anglicised Diagonalizer -> Diagonaliser

bhourahine (approved):
Ready to merge, but eventually needs follow on pull requests to:
a) Convert direct blas calls to wrapped ones (as this would cause problems on machines with unusual precision models and misses out on any future improvements to calls like gemm). 
b) Enable Stratmann even if not linked against ARPACK.

thomas-niehaus:
Thanks for updating the docu! Can you merge, Ben?

 For the next PR I see the following tasks:
- [ ] Get rid of requirement of ARPACK library -> Stratmann default
- [ ] Onsite correction for Stratmann -> Adriel
- [ ] Use of new indexing routines in PP-RPA -> Adriel 
- [ ] Convert BLAS/LAPACK calls to wrapped ones
- [ ] Derived type to simplify call to transQ
- [ ] cExchange constant does not work as expected
- [ ] Remove overlap construction in linrespcommon ?

bhourahine:
@thomas-niehaus Could I also ask for?

- [ ] folded spectrum for interior states

There will be a slight delay to merging (should not take too long) while we sort out advancing the parser version number, as the Casida input changes are not backward compatible for the existing released code under parser 9.

PR #861, opened 2021-09-24
Changed test to match directory

bhourahine:
Test naming suggests gfn2, test _was_ gfn1.


awvwgk (commented):
How about renaming the directory instead? This would keep the autotest reference in tact.

bhourahine:
@awvwgk counter proposal: have directories with both hamiltonian cases (?)

awvwgk (approved):
I see, there is already a periodic input for GFN1-xTB using succinic acid, means there is currently no periodic input testing GFN2-xTB. Recalculating the benzene testcase with GFN2-xTB it is than.

Let's save the space in the test suite for more interesting systems than molecular crystals to test xTB. Are there mindless-like solid structures? The more different elements in one (small) input the better.

bhourahine:
Sounds like a high entropy alloy might be suitable, does something like AgMnGeSbTe4 sound usable? Its a narrow gap semiconductor but the unit cell would have to be too small for real disorder.

awvwgk:
> Sounds like a high entropy alloy might be suitable, does something like AgMnGeSbTe4 sound usable?

If we can get the SCC to converge this might be a suitable testcase to cover many different element pair interactions in xTB.

awvwgk:
I'm going ahead and merge this patch. Further test cases can be discussed and added separately.

PR #864, opened 2021-09-29
Add external dipole potential in gradient calculation

awvwgk:
Maybe need a more robust mechanism to collect multipole potentials from potential object?


awvwgk:
Included in #887

PR #876, opened 2021-10-21
Update dftd3-lib download

awvwgk:
Update download script to use updated dftd3-lib source.


aradi:
@awvwgk You're really just too fast. :smile: But my PR (#877) contains also an entry in the Change log about the fix. :wink:

PR #878, opened 2021-10-26
Extention of C API for electrostatic potential export

Mazay0:
I didn't add new tests for this functionality, instead I have extended the `fileinit` test.


src/dftbp/api/mm/capi.F90
```diff
@@ -250,6 +250,29 @@ subroutine c_DftbPlus_processInput(handler, inputHandler, atomListHandler)&
 
   end subroutine c_DftbPlus_processInput
 
+  !> Obtain electrostatic potential in specified points
+  subroutine c_DftbPlus_get_potential(handler, n_locations, pot, locations)&
```
aradi, line 254: I'd suggest to consistently add the term `elstat` or `ElStat` as a prefix to the potential.  (e.g. `c_DftbPlus_get_elstat_potential`, `instance%getElStatPotential`, etc.) That would be more specific and aligned with input keyword.

src/dftbp/api/mm/capi.F90
```diff
@@ -250,6 +250,29 @@ subroutine c_DftbPlus_processInput(handler, inputHandler, atomListHandler)&
...
+      & bind(C, name='dftbp_get_potential')
+
+    !> handler for the calculation
+    type(c_DftbPlus), intent(inout) :: handler
+
+    !> Number of requsted points
+    integer, value, intent(in) :: n_locations
```
aradi, line 261: Please honor our variable name convention.
```suggestion
    integer, value, intent(in) :: nLocations
```

bhourahine (approved):
Ready to merge

PR #880, opened 2021-11-05
Optionally fold matrix going to ARPACK solver

bhourahine:
Folded spectrum for Arpack solver, can be used to solve for states from higher transitions without calculating lower transitions. Use case for example to find states around an energy range for larger systems (probably want to simultaneously use an energy window in that case).


bhourahine:
@thomas-niehaus this is possibly working for Arpack, and for the test case it is around 7x faster to get the 20 states near to 6eV compared to solving for ~70 states to get to that energy range. There seems to be some differences in oscillator strengths but not energies in that case.

thomas-niehaus:
@bhourahine not sure if this functionality is really worth having. I never encountered an application where this would be needed. Do you, Ben?

bhourahine:
> @bhourahine not sure if this functionality is really worth having. I never encountered an application where this would be needed. Do you, Ben?

It would allow independent parallelism over transitions, so for cases with many low excited states could be useful.

thomas-niehaus:
ok!

bhourahine:
note: will eventually need a PR to select lower as well as upper energies for transitions.

PR #882, opened 2021-11-11
Neighclass

aradi:
This is not a PR, just base for discussions about hiding the neighbour list details.

I've implemented a simple test `app/neightester/test_neighlist` which sums up repulsives, up to max. 4 neighbours per atom using 3 different approaches:

* Classical one: iterating directly over the arrays in `TNeighbourLIst`
* Using an abstract factory, which (for every thread) creates a neighbour iterator, which retursn neighbours in chunks for any atoms.
* Using an abstract neighbour map, which returns the neighbours as allocatable arrays with the correct size.

Testing on SiC512, it seems the the abstract approaches come with a penalty of ca. 0.5ms. Does not sound much, but makes the calculation 10 times slower in this case...

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -10,5 +10,7 @@ if(NOT BUILD_EXPORTED_TARGETS_ONLY)
   endif()
 endif()
 
+add_subdirectory(neightester)
+
 # Propagate PKG_CONFIG_LIBS upwards
 set(PKG_CONFIG_LIBS "${PKG_CONFIG_LIBS}" PARENT_SCOPE)
--- a/app/neightester/CMakeLists.txt
+++ b/app/neightester/CMakeLists.txt
@@ -1,0 +1,12 @@
+set(projectdir ${PROJECT_SOURCE_DIR})
+
+set(sources-fpp
+    test_neighlist.F90)
+
+set(fypp_flags ${FYPP_BUILD_FLAGS})
+list(APPEND fypp_flags -I${projectdir}/src/dftbp/include)
+
+dftbp_preprocess("${FYPP}" "${fypp_flags}" "F90" "f90" "${sources-fpp}" sources-f90-preproc)
+
+add_executable(test_neighlist ${sources-f90-preproc})
+target_link_libraries(test_neighlist PRIVATE dftbplus)
--- a/app/neightester/test_neighlist.F90
+++ b/app/neightester/test_neighlist.F90
@@ -22,10 +22,10 @@ module test_neighlist
 #:endif
   implicit none
 
-  integer, parameter :: maxAtom = 512
-  integer, parameter :: maxNeigh = 1000
-  integer, parameter :: nRepeat = 10000
-  real(dp), parameter :: cutoff = 20.0_dp
+  integer :: maxAtom
+  integer :: maxNeigh
+  integer :: nRepeat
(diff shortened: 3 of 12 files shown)


awvwgk:
GCC 11.1.0, 4 threads
```
 CUTOFF:   20.000000000000000     
 REPETITIONS:        1000
 RES ARRAY:   540.52544028631746        1.2145120650529862E-005
 RES ITER :   540.52544028631746        8.2903087139129641E-005
 RES Map :   540.52544028631746        5.2069962024688724E-005
 TIMEDIFF/ITER:   7.0757970213890077E-005   3.9924841374158861E-005
```

Intel 2021.2.0, 4 threads
```
 CUTOFF:   20.0000000000000     
 REPETITIONS:        1000
 RES ARRAY:   540.525440286318       4.463964700698852E-005
 RES ITER :   540.525440286318       6.377599239349365E-004
 RES Map :   540.525440286318       5.231381058692932E-004
 TIMEDIFF/ITER:  5.931202769279480E-004  4.784984588623047E-004
```

bhourahine:
Probably a good idea to dial the number of repeats up and move the timing calls outside the loop, as the overhead on these will be similar to the neighbour parts:

`
    call cpu_time(tArr1)
    do iRep = 1, nRepeat
      call iterateOverNeighArray(dpmain%neighbourList, pRepCont, dpmain%img2CentCell, dpmain%coord,&
          & dpmain%species, resArr)
      resArrSum = resArrSum + resArr
    end do
    call cpu_time(tArr2)
    tArrSum = tArrSum + (tArr2 - tArr1)

    call cpu_time(tIter1)
    do iRep = 1, nRepeat
      call iterateOverNeighIter(neighIterFact, pRepCont, dpmain%coord, dpmain%species,&
          & dpmain%img2CentCell, cutoff, nAtom, chunkSize, resIter)
      resIterSum = resIterSum + resIter
    end do
    call cpu_time(tIter2)
    tIterSum = tIterSum + (tIter2 - tIter1)

    call cpu_time(tMap1)
    do iRep = 1, nRepeat
      call iterateOverNeighMap(neighMap, pRepCont, dpmain%coord, dpmain%species, cutoff, nAtom,&
          & resMap)
      resMapSum = resMapSum + resMap
    end do
    call cpu_time(tMap2)
    tMapSum = tMapSum + (tMap2 - tMap1)
`

awvwgk:
Another note, `cpu_time` will in case of threaded execution count the time spend in all threads, therefore the timings are likely to be wrong. I didn't observed any faulty behaviour as promised, not sure whether that good or bad ;). I tried a couple more versions of GCC but didn't see any race conditions.

I tried timing with `system_clock`, but the individual runs are too short to be captured with a long int, at least for all systems I have tried (CO2, 5x5x5 supercell). One observation I made in this context is that the setup of the neighbourlist in `handleCoordChange` is actually the time determining step on my machine, the repulsion calculation run in basically no time.

aradi:
OK, thanks for the comments so far. I've rearranged the timing, and played around a bit more. The overhead for the OOP-approaches seems to be in most cases around 1e-6s / atom / iteration. Unfortunately, looking up the most recent OpenMP 5.2 standard, following line can be found among the restirctions for `firstprivate`:

> If the list item is a polymorphic variable with the ALLOCATABLE attribute, the behavior is
unspecified.

That means, the version with the neigbhour iterator factory  is not OpenMP-compliant... :disappointed:

PR #884, opened 2021-11-12
Update version of tblite

awvwgk:
- update periodic test cases to account for change in Ewald sum
- update to dftd4 version 3.3.0
- update to s-dftd3 version 0.4.2
- update to multicharge version 0.1.2


bhourahine (approved):
Suggest adding a comment in CHANGELOG.rst about the cut-off, just in case of future user confusion (e.g. very large molecule in free space vs a periodic box will of course show this in the energy).
Otherwise ready.

bhourahine:
The gcc fail was on 440468c was reported as a compiler internal error, not sure how to progress this other than trying to replicate outside of the CI where the logs can be obtained for a bug report. Options seem to be a) ignore it, leading to that regression test perennially failing b) try for a work around or c) see if other available gcc versions behave.

awvwgk:
> see if other available gcc versions behave.

I'm usually running coverage locally with GCC 11 and in CI with GCC 10. Could also reproduce the ICE with GCC 9 locally when coverage builds are enabled. For our testing in DFTB+ we could just bump the GCC version to 10 for the coverage build, but keep the debug builds at GCC 9.

awvwgk:
Coverage is now a complete mess due to the compiler change...

.github/workflows/build.yml
```diff
@@ -27,14 +27,25 @@ jobs:
...
             config: Coverage
+            version: 9
+          - os: ubuntu-latest
+            mpi: nompi
+            config: Coverage
+            version: 10
+          - os: ubuntu-latest
+            mpi: nompi
+            config: Coverage
+            version: 11
```
awvwgk, line 43:
```suggestion
```

PR #885, opened 2021-11-12
Refactor repulsive & add ChIMES correction

aradi:
Refactors and generalizes the repulsive contribution, so that it can use arbitrary pair-repulsives via an abstract interface. Additionally, it provides a container which can sum up over arbitrary number of repulsives. The container is used to implement the ChIMES repulsive correction.

- [x] The testparameter download must be changed back to use the offical repository. This can be done, once this [PR](https://github.com/dftbplus/testparams/pull/3) had been merged.

Note: The problem with GFortran < 9 had been solved with an appropriate patch of the ChIMES interface (rk-lindsey/chimes_calculator#13).

repulsive


awvwgk:
Introducing a new naming scheme for the interfaces of the interaction container?

src/dftbp/dftb/repulsive/chimesrep.F90
```diff
@@ -0,0 +1,317 @@
...
+    module procedure TChimesRep_construct
+  end interface
+
+#:else
+
+  type :: TChimesRepInp
+  end type TChimesRepInp
+
+  type :: TChimesRep
+  end type TChimesRep
```
awvwgk, line 77: I think it is preferable to remove the (private) members of the type which are unavailable without Chimes and but keep the type mostly intact.

src/dftbp/dftb/repulsive/chimesrep.F90
```diff
@@ -0,0 +1,317 @@
...
+
+    stress(:,:) = this%stress
+
+  end subroutine TChimesRep_getStress
+
+#:else
+
+  !> Dummy initializer in case code was compiled without ChIMES
+  subroutine TChimesRep_init()
+  end subroutine TChimesRep_init
```
awvwgk, line 312: Similarly, just remove the body of the procedures and replace it with an error instead if having a separate stub?

aradi: This is open for discussion. I can see following advantages, if we only provide names without content (empty type, dummy subroutine without args):
* we do not need to guard the module imports by preprocessor instructions, as the name exists, 
* we are forced, however, to guard all places, where the non-existing types and subroutines are actually used, by `#:if WITH_*` instructions. Forgetting such a guard would cause a compile time error (as opposed to a run-time error, we would get by using dummies with the same public interface as the original ones),
* coverage won't decrease due to all the `error stop`s in the dummy routines, which are (hopefully) never executed during the tests.

aradi:
@bhourahine All issues raised in #899 had been fixed here now.

aradi:
I'll will squash it and rebase on current master.

awvwgk (approved):
Looks good to me.

PR #886, opened 2021-11-14
Change nrOfAtoms => getNrOfAtoms in api

bhourahine:
To match other calls to get information. Also changed to return
number of central cell atoms (just in case called for a transport
geometry).


aradi (commented):
My original intention was to name it as the quantity (without the `get` prefix), as it returns it as a function and not as a subroutine. I think, it looks better in expressions, if you do not have the `get` prefix, e.g. `allAtoms = inst1%nrOfAtoms() + inst2%nrOfAtoms()`. Let's discuss first, whether this convention makes sense at all. (If not, change as proposed in this PR, if yes, we would have to change `getMaxAngFromSlakoFile` to `maxAngFromSlakoFile` instead, probably...)

PR #887, opened 2021-11-15
Fix error in electric field for dipoles

awvwgk:
Notes on conventions:
- atomic dipole moment (`multipole%dipoleAtom`) are given as –μ
- electric field energy is U = –μ·E
- external dipole potential is dU/d(–μ) = E

Now matches with `tblite` implementation.


bhourahine:
Confirm that energy and dipole moment are now correctly related, but forces are still not right (H2O GFN1 example from tests)
Difference between finite difference and analytical forces:
       2.558984971768E-17       2.529434499556E-02       3.254013112451E-17
       4.886487205982E-17      -1.264717248603E-02      -8.196499003212E-03
      -7.445472177750E-17      -1.264717248603E-02       8.196499003212E-03

bhourahine (approved):
LGTM

PR #891, opened 2021-11-18
Replace dftd3-lib with s-dftd3 library

### LINKED ISSUE #708: Replace dftd3-lib

DFT-D3 is currently implemented using a redistributed version of the original `dftd3` program licensed under GPL-1.0 (or later). Due to this license incompatibility DFTB+ is usually shipped without DFT-D3 which is a pity since dispersion interactions are missing in DFTB since it is based on semi-local DFT and must therefore be included by some means.

Other alternatives are LJ type, Slater-Kirkwoord, DFT-D4 or MBD dispersion corrections, which are unproblematic since they are either built-in or compatibly licensed, but none of them allows to reproduce results calculated with DFT-D3.

Parts of a DFT-D3 reimplemenation have already been added to DFTB+ as `SimpleDFTD3` this implementation could be expanded and completely replace the `dftd3-lib` or the [upstream project](https://github.com/awvwgk/simple-dftd3) can be used as feature-complete external dependency and substitute the current `dftd3-lib`.

**Additional context**

See https://github.com/dftbplus/dftbplus/issues/703#issuecomment-764855853 for some details.

awvwgk:
- remove option WITH_DFTD3, add new option WITH_SDFTD3
- purge dftd3-lib integration from code base
- enable modified zero damping function

Closes #708


awvwgk:
Almost living up to the promise of being a drop in replacement for `dftd3`. I only had to recalculate forces in one testcase which were slightly above the test tolerance. But otherwise there shouldn't be any change noticeable.

This is ready to go from my side, however there are some open questions:

- imply `WITH_SDFTD3` in case we have `WITH_TBLITE`?
- keep native SimpleDFTD3 implementation with only D3(BJ) for supporting helical boundary conditions in the future?
- make `WITH_DFTD3` an error if provided to CMake?

bhourahine (approved):
LGTM

awvwgk:
Thanks, I'll go ahead an merge.

PR #892, opened 2021-11-21
Add CMake find module for ELSI

awvwgk:
- ELSI CMake config currently exports incomplete dependency (https://gitlab.com/elsi_project/elsi_interface/-/issues/57)


aradi (commented):
We also rely on the `elsi::pexsi` target, especially as that decides, whether C++ is enabled as language or not. Is there any way to find out, whether the ELSI found via pkg_conf was compiled with or without PEXSI?

awvwgk:
Actually, this would be ELSI's job to set the C++ dependency correctly in the `elsi::elsi` target in the CMake config file as well as export the dependency on the `-lstdc++` in the pkg-config file. However, I don't think we can count on this. We could scan the `ELPA_LIBRARIES` to see whether `-lpexsi` is present and generate the `elsi::pexsi` target in this case.

Note that I still can't build PEXSI because some incompatibilities with SuperLU_dist > 6.2.0.

awvwgk:
Currently fails on buildbot since MBD also tries to use ELSI but does not profit from the new mechanism for some reason.
```
/mnt/scratch/buildbot/worker/dftb_reg_prod_gnu-11/build/external/mbd/origin/src/mbd_elsi.F90:8:5:
    8 | use elsi, only: elsi_handle, elsi_init, elsi_set_mpi, elsi_set_blacs, &
      |     1
Fatal Error: Cannot open module file ‘elsi.mod’ for reading at (1): No such file or directory
compilation terminated.
make[2]: *** [external/mbd/origin/src/CMakeFiles/mbd.dir/build.make:348: external/mbd/origin/src/CMakeFiles/mbd.dir/mbd_elsi.F90.o] Error 1
make[2]: Leaving directory '/mnt/scratch/buildbot/worker/dftb_reg_prod_gnu-11/build/_build'
make[1]: *** [CMakeFiles/Makefile2:1334: external/mbd/origin/src/CMakeFiles/mbd.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

cmake/Modules/Findelsi.cmake
```diff
@@ -0,0 +1,24 @@
...
+    elsi::elsi
+    INTERFACE
+    "${ELSI_LINK_LIBRARIES}"
+  )
+  target_include_directories(
+    elsi::elsi
+    INTERFACE
+    "${ELSI_INCLUDE_DIRS}"
+  )
+
```
awvwgk, line 17:
```suggestion

  # Set this for ELSI detection in libmbd
  set(ELSI_LIBRARIES "${ELSI_LINK_LIBRARIES}")

```

external/mbd/CMakeLists.txt
```diff
@@ -6,6 +6,14 @@ set(ENABLE_C_API OFF)
 if(WITH_MPI)
   get_target_property(SCALAPACK_LIBRARIES Scalapack::Scalapack INTERFACE_LINK_LIBRARIES)
 endif()
+if(WITH_ELSI)
+  # FIXME: Hacky CMake magic
+  #
+  # libmbd will link against ELSI using the ELSI_LIBRARIES variable and ignore the
+  # elsi::elsi target, but that doesn't mean we can't trick it via CMake variable
+  # expansion to still use the correct target:
+  set(ELSI_LIBRARIES elsi::elsi)
```
awvwgk, line 15: @aradi Hope this is okay for you.

aradi: @awvwgk Could you test, whether libmbd/libmbd#31 would solve the problem? Then probably won't need this horrible hack. :wink:

awvwgk: That would work even better.

awvwgk:
@aradi started a new run on buildbot, but I think some of the GCC toolchains still export the CPATH environment variable for ELSI and might therefore fail.

aradi:
@awvwgk Should work now, I've commented out `CPATH` in all ELSI modules on dpdevel.

awvwgk:
Regression tests are passing now.

awvwgk:
Anything left to be addressed? Tomorrow is the ELSI meeting I might want to demonstrate this feature, preferably from the DFTB+ default branch.

cmake/Modules/Findelsi.cmake
```diff
@@ -0,0 +1,34 @@
...
+      "${ELSI_LINK_LIBRARIES}"
+    )
+    target_include_directories(
+      elsi::elsi
+      INTERFACE
+      "${ELSI_INCLUDE_DIRS}"
+    )
+
+    if("${ELSI_LINK_LIBRARIES}" MATCHES "pexsi")
+      add_library(elsi::pexsi INTERFACE IMPORTED)
```
aradi, line 20: I'd suggest to define an explicit dependency as well:
```
target_link_library(elsi::pexsi INTERFACE elsi::elsi)
```
So linking against `elsi::pexsi` only would also automatically add the right linking directives.

cmake/Modules/Findelsi.cmake
```diff
@@ -0,0 +1,34 @@
...
+      target_link_libraries(
+        elsi::elsi
+        INTERFACE
+        "stdc++"
+      )
```
aradi, line 25: I think, we should not link `stdc++` directly here, as it may be called differently (or require multiple libraries) on some systems. It should be delegated to the caller to call `enable_language(cxx)` if the target `elsi::pexsi` exists. (We have to do that with the default finder as well...)

aradi (approved):
LGTM, if it works with the suggested two minor changes, I'd recommend to apply them before merging.

cmake/Modules/Findelsi.cmake
aradi: Just an additional note on this. Calling `enable_language(cxx)` in the host `CMakeLists.txt` may be not enough, as the caller won't know, that this target must be linked with the C++ standard library. It seems that we have additionally to set the `IMPORTED_LINK_INTERFACE_LANGUAGES` property of `elsi::pexsi` to `CXX` in the finder.

aradi: OK, this does not work as I thought it should, since the mentioned property can not be set for `INTERFACE` targets. :disappointed:  I'll play around with it more, but for the moment, probably we should go with `stdc++` in there, although that is not fully transferable...

awvwgk: Observed the same thing. It would be preferable to have ELSI correctly export the dependency on the C++ libraries required for PEXSI. In DFTB+ we shouldn't have to care whether or not a C++ dependency ends up in our link line.

aradi:
I think, I have a solution which abstracts out the libstc++ dependeny. I've pushed it directly here. As far as I can judge, it is fully transferable. I also fixed the `elsi::pexsi` - `elsi::elsi` dependency, my previous suggestion was wrong, it needs to be in the opposite order.

aradi (approved):
In case it passes the tests and you have no objections, it is now really ready to go. :wink:

awvwgk:
All tests on buildbot.

PR #896, opened 2021-11-24
Rationalise the ConvergentOnly options

### LINKED ISSUE #895: ConvergentSCCOnly for general properties
Currently the ConvergentSCCOnly option is only available for some force-related modes of operation, which prevents results files being produced if other ground state calculations have not converged.

Rationalise use of ConvergentSCCOnly in the parser for general SCC calculations, not just to allow geometry continuation.

bhourahine:
As both force and the dispersion were funnelling through the same
internal variable (and there was no way to get tag files unless
either converged or using one of these modes).

Closes #895


awvwgk (approved):
Looks good to me.

PR #897, opened 2021-11-26
Make lapack public instead of interface

steinmig:
This caused some problems for me in building a conan package of DFTB+.

Is there a special reason / argument to be made to have lapack be linked as an interface?

### DIFF
--- a/src/dftbp/CMakeLists.txt
+++ b/src/dftbp/CMakeLists.txt
@@ -129,7 +129,7 @@ if(WITH_ARPACK)
   target_link_libraries(dftbplus PRIVATE Arpack::Arpack)
 endif()
 
-target_link_libraries(dftbplus INTERFACE LAPACK::LAPACK)
+target_link_libraries(dftbplus PUBLIC LAPACK::LAPACK)
 
 if (WITH_GPU)
   target_link_libraries(dftbplus PUBLIC Magma::Magma)


awvwgk (approved):
Thanks for sharing.

Can you post a link to the conan package once it is available?

steinmig:
@awvwgk I am sorry, I didn't have time to check on this more in the last days and I just found out that the linking problem that I encountered with conan is a bit more complicated than I thought when opening this PR.

I thought I got it to work with the `public` link, but was mistaken, because it only worked because cmake was executed twice in the pipeline, which somehow fixed the linking. I was trying to build the package with tblite and its dependencies and I think the problem comes from the fact that tblite also links lapack `public`. If I comment out the changed line of this PR, I get the conan cmake to work in the first try, but I think this is only valid when building with tblite. When I try to build dftb+ locally with the line comment out without tblite, I get linking errors again. So this line could be conditional on tblite, but that feels backwards to me, but rather it should be changed for tbllite? I have not checked the other repos from you though, do they also bring their own lapack?

awvwgk:
Do you build tblite as subproject in DFTB+ or as a separate conan package and than find it via CMake / pkgconf again? While it is convenient to use the vendored version of a library this is usually highly discouraged in any packaging ecosystems, at least all I have worked with so far.

From my understanding of conan it is a source package manager rather than a binary one (it builds everything on the fly like spack). However, I'm not sure about the details, last time I checked conan using Fortran was a bit messy, not sure if this has changed. As a tip, most of the DFTB+ dependencies are available on conda-forge, the recipes in the feedstock repos can serve as a knowledge base for building even the most complex projects.

The caveat might be that DFTB+ provides a custom LAPACK and BLAS finder to overwrite the CMake default ones, while tblite and its dependency stack use the CMake default ones. DFTB+ handles this by finding LAPACK/BLAS first and have the tblite stack just use the same target and everything turns out to be consistent. It becomes easier when building tblite separately using meson and finding it in DFTB+ via pkgconf.

steinmig:
I am building a single conan package with DFTB+ handling the build of all dependencies (besides cmake and lapack, which are both imported as custom conan packages).

The used container only contains this conan version of lapack and I am setting all the env variables so that every library should find it. Also the build of dftb+ works and even the linking of libdftbplus works, but the linking of lapack with the dftb+ binary (+ all test binaries) fails in the first try with the current lapack linking setup

PR #899, opened 2021-11-29
Refactor repulsive and add ChIMES support (pointer based version)

aradi:
Functionality identical to #885, but using pointers instead of allocatables.

It needs less workaround for the Intel compiler during the allocation phase, however it passes a `type(TRepulsivePtr)` around (instead of `class(TRepulsive)`) and needs dereferencing (`call repulsive%ptr%getEnergies`) when components are accessed.


src/dftbp/dftbplus/initprogram.F90
```diff
@@ -5984,24 +5986,48 @@ subroutine initSccCalculator_(env, orb, ctrl, boundaryCond, coulombInput, shortG
...
-    input%nAtom = nAtom
-    input%isHelical = isHelical
-    input%twoBodyCont = twoBodyCont
-    allocate(splinePolyRep)
-    call TSplinePolyRep_init(splinePolyRep, input)
-    call move_alloc(splinePolyRep, repulsive)
+    #:if WITH_CHIMES
+      if (allocated(chimesInp)) then
+        if (.not. isPeriodic) then
+          call error("ChIMES repulsive requires currently periodic boundary conditions")
```
bhourahine, line 6013:
```suggestion
          call error("ChIMES repulsives currently require periodic boundary conditions")
```

src/dftbp/dftbplus/parser.F90
```diff
@@ -7654,6 +7643,35 @@ subroutine readSpinTuning(node, ctrl, nType)
...
+    type(string) :: buffer
+    type(string), allocatable :: searchPath(:)
+    character(:), allocatable :: chimesFile
+
+    call getChild(root, "Chimes", chimes, requested=.false.)
+    if (.not. associated(chimes)) return
+    #:if WITH_CHIMES
+      allocate(chimesRepInput)
+      call getChildValue(chimes, "ParameterFile", buffer, default="chimes.dat")
+      chimesFile = unquote(char(buffer))
```
bhourahine, line 7661: See comment about trailing white space in the buffer with gcc11.2

src/dftbp/include/allocatablelist.fypp
```diff
@@ -0,0 +1,250 @@
...
+#!  Copyright (C) 2006 - 2021  DFTB+ developers group                                              !
+#!                                                                                                 !
+#!  See the LICENSE file for terms of usage and distribution.                                      !
+#!-------------------------------------------------------------------------------------------------!
+#!
+#:mute
+#:if not defined('_ALLOCATABLELIST_FYPP_')
+#:set _ALLOCATABLELIST_FYPP_
+
+#! Disable debugging asserts if unit testing to avoid collision with ASSERT there
```
bhourahine, line 12: Probably should find a workaround for this as a separate PR, as this reduces(?) testing possibilities.

test/src/dftbp/unittests/CMakeLists.txt
```diff
@@ -1,2 +1,3 @@
 set(unit-test-prefix "unittests")
 add_subdirectory(io)
+add_subdirectory(include)
```
bhourahine, line 3: Missing new line at end

utils/get_opt_externals
```diff
@@ -29,7 +29,9 @@ Possible choices for OPTEXT:
   RESTRICTED -- download optional components with stricter license than the
       LGPLv3 license of DFTB+ (typically GPL). The distribution of the compiled
       code will be restricted beyond the LGPL license and governed by the
-      strictest license of the included optional components (typically GPLv3).
+      strictest license of the included optional components (typically GPLv3)
+      or even forbidden in case of conflicting licenses (as with GPLv2). In
```
bhourahine, line 33: Is this left over from the original ChIMES licensing decision, or are there other GPL2 components?

utils/get_opt_externals
```diff
@@ -38,7 +40,8 @@ Possible choices for OPTEXT:
 
   slakos -- Parametrisations for testing DFTB+ [license: CC-BY-SA, UNRESTRICTED]
 
-  dftd3 -- Library for calculating D3 dispersion [license: GPL, RESTRICTED]
+  dftd3 -- Library for calculating D3 dispersion [license: GPL, RESTRICTED,
```
bhourahine, line 43: Extraneous to the purpose of the PR, but harmless

utils/workarounds/multipolyassign.f90
```diff
@@ -0,0 +1,29 @@
+! Repeated polymorhpic assignment to an allocatable may fail, if the allocatable was transfered
```
bhourahine, line 1:
```suggestion
! Repeated polymorphic assignment to an allocatable may fail, if the allocatable was transferred
```

utils/workarounds/multipolyassign.f90
```diff
@@ -0,0 +1,29 @@
...
+! via move_alloc to an other allocatable.
```
bhourahine, line 2:
```suggestion
! via move_alloc to another allocatable.
```

bhourahine (commented):
See comments, but also:

1) chimes/CNOH fails with gcc11.2:
ERROR!
-> Could not find ChIMES parameter file 'slakos/origin/chimes/test_params.CHON.txt  '

(Note the extra white space, which isn't in the hsd file).
parser.F90 line 7661
chimesFile = trim(unquote(char(buffer))
would pass this, but not clear why the buffer has trailing spaces.

2) Is the puml file in the right place in the source tree?

3) twobodyrep.F90 implies there should be a threebodyrep.F90 somewhere.

bhourahine:
Recompiling ran the test OK (the white space is still there after the unquote, but no fail reported this time). No idea what is happening...

aradi:
Note, the relevant PR is the one with the allocatables (#885). Nevertheless, I'll fix the issues mentioned here, and also comment here to the questions, raised. (Sorry for the confusion)

aradi:
As for the general comments:
* Space in the file name string: That is an artifact of our `unquote()` implementation, which leaves the string length unchanged and padds with blanks on the right, instead.
* I thought, it would be best to put the diagram into the folder, where the objects it represents are created.
* Once we have a proper interface for three-body repulsives, we may have such a file. Actually, I'd like to redesign the interface a bit more later, when ChIMES is moved to use the lower level interface (in order to support arbitrary boundary conditions...), but for the moment this should be OK.

doc/dftb+/manual/manual.bib
```diff
@@ -188,7 +188,7 @@ @BOOK{lapack3
       PUBLISHER = {Society for Industrial and Applied Mathematics},
       YEAR = {1999},
       ADDRESS = {Philadelphia, PA},
-      ISBN = {0-89871-447-8 (paperback)} 
+      ISBN = {0-89871-447-8 (paperback)}
```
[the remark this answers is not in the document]
aradi: That's funny, because I can not see them with any of my editors...

src/dftbp/include/allocatablelist.fypp
aradi: You're absolutely right. I've fixed that now.

utils/get_opt_externals
aradi: Remaining from the original ChIMES license, deleted.

aradi:
Continued and fixed in #885.

bhourahine:
> * Space in the file name string: That is an artifact of our `unquote()` implementation, which leaves the string length unchanged and padds with blanks on the right, instead.

No idea why this threw an error (presumably something non-deterministic in the file IO library or OS rather than at the Fortran level). Yes, its clear why unquote() does this, but if we see it again it might be worth switching to the pattern trim(unquote()), as presumably getting something the right length from the function isn't going to work.

PR #900, opened 2021-11-30
Remove unused N2 density matrix routines

bhourahine:
Removes unused O(N^2) density matrix building routines from main code


awvwgk (approved):
What is the reason for having those in the first place? Can they be reactivated?

Other than that, looks good to me.

bhourahine:
@awvwgk This is quite old code, evaluating the density matrix only for atomic/diatomic blocks where S/=0, instead of using SYRK (effectively GEMM) to build the whole matrix. It only becomes competitive for several hundreds of atoms (and for ground state eigen-vector based calculations, the O(N^3) density matrix build is ~5% of the time spent for the eigen-solver anyway).

Issue #846, opened 2021-09-08
Allow setting up k-points from k-point density descriptor


awvwgk:
**Is your feature request related to a problem? Please describe.**

Setting up several DFTB+ inputs in an automated way requires knowledge about the k-point grid. Using one k-point grid (like MP 2x2x2) for all input might be suboptimal for some inputs

**Describe the solution you'd like**

Specify a k-point density and have DFTB+ generate the corresponding MP grid for the respective cell.

**Describe alternatives you've considered**

Extend my input generator to figure out the right k-point grid from the cell information.

aradi:
Very good idea! Maybe, we should couple it with the possibility of providing simplified input for the Monkhorst-Pack scheme:
```
KPointsAndWeights = MonkhorstPack {
  GridDensity [lenghtUnit]  = d1 d2 d3  # Density in along the 3 reciprocal space vectors
  GridDensity [lengthUnit] = d            # alternative form, if unique density is needed in all 3 directions
  GridPoints =  n1 n2 n3                     # Exclusive with GridDensity for simplified input for MP-scheme
  GammaCentered = Yes/No
}
```

PR #903, opened 2021-12-01
Implement keyword for Monkhorst-Pack k-point sampling

awvwgk:
Related #846

### DIFF
--- a/src/dftbp/dftbplus/parser.F90
+++ b/src/dftbp/dftbplus/parser.F90
@@ -2706,13 +2706,14 @@ contains
 
     type(fnode), pointer :: value1, child
     type(string) :: buffer, modifier
-    integer :: ind, ii, jj, kk
-    real(dp) :: coeffsAndShifts(3, 4)
+    integer :: ind, ii, jj, kk, coeffs(3)
+    real(dp) :: coeffsAndShifts(3, 4), shifts(3)
     real(dp) :: rTmp3(3)
     type(TListIntR1) :: li1
     type(TListRealR1) :: lr1
     integer, allocatable :: tmpI1(:)
     real(dp), allocatable :: kpts(:,:)
+    logical :: gammaCentered
 
     call getChildValue(node, "KPointsAndWeights", value1, child=child, &
         &modifier=modifier)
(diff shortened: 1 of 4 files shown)


src/dftbp/dftbplus/parser.F90
```diff
@@ -2744,6 +2745,24 @@ subroutine getEuclideanKSampling(tBadIntegratingKPoints, ctrl, node, geo)
       end if
       ctrl%nKPoint = size(ctrl%kPoint, dim=2)
 
+    case ("monkhorstpack")
+      call getChildValue(value1, "Repeat", coeffs)
+      call getChildValue(value1, "GammaCentered", gammaCentered, .false.)
+      if (gammaCentered) then
+        shifts = 0.0_dp
+      else
+        shifts = merge(0.5_dp, 0.0_dp, mod(coeffs, 2) == 0)
```
bhourahine, line 2754: Not sure if this logic is generally correct, as while most unit cells should have a shift of .5,.5,.5 but hexagonal cases are need to be shifted by 0,0,.5 to retain the symmetry for even grids

bhourahine (commented):
Might be an issue for the automatic shifts

src/dftbp/dftbplus/parser.F90
awvwgk: Means we have to detect the space group of the lattice first. Guess that is simple enough to do for the hexagonal case.

bhourahine:
Not sure where they are getting this from, but the Vasp [documentation](https://www.vasp.at/wiki/index.php/KPOINTS) is also suggesting un-shifted grids that include Γ should be used for FCC and FC-orthorhomic (might be due to their symmetrization method). My understanding is the hexagonal case should only be shifted along c* (they say no shift at all).

awvwgk:
Crystal symmetry might become difficult. We could use spglib, but maybe it is simple enough to implement at least the crystal systems ourselves to get some basic symmetry properties in. I tried the hexagonal one, but haven't tested it much yet.

bhourahine:
Probably a good idea to have an alternative keyword to manually set the shift, just in case (another use case for this would be to make sure there is a sampling point on top of the Dirac cone in materials like graphene).

PR #904, opened 2021-12-01
Fix sign of charge model 5 contributions

### LINKED ISSUE #805: Sign error in CM5 charges
**Describe the bug**

DFTB+ passes populations instead of partial charges to `updateCharges` in `addChargePotential`. CM5 charges are defined for partial charges rather than populations, but the sign convention was overlooked in the initial implementation.

**Expected behaviour**

Add negative CM5 correction to populations.

**Additional context**

Document and/or rename the `updateCharges` routines.

awvwgk:
- added as partial charge rather than charge population correction

Closes #805


bhourahine:
@awvwgk do you want to open another issue about clarifying the updateCharges naming, if this closes #805 ?

bhourahine (approved):
LGTM

PR #907, opened 2021-12-06
Getfileid4

helmutwecke:
This is my new attempt for the Deactivate getfileid pull request. In that previous pull request had some problems while changing the logic structure related to the file ID numbers. 

In this new pull request I deactivated getfileid completely without messing up the logic structure at linresp. 

@aradi Please take a look at this to see if everything is fine, so that I can proceed by rebasing on the master branch


helmutwecke:
I am currently trying to eliminate the passing of file ID numbers for the logical structure.

aradi:
@helmutwecke Could you please delete the file `timedep/rslinresp.F90`, commit the change and then squash all three commits into one and force-push it to here? You can then also mark the PR as ready for review.

I'd then make my changes on the squashed version without the superfluous files in it....

helmutwecke:
@aradi The changes you requested are done. Please tell me if I can help again.

bhourahine (commented):
The convention that would be consistent with the rest of the code for logical variables would be something like
isXPlusYWritten (the old convention was tXPlusYWritten, but this can be confused with our convention for Fortran types), instead of your use of fd* or write*.

If you change this, then the various integer :: newfd* could be renamed fd*.

Also, these variables are logical flags, not (integer) file units, so the comments in linearresp.F90 should be changed.

aradi:
Note: Codecov failure due to coverage decrease seems to come from the fact, that the affected files contain less lines now, so that the uncovered have a higher weight. (All deleted and newly added lines had/have full coverage).

@helmutwecke Commit 9c7a7cf contains my fixes on top of your PR, have a detailed look at it. Commit 12e0641 is a further extension, as the global file unit based registry can not be used with the `newunit` mechanism, if a file is opened and closed repeatedly (as done with `fdMd` and `fdDetailedOut`.

PR #908, opened 2021-12-07
Fix api

### LINKED ISSUE #834: Python API segfault
**Describe the bug**

API breaking changes and Python API segfaults due to [50158a5](https://github.com/dftbplus/dftbplus/pull/686/commits/50158a54f5c075b9bd109d7575ead7ca71ffd93c)

**To Reproduce**
Appears for current Python regression tests as a segfault

**Expected behaviour**

Tests run as expected.

**Additional context**

Previously discussed, but should be addressed. Also, shows CI regression should be running this.

aradi:
* Turns back the API to the original form without `atomList` argument in the `dftbp_process_input()` call
* Fixes the library paths in the pyapi tests, so that they work properly.
* Fixes several problems experienced during shared library builds
* Introduces new CMake option `ENABLE_DYNAMIC_LOADING`, which governs `dlopen()` compatibiliy. `WITH_PYTHON=True` installs dptools only, if `ENABLE_DYNAMIC_LOADING` is `False`, otherwise the Python API is also installed.

The setting of the species via API should be implemented in an other PR using a different method (allowing C interface to build up and manipulate an HSD-tree).

Closes #834.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,6 +18,7 @@ env:
     -DFYPP_FLAGS='-n;-DTRAVIS'
     -DWITH_UNIT_TESTS=true
     -DWITH_CHIMES=true
+    -DWITH_PYTHON=true
 
 jobs:
   gcc-build:
@@ -128,7 +129,7 @@ jobs:
     - name: Set extra CMake flags (Linux)
       if: contains(matrix.os, 'ubuntu')
       run: |
-        echo "CMAKE_OPTIONS=${CMAKE_OPTIONS} -DBUILD_SHARED_LIBS=true -DWITH_PYTHON=true" >> $GITHUB_ENV
+        echo "CMAKE_OPTIONS=${CMAKE_OPTIONS} -DENABLE_DYNAMIC_LOADING=true" >> $GITHUB_ENV
 
     - name: Configure build
       run: >-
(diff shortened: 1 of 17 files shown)


bhourahine (commented):
LGTM, but running ctest fails all of the Python regressions for MKL (buildbot for intel2021 and other system).

aradi:
I'll have a look.

aradi:
Oh, the solution is really trivial. You just have to run
```
env LD_PRELOAD=$MKLROOT/lib/intel64/libmkl_def.so:$MKLROOT/lib/intel64/libmkl_core.so:$MKLROOT/lib/intel64libmkl_intel_lp64.so:$MKLROOT/lib/intel64/libmkl_intel_thread.so:/opt/intel/2019/compilers_and_libraries_2019.5.281/linux/compiler/lib/intel64/libiomp5.so  ctest
```
then it works. It seems to be an MKL-loading problem, as explained in https://community.intel.com/t5/Intel-oneAPI-Math-Kernel-Library/mkl-fails-to-load/m-p/1155541/highlight/true#M27513 . Do not ask me, why one needs it...

awvwgk:
The issue is more how DFTB+ links MKL. To load the libdftbplus.so with dlopen correctly (*i.e.* via ctypes in Python) we cannot link against the usual MKL libraries but must use the `mkl_rt` library instead. I think this is described in the linked thread as option 2.

aradi:
Thanks, indeed that is the solution! If I use that library, it works. From CMake 3.17, we can even ask CMake politely to search for the dynamical version. For that, we would have to bump the minimal CMake version from 3.16 to 3.17, though...

awvwgk (approved):

PR #910, opened 2021-12-10
Test for serial code init with comm world

bhourahine:
Doesn't seem to be a way to test for a serial library init from
a parallel driving code if it doesn't pass a comm world.


awvwgk (approved):
Looks good to me.

PR #913, opened 2021-12-12
Stage 21.2

aradi:
Release candidate 21.2.

Note for the review: Only commit 3823313 is new, everything else comes from master.


awvwgk (approved):

awvwgk:
What a huge diff from all the refactors and new features. I'm really looking forward to wrap this version up.

Getting to the approve button in the review tab turned out to be a bit difficult since so many files are loaded slowing down the UI :sweat_smile:.

bhourahine (approved):
There are a couple of changes on master which aren't here but shouldn't get lost, they can go into the next release if necessary.

PR #917, opened 2021-12-17
MPI loops only over atoms stored on processor

bhourahine:
Stores indexing for atoms that correspond to local parts of BLACS matrices.

Should improve parallel performance on larger numbers of processors (marginal but a couple of percent speed-up present on tests with 4 processors)

- [x] BLACS unpack routines in main code
- [ ] BLACS pack routines in main code
- [ ] BLACS pack/unpack outside of main code
- [ ] Other BLACS operations on orbitals
- [ ] Trap for BLACS vs CSR
- [ ] CSR equivalent indexing


bhourahine:
Probably not noticably faster until multi-thousand atom systems with SCC loops (indexing costs about the same as the usual non-scc unpack/pack wastage over non-present atoms)

PR #922, opened 2022-01-05
External testing for non-tagged files

bhourahine:
In practice, grep for broken patterns in files and remove autotest.tag if found. Currently just looking for NaN values in
detailed.out and output, but would have caught the bug that #921 resolves for at least one tested compiler.


bhourahine:
#923 should be merged first, as this fixes an uninitialised variable that trips these tests with one compiler and test case.

test/app/dftb+/bin/globalpostrun.sh
```diff
@@ -0,0 +1,27 @@
...
+#  See the LICENSE file for terms of usage and distribution.                   #
+#------------------------------------------------------------------------------#
+
+############################################################################
+#
+#  Script for automated testing outside of tagged files
+#
+############################################################################
+
+if grep -l NaN output; then
```
awvwgk, line 15: Check for whole words here?

test/app/dftb+/bin/globalpostrun.sh
```diff
@@ -0,0 +1,27 @@
...
+    rm autotest.tag
+    exit 1
+else
+    echo "didn't run into NaN in output"
+fi
+
+if grep -l NaN detailed.out; then
```
awvwgk, line 22: Same here, better check for a whole word.

awvwgk (changes requested):

test/app/dftb+/bin/globalpostrun.sh
bhourahine: How about 
grep -l "\<NaN\>"
?

bhourahine: Unfortunately only on linux, not BSD derived (my construct might also fail there): https://stackoverflow.com/questions/9792702/does-bash-support-word-boundary-regular-expressions

awvwgk: How many DFTB+ users do we have on BSD anyway? cc @aradi ;)

aradi: I think, we should probably still keep compatibility, if possible, in order to avoid bug reports from BSD-packagers. :wink:

bhourahine:
Merge after #1055

bhourahine:
Rebased after fix from #1055 merged.

awvwgk (approved):
Let's try whether this works as expected. Ready to get merged.

PR #924, opened 2022-01-06
Check hermitian failure in matrices

bhourahine:
Also overhaul some of the error propagation in the main code to
switch from my old macro to the newer one.

Note, one of the helical cases fails the hermiticity check.


aradi (approved):
I'm in general fine with the PR, but could you please move those large debug-blocks into an internal subroutine (so that you don't need pass arguments) before merging? Otherwise, it unnecessary bloats the part with the actual algorithm.

bhourahine:
> I'm in general fine with the PR, but could you please move those large debug-blocks into an internal subroutine (so that you don't need pass arguments) before merging? Otherwise, it unnecessary bloats the part with the actual algorithm.

Code reuse for the H and S cases would suggest it's better to actually pass to a separate subroutine (see f2e5f1c which is now squashed). But if you really want internal subroutines, will change.

PR #926, opened 2022-01-11
Add workaround for ifort 202x compiler bug

aradi:
Offers workaround for the iforts ICE (reported in https://community.intel.com/t5/Intel-Fortran-Compiler/ICE-for-simple-matmul-expression-when-checks-are-turned-on/m-p/1350570#M159304)


bhourahine (approved):
Looks fine, but please add a comment that this is a workaround.

PR #930, opened 2022-01-15
DFTB sparse matrices with dense operations

bhourahine:
Expanded boundary condition derived type, for use with helical transforms in sparse blas and for later PRs for sparse/dense transforms. Move rotation for vectors to type.

symv operations
Complex or real vector with real(/Gamma), k-points and helical sparse multiplications with vectors

symm operations
Complex, real or helical matrices with real and k-point sparse multiplications

Start of mpi parallel sparse with dense by reorganizing BLACS matrices to put entire rows on processors, then enabling distributed use of serial GEMM/GEMV

Note: coverage will reduce as most of the routines are not in use yet.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,12 +1,12 @@
 [submodule "external/mpifx/origin"]
-    path = external/mpifx/origin
-    url = https://github.com/dftbplus/mpifx.git
+	path = external/mpifx/origin
+	url = https://github.com/dftbplus/mpifx.git
 [submodule "external/libnegf/origin"]
 	path = external/libnegf/origin
 	url = https://github.com/libnegf/libnegf.git
 [submodule "external/scalapackfx/origin"]
-    path = external/scalapackfx/origin
-    url = https://github.com/dftbplus/scalapackfx.git
+	path = external/scalapackfx/origin
+	url = https://github.com/dftbplus/scalapackfx.git
 [submodule "external/mbd/origin"]
 	path = external/mbd/origin
 	url = https://github.com/libmbd/libmbd.git
--- a/src/dftbp/dftb/boundarycond.F90
+++ b/src/dftbp/dftb/boundarycond.F90
@@ -6,23 +6,28 @@
 !--------------------------------------------------------------------------------------------------!
 
 #:include 'common.fypp'
+#:include 'error.fypp'
 
 !> Contains geometrical boundary condition information on the calculation
 module dftbp_dftb_boundarycond
   use dftbp_common_accuracy, only : dp
+  use dftbp_common_constants, only : pi
+  use dftbp_common_status, only : TStatus
+  use dftbp_math_angmomentum, only : rotateZ
+  use dftbp_math_quaternions, only : rotate3
   use dftbp_type_commontypes, only : TOrbitals
   implicit none
 
   private
-  public :: zAxis
-  public :: boundaryConditions
+  public :: zAxis, boundaryConditions, TBoundaryConditions, TBoundaryConditions_init
 
+#:set FLAVOURS = [('complex', 'cplx'), ('real', 'real')]
 
-  !> z direction vector for rotation
+  !> z direction vector for screw rotation
   real(dp), parameter :: zAxis(3) = [0.0_dp, 0.0_dp, 1.0_dp]
 
 
-  !> Enumerator containign possible boundary conditions
+  !> Enumerator containing possible boundary conditions
   type :: TBoundaryConditionEnum_
 
     !> Unknown/undefined boundary condition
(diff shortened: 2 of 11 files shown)


.gitmodules
```diff
@@ -6,7 +6,7 @@
 	url = https://github.com/libnegf/libnegf.git
 [submodule "external/scalapackfx/origin"]
     path = external/scalapackfx/origin
-    url = https://github.com/dftbplus/scalapackfx.git
+    url = https://github.com/bhourahine/scalapackfx.git
```
awvwgk, line 9: Are the ScaLapackFx changes upstreamed already?

bhourahine: It is now

src/dftbp/dftb/boundarycond.F90
```diff
@@ -44,4 +49,207 @@ module dftbp_dftb_boundarycond
...
+
+    !> Boundary condition choice
+    integer, intent(in) :: iBoundaryCondition
+
+    !> Status of routine
+    type(TStatus), intent(out) :: errStatus
+
+    if (.not. any([boundaryConditions%cluster, boundaryConditions%pbc3d,&
+        & boundaryConditions%helical] == iBoundaryCondition)) then
+      @:RAISE_ERROR(errStatus, -1, "Unknown boundary condition specified")
```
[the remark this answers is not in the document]
bhourahine: There is a plan to propagating errors up to the API (so avoiding code halts inside the library which obviously take down the calling code).

.gitmodules
bhourahine: And just remembered to update the submodule path and hash as well...

src/dftbp/dftbplus/initprogram.F90
```diff
@@ -4790,6 +4804,21 @@ subroutine allocateDenseMatrices(this, env)
       allocate(this%eigVecsReal(nLocalRows, nLocalCols, nLocalKS))
     end if
 
+  #:if WITH_SCALAPACK
+
+    if (this%isSparseReorderRequired) then
+      call scalafx_getlocalshape(env%blacs%rowOrbitalGrid, this%denseDesc%blacsColumnSqr,&
+          & nLocalRows, nLocalCols)
+      write(*,*)'HERE0', nLocalRows, nLocalCols, nLocalKS
```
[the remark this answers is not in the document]
bhourahine: Yes, thanks for spotting it, now removed.

bhourahine:
> I went through your changes from a developer's point of view, however I can hardly judge the physics behind it. The code looks fine to me.

Thanks, I was interested in you opinion about the BLACS re-distribution. The grid block size seems to have to be unusual to get a balanced distribution of entire rows onto individual processors.

bhourahine:
Now with correctly updated submodule path and hash.

bhourahine:
Rebased for on main for changes required to run on buildbot server (MPI+GPU, which are irrelevant to this PR).

src/dftbp/math/sparseblas.F90
```diff
@@ -0,0 +1,1183 @@
...
+          end do
+        end if
+        y(iy:jy) = y(iy:jy) + alphaTmp * matmul(sqrTmp(:nOrb2,:nOrb1),x(ix:jx))
+        ! other triangle due to symmetry of matrix
+        if (iAtom1 /= iAtom2f) then
+          y(ix:jx) = y(ix:jx) + alphaTmp * matmul(transpose(sqrTmp(:nOrb2,:nOrb1)),x(iy:jy))
+        end if
+      end do
+    end do
+    !$OMP  END PARALLEL DO
```
terminationshock, line 160: This code block seems to be duplicated several times. Can you somehow reuse it?

src/dftbp/dftbplus/initprogram.F90
```diff
@@ -4894,6 +4919,9 @@ subroutine getDenseDescBlacs(env, rowBlock, colBlock, denseDesc)
     call scalafx_getdescriptor(env%blacs%orbitalGrid, nn, nn, rowBlock, colBlock,&
         & denseDesc%blacsOrbSqr)
 
+    call scalafx_getdescriptor(env%blacs%rowOrbitalGrid, nn, nn, rowBlock * colBlock, 1,&
+        & denseDesc%blacsColumnSqr)
```
terminationshock, line 4923: Is `rowBlock * colBlock` equal to the number of columns? From my understanding, you will otherwise not get an entire row on the processor.
What happens anyway if the number of rows is different from the number of processors?

bhourahine: Well spotted, yes this should be 
```suggestion
    call scalafx_getdescriptor(env%blacs%rowOrbitalGrid, nn, nn, nn, 1,&
        & denseDesc%blacsColumnSqr)
```
The [tester](https://github.com/dftbplus/scalapackfx/blob/main/test/test_gemr2d.f90) for scalapackfx was correct.

bhourahine: If nRow /= nProc then some processors will have fewer rows stored than others (assuming n_matrix > nProc, which is always the case at the moment, excepting for the use of processor groups for k-points etc). The default of choosing a square BLACS processor grid doesn't seem to balance this well (probably NUMROC only really cares about cyclic block distribution, not the effective row cyclic here, at least for the ScaLAPACK implementations I've tested). Hence the psudo-1D processor grid.

src/dftbp/math/sparseblas.F90
bhourahine: The _bc_ cases can also do the job of the other routines, but there will be a performance penalty. This will need benchmarking to decide once they get used for more operations in the code in a future PR.
Possibly the k-point case can also be used for non-k-point as well (but will need some refactoring in the main code to make sure all of the extra indexing arrays are consistently either allocated or not).

PR #933, opened 2022-01-16
Fix DESTDIR support for Python components

### LINKED ISSUE #931: Fails to install the Python binding into the stage directory
It prints this error
```
error: could not create '//usr/local/lib/python3.8/site-packages/__init__.py': Permission denied
```
and then succeeds anyway.

The ```install``` target should respect the ```DESTDIR``` variable which contains stage directory.

Version: 21.2
OS: FreeBSD 13

awvwgk:
Closes #931


bhourahine (commented):
`\$ENV{DESTDIR}` seems to be the usual solution, or does this not work here?

awvwgk:
`$ENV{DESTDIR}` will access the variable at configure time, but `DESTDIR` is usually set at install time.

bhourahine (approved):
Ready to merge

PR #936, opened 2022-01-16
Reduce explicit dependencies on transient dependencies

awvwgk:
- tblite can find dftd4
- dftd4 can find multicharge
- tblite/dftd4/s-dftd3 can find mstore


awvwgk:
We'll keep the current state for now.

PR #938, opened 2022-01-17
Always use env in shebang

awvwgk:
Not all platforms have `/bin/bash` and `/bin/sh`.

### DIFF
--- a/test/app/dftb+/bin/autotest2
+++ b/test/app/dftb+/bin/autotest2
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 #------------------------------------------------------------------------------#
 #  DFTB+: general package for performing fast atomistic simulations            #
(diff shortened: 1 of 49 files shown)


test/app/dftb+/bin/autotest2
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env sh
```
bhourahine, line 1: Specifically for the sh scripts there might be an argument for either 1) keeping as /bin/sh or 2) switching to /usr/bin/env bash 
The original /bin/sh is explicitly getting POSIX compliance instead of whatever env serves up. The bash cases definitely should be via env.

awvwgk: Does POSIX define the location of `sh`? In case a user provided POSIX compliant shell is used we want to pick this up.

awvwgk: Done, using `bash` and `python3` now everywhere.

PR #939, opened 2022-01-17
API nr of kpts

bhourahine:
Adds onto #929 as a pre-requirement to a future PR for the density matrix return via the API (as will be needed to size the array).


bhourahine:
As per devel discussion 26/01/22, revert comment about (redundant) species routine and renamed Fortran returns as per convention.

src/dftbp/api/mm/dftbplus.h
```diff
@@ -258,6 +258,16 @@ void dftbp_set_coords_lattice_origin(DftbPlus *instance, const double *coords,
...
 
 
+/**
+ * Queries the nr. of k-points in the system.
+ *
+ * \param[inout] instance Handler of the DFTB+ instance.
+ *
+ * \return Nr. of k-points
+ */
+int dftbp_get_nr_kpoints(DftbPlus *instance);
```
aradi, line 268: This way, the name C-name is inconsistent with the Fortran one....
```suggestion
int dftbp_nr_kpoints(DftbPlus *instance);
```
We really should somehow make the names in the Fortran an C interface consistent, probably I'll make an additional separate PR on this.

src/dftbp/api/mm/dftbplus.h
```diff
@@ -278,6 +288,26 @@ void dftbp_get_energy(DftbPlus *instance, double *mermin_energy);
...
 
 
+/**
+ * Queries the number of basis functions for each atom in current geometry.
+ *
+ * \param[inout] instance Handler of the DFTB+ instance.
+ *
+ * \param[out] nOrbitals number of orbitals on each atom
+ */
+void dftbp_get_nr_basisfns(DftbPlus *instance, int *nOrbitals);
```
aradi, line 298:
```suggestion
void dftbp_get_nr_orbitals(DftbPlus *instance, int *nOrbitals);
```

aradi (approved):
I am fine, but see comments on naming.

src/dftbp/api/mm/dftbplus.h
bhourahine: Probably _get_nr_basisfns is more accurate in the case of non-minimal basis (as they don't neccessarily correspond to atomic orbitals. and also nr_orbitals could be mistaned with total matrix order), but I take the point in terms of matching internal variable names.

src/dftbp/api/mm/dftbplus.h
bhourahine: Changed

PR #941, opened 2022-01-19
Use atomic units for masses rather than amu

### LINKED ISSUE #940: Incorrect masses when using TBlite.
**Describe the bug**
Using tblite with the velocity verlet integrator to do simple MD calculations on a H/B/N/C/Fe system I found that the system tended to fly apart due to unexpectedly high velocities. Using to a simple Fe32 system I got the same problem.

Can be reproduced with this dftb_in.hsd

[dftb_in.hsd.txt](https://github.com/dftbplus/dftbplus/files/7893583/dftb_in.hsd.txt)

**Additional context**
By doing a little digging I found that the DFTB+ needs a.m.u masses when calculating accelerations from forces but the internal calculator for units DFTB+ uses when using TBlite is in au (I think theyre listed in common/atomicmass.F90) so the masses when using tblite are 3 orders or magnitude smaller than they should be. I get something closer to expected velocities and dynamics by modifying line 1497 in the dftbplus/initprogram.F90 file from: 

this%speciesMass(:) = getAtomicMass(this%speciesName)

to: 

this%speciesMass(:) = getAtomicMass(this%speciesName)*amu__au

I don't know what other features this will affect and more testing needs to be done to see if it actually fixes the issue. I'll also note that the masses are still off with the slko mass conversion ratio being 1822.776796 while amu__au=1822.888485 so the masses the slko and tblite use still are not the same.

awvwgk:
Fixes #940

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,8 @@ Changed
 Fixed
 -----
 
+- incorrect atomic mass unit for xTB calculations
+
 
 
 21.2 (2021-12-13)
--- a/app/phonons/initphonons.F90
+++ b/app/phonons/initphonons.F90
@@ -767,8 +767,8 @@ contains
       defmass = getAtomicMass(trim(geo%speciesNames(iSp)))
       call getChildValue(value, geo%speciesNames(iSp), mass, defmass,&
                &modifier=modif, child= child2)
-      speciesMass(iSp) = mass * amu__au
-      write(stdOut,*) trim(geo%speciesNames(iSp)),": ", mass, "amu", &
+      speciesMass(iSp) = mass
+      write(stdOut,*) trim(geo%speciesNames(iSp)),": ", mass/amu__au, "amu", &
             &SpeciesMass(iSp),"a.u."
     end do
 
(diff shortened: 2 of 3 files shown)


aradi:
Note, that `initphonons.F90` must also be adapted, as it also calls `getAtomicMass()`.

bhourahine:
Might also be an idea to have a look at app/modes/modes.F90 as that probably should be calling `getAtomicMass()` now.

awvwgk:
It probably could use `getAtomicMass` instead of requiring SK files, but `modes` currently doesn't use this feature

```
❯ rg getAtomicMass -w -i
src/dftbp/dftbplus/initprogram.F90
14:  use dftbp_common_atomicmass, only : getAtomicMass
1497:      this%speciesMass(:) = getAtomicMass(this%speciesName)

src/dftbp/common/atomicmass.F90
17:  public :: getAtomicMass
21:  interface getAtomicMass
24:  end interface getAtomicMass
73:    mass = getAtomicMass(symbolToNumber(symbol))

app/phonons/initphonons.F90
767:      defmass = getAtomicMass(trim(geo%speciesNames(iSp)))
```

PR #944, opened 2022-01-26
Add CM5 charges to the API

steinmig:
Changes:
 - Moved CM5 defaults from parser to CM5
 - Made update CM5 methods public
 - Added CM5 charges API call with input as argument, which allows to get the charges without specifying them in the input
 
Caveat: I am usually not programming in Fortran and it is not unlikely that I broke some design principles.
 
 ToDo:
 - [x] Correct unittest expected values
 - [x] Revert CM5 changes
 - [x] Remove input from API call
 - [x] Change Unittest inputs to ensure CM5 analysis


bhourahine:
As per devel discussion 26/01/22, revert comment about (redundant) species routine and renamed Fortran returns as per convention.

bhourahine:
Apologies, commented on wrong PR.

src/dftbp/solvation/solvparser.F90
```diff
@@ -420,7 +420,7 @@ subroutine readCM5(node, input, geo)
     real(dp) :: conv
     real(dp), allocatable :: atomicRadDefault(:)
 
-    call getChildValue(node, "Alpha", input%alpha, 2.474_dp/AA__Bohr, &
```
bhourahine, line 423: The usual convention we use is for internal code variables to be in atomic units, and this quantity is an inverse length. The original default value is a conversion is from 1/AA to 1/Bohr (the modifier on line 425 is also converting in this way if there is a user supplied value).

src/dftbp/solvation/solvparser.F90
```diff
@@ -447,7 +447,7 @@ subroutine readCM5(node, input, geo)
     end if
     input%atomicRad(:) = input%atomicRad * conv
 
-    call getChildValue(node, "Cutoff", input%rCutoff, 30.0_dp, &
```
bhourahine, line 450: Is there a reason for removing the default of 30 Bohr?

steinmig: I didn't remove it, I moved the default definition from here within the parser to the definition of `TCM5Input` in `cm5.f90`, since this allows to construct the cm5 input with defaults without parsing a file and it made sense to me in a general sense, but I saw that you have many defaults in the parser so I don't know how you would want to handle that.

Alternatively one would need to duplicate the defaults in the cm5 construction, which I tried to avoid, since it is an invitation for diverging code in the future. 

Previously the cm5 model could not be constructed without parsing something (due to these defaults). I implemented that so the API call does not fail, if the input is missing the CM5 keyword

src/dftbp/solvation/solvparser.F90
steinmig: As with the `rCutoff`, I simply moved this value. I guess if you think this could introduce a unit error in the future if someone changes the default in cm5.f90, we could change it here already :shrug:

awvwgk: We don't want mixed initialization, the input data is defined completely in the parser, this way we have a single source of truth. Adding part of the default values to the derived type would scatter the initialization. Generally, for input data types the parser is place where they are initialized completely, while for container types the `*_init` routine contains all initialization logic.

steinmig: Hmm, I thought the CM5 logic is relatively close to the logic of the Born model, where the defaults are also defined in a separate type and passed to the parser.

If you want to avoid this for the CM5, I can revert the logic in the parser and cm5 and simply throw an error if the API method is called with a missing CM5 input

awvwgk: So far DFTB+ does not allow to alter the setup easily from the API, since most logic is embedded in the parser. It would be practical to have the possibility to create and attach containers after the fact to the DFTB+ instance to get additional properties from the API. This requires them to be easy to construct without the parser and self-contained that they needn't be present in the actual calculation (or actually impact the outcome of the calculation).

steinmig: I guess that's what I changed the CM5 to^^

Anyway I don't require to have the case of the missing input covered by DFTB+ for my application since my call already ensures it, so I don't mind either way

awvwgk: I don't know how reliable it is to reparse the input and rerun from an API call (might drop some data or perform redundant calculation, but haven't checked). The property you are aiming for is rather cheap compared to a full DFTB calculation.

Also, to get the reparsing and rerunning a lot of code duplication is required, scattering the additional initializations in the code base makes it more difficult to change the CM5 routines in the future.

src/dftbp/api/mm/capi.F90
```diff
@@ -482,6 +484,43 @@ subroutine c_DftbPlus_getGrossCharges(handler, atomCharges)&
 
   end subroutine c_DftbPlus_getGrossCharges
 
+  !> Obtain CM5 charges
+  subroutine c_DftbPlus_getCM5Charges(handler, inputHandler, atomCharges)&
+      & bind(C, name='dftbp_get_cm5_charges')
```
aradi, line 489: As others noted (thanks!), this is something we should discuss in detail as it breaks the logic of the current API. The current logic is to build the input structure (from file or in Fortran also via API), use it to initialize the DFTB+ calculator, and then destroy it. 

Miguel (@steinmig), is there any reason, why we can not expect the consumer of our API to set up the need (and the parameters) for the CM5 charges already in the input? Is there any scenario you can think about, where the possibility of an "on-the-fly" decisions after the initialization is really needed?

src/dftbp/solvation/solvparser.F90
aradi: I absolutely agree here (see also my other comment): We are currently not well prepared for a reparsing of the input...

src/dftbp/solvation/solvparser.F90
aradi: I'd like to add additionally, that IMO this is not valid Fortran. :wink:  You are passing the same actual argument for two different dummy arguments, one them being `intent(out)`, the other being `intent(in)`. The compiler could in theory reset/nullify the `intent(out)` at entry at the subroutine, but that would change the value of your `intent(in)` argument, where you passed the same variable.
If we really wanted this (in order to ensure, that you use the default value stored in the default initialization of the derived type) you would have to do:
```
call getChildValue(node, "Cutoff", input%rCutoff, (input%rCutoff), &
```
which would create an expression (a copy) for the `intent(in)` argument.
But I personally would also opt for the more explicit original initialization here, I think, that is less error-prone.

aradi (commented):
We should probably indeed have a short discussion about the strategy to follow (see comments below). I'd vote for keeping the original API logic. However, we would then have to allow for a graceful fail, when CM5 charges are requested through the API from a calculator, which was initialized without the appropriate CM5 charge options in the input.

src/dftbp/solvation/solvparser.F90
steinmig: Ah tricky Fortran being tricky^^

Thanks for pointing it out

src/dftbp/api/mm/capi.F90
steinmig: @aradi Sorry I forgot to write here again, this is finished from my side

awvwgk (approved):
Looks good to me now.

src/dftbp/dftbplus/mainapi.F90
```diff
@@ -183,6 +183,39 @@ subroutine getGrossCharges(env, main, atomCharges)
...
+
+    call recalcGeometry(env, main)
+    if (.not. allocated(main%cm5Cont%cm5)) then
+      call error("CM5 could not be calculated.")
+    end if
+    atomCharges(:) = sum(main%q0(:, :, 1) - main%qOutput(:, :, 1), dim=1) + main%cm5Cont%cm5
+
+    !> Pass to the charges of the excited state if relevant
+    if (main%isLinResp) then
+      atomCharges(:) = atomCharges(:) + main%dQAtomEx(:)
```
aradi, line 214: I am wondering, why you add `main%dQAtomEx` here?

steinmig: That's a copy/paste mistake from the gross_atom_charges API call. Are there CM5 charges for the exited state available or should we just omit this?

bhourahine: Not at the moment (but as they are additive, should be simple to implement).

bhourahine:
@steinmig @aradi who want to resolve the couple of conflicts against main, and anything left to do before merging?

aradi:
@bhourahine I'll have a look at it.

steinmig:
@bhourahine @aradi I just resolved the conflicts with main, but I fear I won't have time in the next 2 weeks to add the excited state support

aradi:
@steinmig Thanks! (I've pushed a commit with some whitespace corrections). As for the excited states charges: Aren't they already covered by the current implementation? As the CM5 correction is purely geometry dependent, there is nothing extra to do for the excited state charges. Or do I overlook something?

PR #946, opened 2022-01-26
Handle JSON Fortran dependency in mctc-lib

awvwgk:
- disable JSON Fortran dependency if mctc-lib is included as subproject
- JSON Fortran is only needed for IO interface, which is not used in DFTB+


bhourahine (approved):
LGTM

PR #947, opened 2022-01-26
Gamma point periodic range separation

bhourahine:
Manual rebase of #702 on current main branch.

Not fully tested and needs further development in later PRs (check for cell size, supercell evaluation of exchange if needed, stress tensor, use with Casida routines, ...).

- [x] gamma point expressions
- [x] documentation
- [ ] test forces
- [ ] regression tests (including those checked against #702 )

### DIFF
--- a/doc/dftb+/manual/dftbp.tex
+++ b/doc/dftb+/manual/dftbp.tex
@@ -4300,8 +4300,14 @@ has the following option:
       \is{Thresholded} block.
     \end{description}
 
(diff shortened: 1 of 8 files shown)


vanderhe:
Thanks for splitting the PR, I'll check the forces against [calcderivs](https://github.com/dftbplus/dftbplus/blob/main/utils/numcheck/calcderivs) (probably next week).

PR #950, opened 2022-02-02
Forces gamma rs

bhourahine:
Central cell shifting to evaluate gamma point forces for range separation in periodic geometries


bhourahine:
Based on #947 so includes it.

vanderhe:
Seems to improve agreement of analytical/numerical forces, at least for my test cases, by about one order of magnitude, i.e. ~5e-03a.u. (#947) vs. ~3e-04a.u. (#950).

PR #952, opened 2022-02-03
Fix minNeighDist to minNeighDist2 for neighbours

inseonglee:
Current code compares neigh%neighDist2 and minNeighDist.

### DIFF
--- a/src/dftbp/dftb/periodic.F90
+++ b/src/dftbp/dftb/periodic.F90
@@ -563,7 +563,7 @@ contains
     ! check for atoms on top of each other
     do iAtom1 = 1, nAtom
       do nn1 = 1, neigh%nNeighbour(iAtom1)
-        if (neigh%neighDist2(nn1, iAtom1) < minNeighDist) then
+        if (neigh%neighDist2(nn1, iAtom1) < minNeighDist2) then
           iAtom2 = img2CentCell(neigh%iNeighbour(nn1, iAtom1))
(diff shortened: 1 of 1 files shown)


CHANGELOG.rst
```diff
@@ -23,6 +23,8 @@ Fixed
 
 - incorrect atomic mass unit for xTB calculations
 
+- Check of minimum distances from neighbours
+
```
aradi, line 27: I'd suggest not to mention this in the change log, as the change is basically invisible for the users. (The only thing they would see, that they have to put the atoms even closer as so far, in order to trigger an error message from DFTB+).

aradi (approved):
Thx! I am fine with the PR, just see the one comment about the change log.

CHANGELOG.rst
inseonglee: @aradi I removed the change log.

inseonglee: This bug exists in version 21.1 and 21.2 (I did not check the version 20.2).

CHANGELOG.rst
```diff
@@ -24,7 +24,6 @@ Fixed
 - incorrect atomic mass unit for xTB calculations
 
 
-
```
bhourahine, line 27: Could you add the blank line back here and then squash the commits in this branch down to a single change?

CHANGELOG.rst
bhourahine: The bug probably has been here for 10+ years. It will trigger for atoms at < 0.01 Bohr separation, so probably never manifested in real systems.

inseonglee:
I reset and repush single commit.

PR #955, opened 2022-02-09
Allocate linresp input only on demand

### LINKED ISSUE #945: Inconsistent option "EigenvectorsAsText = No" in dftb_pin.hsd
**Describe the bug**
The contents of the file "dftb_pin.hsd" seem to change, even though the original file "dftb_in.hsd" stays the same.
Specifically, the option "EigenvectorsAsText = No" seems to sometimes appear in "dftb_pin.hsd", and sometimes not.

**To Reproduce**
1) Download the stable source from https://dftbplus.org/download/dftb-stable
2) Compile using the instructions given at https://github.com/dftbplus/dftbplus
3) Use the dftb_in.hsd-file given at https://dftbplus-recipes.readthedocs.io/en/latest/basics/firstcalc.html
4) Run the program in _build/app/dftb+ like `./dftb+ && grep "Eigenvectors" dftb_pin.hsd`

**Additional Info**
1. Input-file see above, the only thing I changed is the location of the SK-files
2. DFTB+ Version 21.2, compiled from source (see above), gfortran 9.3.0, cmake 3.16.3 on Ubuntu 20.4.3 LTS
3. For me, sometimes the option "EigenvectorsAsText = No" appears in the dftb_pin.hsd-file, and sometimes it does not. It may take a few runs to see it happening.

I just recently started working with dftb+, so I just might be missing something, apologies if that is the case. The behaviour just seems strange to me. I did not make any changes to the code.

aradi:
Rationalize the usage of the linresp input type.

Closes #945

### DIFF
--- a/src/dftbp/dftbplus/initprogram.F90
+++ b/src/dftbp/dftbplus/initprogram.F90
@@ -1817,7 +1817,7 @@ contains
 
     this%tPrintForces = input%ctrl%tPrintForces
     this%tForces = input%ctrl%tForces .or. this%tPrintForces
-    this%isLinResp = input%ctrl%lrespini%tInit
+    this%isLinResp = allocated(input%ctrl%lrespini)
     if (this%isLinResp) then
       allocate(this%linearResponse)
     end if
@@ -5226,6 +5226,8 @@ contains
     type(TInputData), intent(in), target :: input
 
     character(lc) :: tmpStr
+
+    @:ASSERT(allocated(input%ctrl%lrespini))
 
     if (withMpi) then
       call error("Linear response calc. does not work with MPI yet")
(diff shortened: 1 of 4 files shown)


src/dftbp/dftbplus/initprogram.F90
```diff
@@ -5227,6 +5227,8 @@ subroutine ensureLinRespConditions(tSccCalc, t3rd, tRealHS, tPeriodic, tCasidaFo
 
     character(lc) :: tmpStr
 
+    @:ASSERT(allocated(input%ctrl%lrespini))
```
bhourahine, line 5230: Could just return true if not allocated

bhourahine (approved):
Looks fine, but see suggestion about ensureLinRespConditions (which is quite a long routine name, but that's a separate issue...)

src/dftbp/dftbplus/initprogram.F90
bhourahine:
```suggestion
    if (.not.allocated(input%ctrl%lrespini)) then
      return
    end if
```

awvwgk (approved):

PR #956, opened 2022-02-10
API extension for export of DM, H, and S

PavelStishenko:
Lack of tests and code comments instead of TODO's


Mazay0:
@bhourahine How many MPI tasks are used for the [gcc-build (ubuntu-latest, openmpi, Debug, 9)]  check? I mean TEST_MPI_PROCS. It looks like there is only one task. It seems that due to that tests api/mpisubgrids and api/resetspecies_NaCl are not working. And my test api/mpiexpdmhs is pointless either  in that case. If it is true, then what is the reason for that?

aradi:
I like the idea, to collect all APIs into a derived type!

My biggest trouble with the current interface is, that it is very C-based. A Fortran host, whishing to register a callback, would have to convert its callback routine pointer to `type(c_funptr)` and its context pointer to `type(c_ptr)`, loosing all possibilities for error checks, when converting things back. I'll give it a thought, but probably we should implement it similarly to the external potential generator:

* A generic Fortran derived type, providing the API-callback Fortran interface (the context should be a `class(*)` pointer)
* An extension of the derived type, which contains the C-equivalent routines, a `type(c_ptr)` context pointer, and turns the Fortran call into a C-call.

This way, we also don't have to propagate `iso_c_binding` constructs into the very heart of DFTB+, but they stay, where they should, in the API-layer. I'll set this as topic for the next dpdevel meeting.

Mazay0:
@aradi , thanks for the review! 
I like the idea to split Fortran and C interfaces. Will do it.

Could you please also help me with this question: https://github.com/dftbplus/dftbplus/pull/956#issuecomment-1040869391 ? Do I understand correctly, that all checks including the one with MPI option run with a single process? So the multi-process MPI functionality is not tested by CI?

Mazay0:
Hello @aradi, 
I have added the pure Fortran layer to the API as discussed. Please review.

PavelStishenko:
Bump! @aradi @bhourahine , please take a look. I am writing the technical report for eCSE, so it would be better to have merging completed.

PR #957, opened 2022-02-10
Small changes for timeprop user experience

bhourahine:
* Describe where to look for units for dipole output files.

* Set modifier for phase units with fields (90 degrees being easier
than pi/2 to suply).

* Clear out a dead variable in the API tester


awvwgk (approved):
Looks good to me.

PR #959, opened 2022-02-21
Added missing import in fermihelper

terminationshock:
If EXP_TRAP is defined, the variable mExpArg is missing.


awvwgk (approved):
Looks good to me.

aradi (approved):

PR #960, opened 2022-02-21
Performance coulomb

terminationshock:
We can enhance the performance of the erf(c)wrap calls in the coulomb module. These are hotspots, so that we want the compiler to vectorize the expressions. To guide that, a temporary variable was introduced and SIMD directives were added.

### DIFF
--- a/src/dftbp/dftb/coulomb.F90
+++ b/src/dftbp/dftb/coulomb.F90
@@ -2226,7 +2226,7 @@ contains
     real(dp) :: dewr(3)
 
     real(dp) :: rNew(3)
-    real(dp) :: rr
+    real(dp) :: rr, factor
     integer :: iR
 
     dewr = 0.0_dp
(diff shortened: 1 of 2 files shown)


awvwgk (approved):
Did you check whether this patch actually improves the performance for evaluating the wrapped error function?

Apart from any potential performance gain, this patch improves at least the readability of the code.

terminationshock:
When compiling with Intel Fortran compiler, the SIMD directive helps vectorizing the erfc call. The optimization report says:

**Without SIMD directive:**
`remark #15525: call to function 'ERFCWRAP' is serialized
...
remark #15476: scalar cost: 254 
remark #15477: vector cost: 150.000
...
remark #15545: SIMD annotation was not seen, consider adding 'declare simd' directives at function declaration`

**With SIMD directive:**
`remark #15476: scalar cost: 254 
remark #15477: vector cost: 91.250
...
remark #15492: A suitable vector variant was found (out of 2) with xmm, simdlen=2, masked, formal parameter types: (vector)`

PR #962, opened 2022-02-23
Added more choices for ELPA configuration

terminationshock:
The configuration for ELPA was extended by two settings: "Autotune" and
"GPU", with which you can control ELPA's autotuning and GPU usage,
respectively. Both are switched off by default resembling the previous
behaviour.


terminationshock:
Dear DFTB+ developers, it is unclear to me how can I improve code coverage, because it seems that ELPA tests are not executed anyway. Hints are very much appreciated. Thank you!

awvwgk:
I started a test on the dpdevel buildbot for a04c6a57470683232e10ba3d837435e4c07c5f23.

bhourahine:
One issue that might have to be passed upstream to the ELSI team is that for an ELPA build without GPU support this will silently proceed (autotune might avoid this).

terminationshock:
> for an ELPA build without GPU support this will silently proceed

It can be easily fixed in ELSI by checking the error code when setting `...%set('gpu', 1, ierr)` or `...%set('nvidia-gpu', 1, ierr)`. From ELPA version 2021.11.001 onwards, `ierr /= ELPA_OK` if ELPA has not been compiled with GPU support.

aradi:
Amazing, I've tried the SiC512 example on dpdevel, and got a speed-up of 4-5x when switching on the GPU support.

@terminationshock We have merged build system changes to main, which would enable a proper build with ELPA and GPU. Could you rebase this branch on main and force push it to here?

terminationshock:
> Could you rebase this branch on main and force push it to here?

@aradi Done.

PR #963, opened 2022-02-25
Enable ELSI with GPU support, add cmake flag WITH_MAGMA

aradi:
Allows building DFTB+ with an ELSI library compiled with GPU support.

*Note:* This PR does not use the GPU-capabilities of the ELSI library yet, it only provides the framework to build against a GPU-aware ELSI library.


awvwgk (approved):

bhourahine (approved):
RTG

PR #973, opened 2022-03-06
Dipole and E fields for blocked cases

bhourahine:
Calculate dipoles for periodic cases and when there is a net charge. The absolute value is undefined, but changes can be used, for example to evaluate Born charges or piezoelectric tensors as a derivative of forces/stress with respect to field or dipole with distortions.

Likewise, allow electric fields for periodic systems, even if interactions cross the saw-tooth discontinuity.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,12 @@ Added
 Changed
 -------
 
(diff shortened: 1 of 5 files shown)


awvwgk (approved):
Thanks for sharing. Looks good to me.

PR #974, opened 2022-03-06
Born finite diff2

bhourahine:
Builds on top of #973 to extend the determinant loop (around SCC loop) for other purposes. In this PR, finite derivatives with respect to E field. Future PRs on top of this can then evaluate S_n Delta-DFTB, REKS determinants, constrained ground states, CI-DFTB, MPI parallelism over determinants.

- [x] DT methods for loop iterations and loop bail-out in the determinant loop
- [ ] Restructure class for other cases than S_0,S_1,T_1 Delta-DFTB determinant loops
  - [ ] Finite difference choices, plus H additions in each case
  - [ ] Collection of other properties
- [ ] Input
- [ ] Regression tests


awvwgk:
Any update on this? We have an application where we want to calculate IR spectra for solids with xTB, but need the dipole derivatives from DFTB+ for this.

bhourahine:
@awvwgk still working on this, a few more days to put it back together. To check, you want the Born charges (i.e. the mixed 2nd derivative of energy wrt to atomic positions and external electric fields, so either 1st derivative of forces wrt electric field or equivalently dipole wrt to atom positions), or something else?

awvwgk:
Actually, I'm looking for the derivative of the dipole moment w.r.t. the cartesian displacement to get the IR intensities, best calculated together with the Hessian. The Born charges would provide this information and allow reconstructing the IR intensities, unless I'm completely mistaken.

bhourahine:
The derivative of forces wrt to applied electric field is equivalent, and more efficiently calculated for systems containing > 3 atoms. Both give the Born charges.

Issue #981, opened 2022-03-10
Version conflict: VASP 4.x/5.x documentation


jubich:
The example in the manual uses VASP 4.x, but links to VASP 5.x.
The main difference is that in VASP 4.x the atom symbols are in line 1, while in VASP 5.x they are in line 6.

https://github.com/dftbplus/dftbplus/blob/5c110200a8b832a9e4f5148e1d6f9d644f1bada6/doc/dftb%2B/manual/dftbp.tex#L214-L227

awvwgk:
DFTB+ can actually read both 4.x and 5.x formatted POSCAR files.

jubich:
Yes, but I would say it is a bit misleading to give an example for VASP 4.x and link to VASP 5.x without mentioning that they are two different formats/versions or that DFTB+ can read both.

jubich:
Ok, I will change it.

PR #986, opened 2022-03-15
Update vasp documentation in manual

jubich:
Fix for Issue #981:  "Version conflict: VASP 4.x/5.x documentation".
However, I am not sure if it should be called VASP/Vasp or 4.x and 5.x formatted POSCAR.


doc/dftb+/manual/dftbp.tex
```diff
@@ -215,33 +215,35 @@ \subsection{VaspFormat\{\}}
 \label{sec:dftbp.VaspFormat}
 
 You can also use Vasp's \href{https://www.vasp.at/wiki/index.php/POSCAR}{POSCAR} or
-\href{https://www.vasp.at/wiki/index.php/CONTCAR}{CONTCAR} format to specify periodic geometries.
-The geometry specification for a molecular ammonia crystal would be the following:
+\href{https://www.vasp.at/wiki/index.php/CONTCAR}{CONTCAR} format to specify periodic geometries. 
+Both VASP 4.x and VASP 5.x are supported.
```
bhourahine, line 219:
```suggestion
Both VASP 4.x and VASP 5.x POSCAR are supported.
```

bhourahine (approved):
Looks good, thanks for your work.
Regarding your comment, does the suggestion fix this?

doc/dftb+/manual/dftbp.tex
jubich: Yes, this should fix it.

PR #989, opened 2022-03-17
Work around in API for external potentials

bhourahine:
Flags geometry as having changed if the potentials change (forcing re-evaluation of the ground state).

Added test for finite difference in field via pyapi (with commented out code for other response properties than dipole moment)


aradi (approved):
LGTM

PR #990, opened 2022-03-21
Use MPI shared memory in updateNeighbourList

terminationshock:
This feature uses MPI shared memory to reduce the memory footprint of the arrays `neigh%iNeighbour` and `neigh%neighDist2`, because now only one copy is kept per node. Furthermore, also the computation itself is accelerated by distributing the workload across all MPI ranks of the node.

Unit tests have been added where useful. The existing test suite of DFTB+ automatically test also this feature.


src/dftbp/dftb/periodic.F90
```diff
@@ -460,16 +481,37 @@ subroutine updateNeighbourList(coord, img2CentCell, iCellVec, neigh, nAllAtom, c
...
+          & startAtom, endAtom, parallelSetupError)
+#:else
+      parallelSetupError = .true.
+#:endif
+      if (parallelSetupError) then
+        parallel = .false.
+      end if
+    end if
+
+    if (.not. parallel) then
```
bhourahine, line 501: Is there a way to unify with  the existing MPI [distribution routine](https://github.com/dftbplus/dftbplus/blob/199d8cf7ea1aba266b10b0d12607bd37e592cc34/src/dftbp/common/schedule.F90#L39)  for similar loops?

src/dftbp/dftb/periodic.F90
```diff
@@ -430,15 +452,15 @@ subroutine updateNeighbourList(coord, img2CentCell, iCellVec, neigh, nAllAtom, c
 
     real(dp) :: dist2
     real(dp) :: rCell(3), rr(3)
-    integer :: ii, iAtom1, oldIAtom1, iAtom2
+    integer :: ii, iAtom1, oldIAtom1, iAtom2, startAtom, endAtom
     integer :: nn1, iAtom2End
-    logical :: symm
-    integer, allocatable :: indx(:)
+    logical :: symm, parallel, parallelSetupError
```
bhourahine, line 457: Could you switch to the more recent variable convention in use elsewhere in the code (and yes, symm doesn't use this unfortunately) of isParallel and isParallelSetupError ?

src/dftbp/dftb/periodic.F90
```diff
@@ -460,16 +481,37 @@ subroutine updateNeighbourList(coord, img2CentCell, iCellVec, neigh, nAllAtom, c
...
+#:if WITH_MPI
+    parallel = present(env)
+#:endif
+
+    if (parallel) then
+#:if WITH_MPI
+      call distributeAtoms(env%mpi%nodeComm%rank, env%mpi%nodeComm%size, nAtom, &
+          & startAtom, endAtom, parallelSetupError)
+#:else
+      parallelSetupError = .true.
```
bhourahine, line 494: Any reason not to just set
```suggestion
      parallel = .false.
```
as lines 496 - 498 are effectively just doing this?

terminationshock: You are right. I just simplified the logic.

src/dftbp/dftb/periodic.F90
terminationshock: Thanks for pointing this out. I will have a look into this. Could well be that those routines already do what I need here.

aradi (commented):
Thanks a lot, this is a very appreciated PR! See my comments, maybe we can discuss about this shortly today.

Please, also adhere to the maximal line length convention (of currently 100 chars). :wink:

src/dftbp/dftb/periodic.F90
```diff
@@ -460,16 +481,37 @@ subroutine updateNeighbourList(coord, img2CentCell, iCellVec, neigh, nAllAtom, c
...
+#:endif
+      if (parallelSetupError) then
+        parallel = .false.
+      end if
```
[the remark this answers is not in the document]
terminationshock: I simplified the logic in a different way based on Ben's comment. Should be much better readable now.

src/dftbp/dftb/periodic.F90
```diff
@@ -525,55 +567,166 @@ subroutine updateNeighbourList(coord, img2CentCell, iCellVec, neigh, nAllAtom, c
...
+    end if
+
+    do ii = 1, nAtom
+      neigh%iNeighbour(0, ii) = ii
+      neigh%neighDist2(0, ii) = 0.0_dp
+    end do
+
+  end subroutine fillNeighbourArrays
+
+
```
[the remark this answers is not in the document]
terminationshock: Yes, with the suggested changes, it is definitely possible.

src/dftbp/dftb/periodic.F90
```diff
@@ -1233,4 +1396,44 @@ subroutine cyl2cart_array(x,r)
...
+    integer, intent(out) :: startAtom, endAtom
+
+    !> Whether an error occurred during domain decomposition
+    logical, intent(out) :: error
+
+    integer :: i
+    real(dp) :: chunkSize
+
+    chunkSize = real(nAtom, dp) / real(mpiSize, dp)
+    if (chunkSize .lt. 1.0_dp) then
```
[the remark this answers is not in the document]
terminationshock: This code has now be replaced by calling `getChunkRanges`.

terminationshock:
@aradi @bhourahine I have implemented all your suggestions. FYI if you have further comments: I will be away next week.

terminationshock:
I would also like to do some more tests with the updated changes before you merge this branch.

terminationshock:
@aradi @bhourahine I have finished my tests: The latest commit was necessary, because the finalizer for the neighbour list did not work as expected. It was called rather early in the run and not at the end. That is why I now call it explicitly where it is needed.

bhourahine (approved):
Ready to merge

bhourahine:
@terminationshock Could you squash to a single commit?

terminationshock:
@bhourahine Squashed to a single commit :heavy_check_mark:

PR #991, opened 2022-03-23
Support for global Hybrids and CAM functionals

vanderhe:
Accounts for recently introduced features in SkProgs (cf. corresponding [PR](https://github.com/dftbplus/skprogs/pull/30)). For now only basic functionality is provided (single-point, analytic forces, probably MD, ...).

Done/remaining tasks:
- [x] extend range-separated type by more general CAM functionals (alpha, beta, omega)
- [x] global hybrids (parse SK extra tag, only full-range Hartree-Fock evaluation)
- [x] check analytical forces against finite differences (global hybrids|CAM)
- [ ] documentation
- [ ] regression tests

Later PR's would have to address:
- periodic boundary conditions (+ eventually k-points)
- (real-time) TD-CAM-DFTB
- xTB
- ...

### DIFF
--- a/src/dftbp/dftbplus/inputdata.F90
+++ b/src/dftbp/dftbplus/inputdata.F90
@@ -114,7 +114,13 @@ module dftbp_dftbplus_inputdata
     !> CAM beta parameter
     real(dp) :: camBeta
 
-    !> True, if a CAM-functional is requested, otherwise LC is assumed
+    !> True, for global hybrids
+    logical :: tHyb
+
+    !> True, for long-range corrected functionals
+    logical :: tLc
+
+    !> True, for general CAM range-separation
     logical :: tCam
 
     !> Choice of range separation method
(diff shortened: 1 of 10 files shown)


src/dftbp/dftbplus/inputdata.F90
```diff
@@ -108,6 +108,21 @@ module dftbp_dftbplus_inputdata
...
+    real(dp) :: camBeta
+
+    !> True, for global hybrids
+    logical :: tHyb
+
+    !> True, for long-range corrected functionals
+    logical :: tLc
+
+    !> True, for general CAM range-separation
+    logical :: tCam
```
awvwgk, line 124: Can we make this an enumerator? Three independent logicals are hard to track, instead we could just use an enumerator (maybe `exchange_kernel`) with the values `none`, `hybrid`, `longRange`, and `cam` (or similar).

PR #995, opened 2022-04-02
Update tblite interface to support spin channels

awvwgk:
Updates to new API of tblite 0.2.0 (still unreleased), but doesn't actually handle spin on the tblite side and initializes only one spin channel.


bhourahine (approved):
LGTM.
Potentially might be a good idea to have a spin-polarised regression case in future PR.

PR #1000, opened 2022-04-06
Enable xTB for Poisson solver and transport calculations

awvwgk:
- [x] export Hubbard parameters from tblite container
- [x] allow removing isotropic electrostatic from container
- [ ] enable Poisson solver with xTB
- [ ] enable NEGF calculations with xTB
- [ ] regression tests
- [ ] documentation
- [ ] changelog

cc @pecchia


bhourahine:
@awvwgk could you rebase on main to pick up the Poisson changes from #1003 ? Then we can find out if transition metals work.

PR #1002, opened 2022-04-09
Read GF electronic temperature correctly

bhourahine:
Also block reading of AdaptTemperature for thermostats with REKS

Two tests re-computed as effected by corrected temperature read


awvwgk (approved):

pecchia:
I think the fix is ok.

PR #1003, opened 2022-04-16
Arbitrary shell ordering in Poisson solver

bhourahine:
Also avoid making contact dir if no contacts present


bhourahine:
@awvwgk probably fixes the issue for transition metals xTB basis. No idea about non-minimal basis though.

awvwgk (approved):
Thanks.

PR #1004, opened 2022-04-17
MPI parallel range separation

bhourahine:
- [x] GEMM for density matrix build to produce both triangles
- [x] Dense distributed Mulliken (may need adjustment for iKS loops)
- [ ] Start of mixer changes (simple is possibly already working)
- [x] Distributed q0 subtraction
- [x] GEMM based algorithm -> pGEMM
- [ ] Test cases


vanderhe:
Hi Ben, I just stumbled across this PR. Can you already foresee if the way of MPI parallelization as you want to implement it here also works for the [gamma-point case](https://github.com/vanderhe/dftbplus/blob/05bd4f92e4a441ef9bc2895a33687ce69aaea7a8/src/dftbp/dftb/rangeseparated.F90#L1638-L1959)? I'm somehow in need for a performance bump there :). In the k-point case we have additional challenges, e.g. all \Delta H(\vec{k}) depend on all other k-points, so the (full) dual-space square \Delta P (\vec{k}) must probably be present.

PR #1007, opened 2022-04-20
Account for changed spherical harmonic ordering in tblite

awvwgk:
See https://github.com/tblite/tblite/pull/20


awvwgk:
The DNA structure is a bad test case because it has huge forces and a lot of strain, here even a small chance of conversion factors results in large contributions to the stress / forces and breaks the regression test.

PR #1008, opened 2022-04-21
Change reallocation constant

bhourahine:
Increases re-use of freed memory without appreciable overhead.


src/dftbp/common/memman.F90
```diff
@@ -15,23 +15,18 @@ module dftbp_common_memman
...
   !> grows more
-  function incrmntOfArray(currentSize)
+  pure function incrmntOfArray(currentSize)
     integer :: incrmntOfArray
 
     !> current array size
     integer, intent(in) :: currentSize
 
-    incrmntOfArray = int(real(currentSize, dp) * arrayIncrement)
+    incrmntOfArray = 3 * (currentSize + 1) / 2
```
awvwgk, line 29: `3 * (currentSize + 1)` could overflow
```suggestion
    incrmntOfArray = currentSize + currentSize  / 2 + 1
```

PR #1014, opened 2022-04-29
BLACS constructing both DM triangles

bhourahine:
Code simpler than pherk if pgemm used. Performance on 4 cores <1%
difference over all regression tests.


awvwgk (approved):
Looks good to me. Thanks for the additional docstrings.

PR #1021, opened 2022-05-04
Add wavelength input time prop

charlymedrano7:
I added the possibility to define the LaserEnergy in the input file for TimeProp directly as wavelength in nanometers (with the modifier "nm"). This is because for the study of optical properties people commonly speak about energy lasers in terms of wavelengths in nm.


aradi:
I think, if possible, we should go for a more general solution, as just to hack `nm` in wherever it is needed. I've tried to come up with something like this in #1025, this would enable to make `nm` part of `energyUnits`. I hope, I got the conversion factor right...

aradi:
Superseeded by #1025

PR #1024, opened 2022-05-10
Append geometry option and device range trap

bhourahine:
* GeometryOptimization environment with AppendGeometries flag
* Fix atom range default for transport calculations in
  GeometryOptimization
* Simplify atom range code for other drivers, reducing preprocessor
  use


awvwgk (commented):
Maybe adapt a test from the old geometry optimization drivers for the new ones as well?

bhourahine:
> Maybe adapt a test from the old geometry optimization drivers for the new ones as well?

If 2fef6bd2 is suitable I'll squash them together (the option doesn't effect any regression tested files).

awvwgk (approved):

PR #1025, opened 2022-05-10
Refactor unit system, implement reciprocal unit conversion

aradi:
Implements a more general unit conversion scheme, where f*x and f/x are both allowed.

It tries to offer a more transferable solution for the problem in #1021 .

### DIFF
--- a/doc/dftb+/manual/unit_mod.tex
+++ b/doc/dftb+/manual/unit_mod.tex
@@ -11,7 +11,9 @@ The allowed modifiers and the corresponding conversion factors are given
 below.\footnote{The conversion factors listed here were calculated with double
   precision on i686-linux architecture.  Depending on your architecture, the
   values used there may deviate slightly.}  (The modifiers are case
-insensitive).
+insensitive). The symbol $\left[x^{-1}\right]$ indicates conversions, where
+not the value, but the reciprocal value is multiplied by the conversion
+factor.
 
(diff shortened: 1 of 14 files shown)


test/app/dftb+/timeprop/benzene_laser/dftb_in.hsd
```diff
@@ -44,7 +44,7 @@ ElectronDynamics = {
   TimeStep [au] = 0.2
   Perturbation = Laser {
     PolarizationDirection = 0.0 1.0 1.0
-    LaserEnergy [eV] = 6.795
+    LaserEnergy [nm] = 182.46386964805214779105
```
bhourahine, line 47: Far too many decimal places, suggest truncating at ~8

doc/dftb+/manual/unit_mod.tex
```diff
@@ -11,7 +11,9 @@ \chapter{Unit modifiers}
 below.\footnote{The conversion factors listed here were calculated with double
   precision on i686-linux architecture.  Depending on your architecture, the
   values used there may deviate slightly.}  (The modifiers are case
-insensitive).
+insensitive). The symbol $\left[x^{-1}\right]$ indicates conversions, where
+not the value, but the reciprocal value is multiplied by the conversion
```
bhourahine, line 15:
```suggestion
not the value, but the reciprocal of the quantity is used in the conversion
```

bhourahine (approved):
Ready to merge after looking at review comments.

PR #1027, opened 2022-05-16
Add T. v. d. Heide to authors

vanderhe:
I always forgot to add myself to the author list. I guess my name will be inserted under "H"?

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -39,6 +39,8 @@ contributed to DFTB+ :
 
 * Sebastian Ehlert (University of Bonn, Germany)
 
+* Tammo van der Heide (University of Bremen, Germany)
+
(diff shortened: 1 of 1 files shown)


bhourahine:
Name ordering seems to be [culture dependant](https://academia.stackexchange.com/questions/8870/how-should-van-names-be-ordered-in-a-bibliography). So shall we do that, based on affiliation, or pick a convention?

aradi:
I'd suggest to go with the German lexical ordering convention: Name prefixes like 'van', 'de', etc. are not considered for the lexical ordering *as long as they start with lower case*. We used the same principle for the author list of the JCP 2020 paper.

bhourahine:
@aradi works fine for the time being (might have to think a bit more in the future if there is an affected US/Belgian/... contributor).

PR #1030, opened 2022-05-17
Minor memory saving in real H propagation

bhourahine:
Approximately 20% saving in routine memory use (<10% overall as
complex H and Sinv still allocated), with possible marginal
performance improvement.

### DIFF
--- a/src/dftbp/timedep/timeprop.F90
+++ b/src/dftbp/timedep/timeprop.F90
@@ -2079,16 +2079,13 @@ contains
     !> Time step in atomic units
     real(dp), intent(in) :: step
 
-    real(dp), allocatable :: T1R(:,:),T2R(:,:)
-    real(dp), allocatable :: T3R(:,:),T4R(:,:),T5R(:,:)
+    real(dp), allocatable :: T1R(:,:), T2R(:,:), T3R(:,:),T4R(:,:)
     integer :: i,j
 
     allocate(T1R(this%nOrbs,this%nOrbs))
     allocate(T2R(this%nOrbs,this%nOrbs))
     allocate(T3R(this%nOrbs,this%nOrbs))
     allocate(T4R(this%nOrbs,this%nOrbs))
-    allocate(T5R(this%nOrbs,this%nOrbs))
-
 
     ! The code below takes into account that Sinv and H1 are real, this is twice as fast as the
     ! original above (propageteRho)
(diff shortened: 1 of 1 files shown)


src/dftbp/timedep/timeprop.F90
```diff
@@ -2097,22 +2094,18 @@ subroutine propagateRhoRealH(this, rhoOld, rho, H1, Sinv, step)
...
-    !$omp parallel do private(i,j)
-    do i=1,this%nOrbs
-      do j=1,this%nOrbs
-        rhoOld(i,j) = rhoOld(i,j) + cmplx(0, -step, dp) * (T4R(i,j) + imag * T5R(i,j)) &
-            + cmplx(0, step, dp) * conjg(T4R(j,i) + imag * T5R(j,i))
-      enddo
-    enddo
-    !$omp end parallel do
+    rhoOld(:,:) = rhoOld + cmplx(0, -step, dp) * (T3R + imag * T4R)&
+        & + cmplx(0, step, dp) * transpose(T3R - imag * T4R)
```
awvwgk, line 2108: If you remove the `parallel do` at least `workshare` this step.

bhourahine: Suspect it would be better to just build with compiler options to use BLAS for matrix intrinsics.

awvwgk: I don't see any BLAS operations in this step, other than maybe `zscale` or `zaxpy`, which as level 1 operations are unlikely to have large impact.

bhourahine: Yep, level 1 only.

Curious how workshare will schedule the four matrix additions (simultaneously?)  if given enough processors.

awvwgk: I guess this is hardly a time consuming step in the computation. You could run the timepropagation with Intel's VTune to see whether this routine appears in the hot spot / threading analysis.

bhourahine: Yep (this routine does get called a lot, and effectively contains the only rank3 operations for fixed nuclei propagation, as overlap inversion would only happen once in that mode). Previous profiling suggested that for DFTB hamiltonians it's the interpolation of parameter tables that is the costly part.

awvwgk (approved):

PR #1032, opened 2022-05-18
Recover minNeighDist to minNeighDist2 agian

inseonglee:
Actually, this PR is done by PR #952 , but this correction is ignored in PR #990 , which shows warning from Codecov ("Added lines #L638 - L640 were not covered by tests").

Please merge this PR before releasing new version.


inseonglee:
@bhourahine @aradi In line number 529, `tolSameDist2` is used as a standard for deciding identical atom. But, in 539 line, `minNeighDist2` is used as a check. Is `minNeighDist2` is correct?

PR #1033, opened 2022-05-19
Ensure error returns for atoms too close

bhourahine:
Covers serial and MPI (both with and without node comm).

- [x] errStatus returned
- [x] synchronised error halt with consistent message if with node comm
- [x] unit testing


bhourahine:
@aradi any views? As discussed, need other PRs to fix assert inconsistencies and path propagation into unit tests.

awvwgk (approved):
Looks good to me.

PR #1034, opened 2022-05-20
Create directory for output of pump frame files

fbonafe:


sys/gnu.cmake
```diff
@@ -70,7 +70,7 @@ set(C_FLAGS_COVERAGE "-O0 -g --coverage")
 
 # LAPACK and BLAS
 # (if the BLAS library contains the LAPACK functions, set LAPACK_LIBRARY to "NONE")
-#set(BLAS_LIBRARY "openblas" CACHE STRING "BLAS libraries to link")
+set(BLAS_LIBRARY "openblas" CACHE STRING "BLAS libraries to link")
```
awvwgk, line 73: Don't commit changes to the toolchain file

bhourahine: @fbonafe this seems to be the culprit for the CI fails, can you revert change to this line and squash commits?

fbonafe: Didn't mean to, I'll revert it now, thanks for noticing!

fbonafe: Just forced-push the same commit without the toolchain file

src/dftbp/timedep/timeprop.F90
```diff
@@ -2234,6 +2236,14 @@ subroutine initTDOutput(this, dipoleDat, qDat, energyDat, populDat, forceDat, co
       end do
     end if
 
+    iErr = -999
+    if (this%tPump) then
+      call execute_command_line("mkdir "//trim(pumpFilesDir), exitstat=iErr)
+      if (iErr /= 0) then
+        write (stdOut,*) 'cannot create '//trim(pumpFilesDir)//', error status of mkdir: ', iErr
```
bhourahine, line 2243: Is this fatal? If so, call error()/propagate a status error upwards

fbonafe: It's not fatal, I tried leaving the directory there and running twice, the result is the same. It's a warning. I got it from the poisson code.

bhourahine: What happens if the directory is missing and can't be created?

fbonafe: Then it will be caught by writeRestartFile, which has the corresponding propagate status error

bhourahine (approved):
Looks ready to merge, @aradi ?

fbonafe:
Just saw a typo in the manual, let me fix it and force push once more

src/dftbp/timedep/timeprop.F90
```diff
@@ -2234,6 +2236,14 @@ subroutine initTDOutput(this, dipoleDat, qDat, energyDat, populDat, forceDat, co
       end do
     end if
 
+    iErr = -999
+    if (this%tPump) then
+      call execute_command_line("mkdir "//trim(pumpFilesDir), exitstat=iErr)
```
awvwgk, line 2241: Can't you call into the C library instead of resorting to a shell escape here?

src/dftbp/timedep/timeprop.F90
```diff
@@ -4107,7 +4117,7 @@ subroutine doTdStep(this, boundaryCond, iStep, coord, orb, neighbourList, nNeigh
         velInternal(:,:) = 0.0_dp
       end if
       call writeRestartFile(this%rho, this%rhoOld, coord, velInternal, this%time, this%dt,&
-          & trim(dumpIdx) // 'ppdump', this%tWriteRestartAscii, errStatus)
+          & trim(pumpFilesDir) // '/' // trim(dumpIdx) // 'ppdump', this%tWriteRestartAscii, errStatus)
```
awvwgk, line 4120: Joining file paths should account for the operating system.

src/dftbp/timedep/timeprop.F90
```diff
@@ -355,6 +355,8 @@ module dftbp_timedep_timeprop
   !> Prefix for dump files for restart
   character(*), parameter :: restartFileName = 'tddump'
 
+  character(*), parameter :: pumpFilesDir = 'pump_frames'
```
awvwgk, line 358: Missing a docstring here

awvwgk (commented):
It would be preferable to keep the code platform agnostic.

src/dftbp/timedep/timeprop.F90
fbonafe: I don't know how to use the C library. Do you have an example in the code where you mkdir using this library? This is how it's done in the Poisson code

src/dftbp/timedep/timeprop.F90
fbonafe: Agreed, would using this procedure be fine?
```
CALL GET_ENVIRONMENT_VARIABLE('PATH',path)
path_separator=path(1:1)
```

src/dftbp/timedep/timeprop.F90
awvwgk: See

- https://www.gnu.org/software/libc/manual/html_node/Creating-Directories.html
- https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/mkdir-wmkdir?view=msvc-170

src/dftbp/timedep/timeprop.F90
awvwgk: If your `PATH` starts with a drive letter `C:\` this will not work.

src/dftbp/timedep/timeprop.F90
fbonafe: Do we have a Fortran interface to this?

aradi: No, not yet. However, I definitely would like to avoid to have a shell-call within the code. (Poisson does not count, it is declared as "external library", meaning, we don't take responsibility, how things are coded up there :wink: ). libNEGF uses something more appopriate (see `ext_system/system_calls.f90`), but that's deep within libNEGF...

I have actually functional wrappers for almost all file manipulation routines of libc, I just have to set up a proper CMake build system for it. I won't manage it, though, before the release. (fragments of some earlier projects of mine....)

src/dftbp/timedep/timeprop.F90
aradi: The correct solution would be IMO a pre-processor based approach, setting the separator based on the system, as reported in the `${_SYSTEM_}$` variable in Fypp. (And actually a function, like `joinPaths()` which hides all those nasty details. Or even better, a Path class, like in recent python releases....)

aradi (approved):
I am not really happy, as the solution is rather hacky. Nevertheless, since we already have a similar solution in the code (shame on us), and improves the usability, I'll merge it into to main and to the release. 

@fbonafe: If I managed to repackage my  old routines, we will have to recode this part. (And hope, that nobody looks at this part of the code in the mean time :wink: )

src/dftbp/timedep/timeprop.F90
fbonafe: Sorry, I did not realise that this would be that complicated! Let's forget about this PR in that case. I'll add the mkdir and move files to the Python scripts and do it as a postprocessing.

fbonafe:
Thanks @aradi! Let'a add the wrapper in the future and we can improve this.

PR #1038, opened 2022-05-23
Minor change to dptools install info

bhourahine:
Replace python with python3 (as usual install name). Also describe
path setting a bit more.

### DIFF
--- a/tools/dptools/README
+++ b/tools/dptools/README
@@ -60,31 +60,12 @@ utils/srccheck/pylint/pylintrc-3.ini tools/dptools/bin/*
 Installation
 ============
 
-Please note, that the package needs at least **Python 2.6** or
-later. It should also work with Python 3.X. It additionally needs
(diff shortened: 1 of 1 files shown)


tools/dptools/README
```diff
@@ -71,14 +71,14 @@ You can install the script package via the standard 'python setup'
 mechanism. If you want to install it system-wide into your normal
 python installation, you simply issue::
 
-  python setup.py
+  python3 setup.py
```
awvwgk, line 74: Shouldn't we recommend installation via `pip install .` instead?

bhourahine: Probably in addition, yes.
But... there is a namespace collision:
https://pypi.org/project/dptools/
so instructions will need to be done carefully.

awvwgk (approved):

awvwgk (approved):

PR #1045, opened 2022-05-28
Refactoring getDensity routine

inseonglee:
The 'getDensity' subroutine is separated into three individual routines; building Hamiltonian, solve equation, get density matrix


inseonglee:
@baradi09 @aradi Please, see the current stage for this PR. The works for only conventional solvers are finished. If there are no other comments, then I will continue modifying separation of getDensity for other solvers such as ELSI or GF case.
Also, the writeHS or writeRealHS option now includes correct contribution from rangeseparated calculation. (see commits c867787 to cf7b541.)

inseonglee:
If there are no comments on the current stage of this PR, I will start the next modification.

inseonglee:
@bhourahine @aradi For this PR, I passed the regression tests for serial compilation with intel21 compilers.
Now, I have to test the parallel cases but there is a problem for link of ELSI libraries, please see the figure.

image

As seen in this figure, it seems that the ELSI libraries may be linked to dftb+, then how can I solve this error?

inseonglee:
> @bhourahine @aradi For this PR, I passed the regression tests for serial compilation with intel21 compilers. Now, I have to test the parallel cases but there is a problem for link of ELSI libraries, please see the figure.
> 
> image
> 
> As seen in this figure, it seems that the ELSI libraries may be linked to dftb+, then how can I solve this error?

I solved this problem by recompiling the ELSI libraries, and the regression tests are passed for ELSI examples.
Currently, I am trying to test the NEGF examples, but I failed to compile the codes.
Could you try the regression tests for NEGF case?

inseonglee:
For NEGF examples, I passed the tests with only release compilation.
The compilation with debug option shows the catastrophic error from following line in complexbands.f90 file of libNEGF.
`tmp=matmul(transpose(conjg(Cn)),matmul(TM*exp(j*kk),Cn))`

inseonglee:
I solved this problem by recompiling the ELSI libraries, and passed all regression tests.
Now, I will try to test the NEGF case.

inseonglee:
@bhourahine @aradi Only two tests are failed with the codes using release option. The failed tests are 'C-chain+U' and 'C-chain+Uspin'. I compared the results using this branch (divideGetDensity) and main branch, and both codes show identical value.
So, could you compare the results using current main and this branchs in your cluster?
It seems that the autotest.tag may be wrong in these examples.
If we solve this issue, then this PR would be ready for review.

inseonglee:
I rebased this branch, and now all regression tests are passed. Please review this PR.

awvwgk (commented):
Storing the dense square matrices for each k-Point will significantly increase the memory demand for periodic calculation and remove any advantage gained from the atomic block sparse storage.

inseonglee:
> Storing the dense square matrices for each k-Point will significantly increase the memory demand for periodic calculation and remove any advantage gained from the atomic block sparse storage.

Yes, that's right. This PR requires more memories for allocation of H and S. But, this PR is essential part to combine the REKS_SCC and SCC loops for concise codes.
Two months ago, I discussed the separation of the 'getDensity' routine with Ben Hourahine and Balint Aradi, and we concluded that this PR and #974 would be proceeded independently.
After that, I will try to combine the REKS_SCC and SCC parts by utilizing the multideterminant loops.

inseonglee:
@bhourahine @aradi Did you see this PR? The last commit was written at August. If more detailed discussion for unifying the REKS_SCC and SCC loops is needed, then I will create new PR for implementation of onsite correction from LC functional and its REKS part.

PR #1046, opened 2022-05-31
Make sure ParserVersion appears in dftb_in.hsd

aradi:
Was missing if InputVersion was set and no conversion was necessary.

### DIFF
--- a/src/dftbp/dftbplus/parser.F90
+++ b/src/dftbp/dftbplus/parser.F90
@@ -331,24 +331,29 @@ contains
     if (present(implicitVersion)) then
       call getChild(node, "ParserVersion", child, requested=.false.)
       if (associated(child)) then
-        call detailedError(child, "Cannot have both ParserVersion and InputVersion")
+        call getChildValue(child, "", inputVersion)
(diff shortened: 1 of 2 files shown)


src/dftbp/dftbplus/parser.F90
```diff
@@ -331,24 +331,29 @@ subroutine readParserOptions(node, root, flags, implicitVersion)
...
-          &child=child)
+      call getChildValue(node, "ParserVersion", inputVersion, parserVersion, child=child)
     end if
+
     if (inputVersion < 1 .or. inputVersion > parserVersion) then
-      call detailedError(child, "Invalid parser version (" // i2c(inputVersion)&
-          &// ")")
-    elseif (inputVersion < minVersion) then
+      call detailedError(child, "Invalid parser version (" // i2c(inputVersion) // ")")
+    else if (inputVersion < minVersion) then
```
bhourahine, line 350: Probably better to have this branch first instead of as an else if clause, as 1 <= minVersion at the moment, but might eventually change

bhourahine (approved):
See comment about if () elseif order, but otherwise looks OK.

PR #1051, opened 2022-06-14
Tests results.tag and fixes 2 dipole bugs

bhourahine:
* Use exisiting regression mechanism on results.tag output for a test
  example

* Fix use of (effectively) a dummy array in atomic dipole printing

* defaults rescale type variable to false (so set if not initialised)


bhourahine:
@awvwgk can you test if this fixes the results.tag bug?

awvwgk (approved):
Thanks, this fixes the broken examples I have here. Tested with GCC 12 in debug mode.

src/dftbp/dftbplus/mainio.F90
```diff
@@ -2207,19 +2205,13 @@ subroutine writeResultsTag(fileName, energy, derivs, chrgForces, nEl, Ef, eigen,
...
-      end if
+      call taggedWriter%write(fd, tagLabels%scaledDipole,&
+          & eFieldScaling%scaledSoluteDipole(dipoleMoment))
     end if
 
     if (allocated(multipole%dipoleAtom)) then
-      block
-        real(dp), allocatable :: dipoleAtom(:, :), qAtom(:)
-        qAtom = sum(qOutput(:, :, 1) - q0(:, :, 1), dim=1)
-        call taggedWriter%write(fd, tagLabels%dipoleAtom,&
```
bhourahine, line 2220: Think this was probably the segfault, as dipoleAtom not allocated

bhourahine:
> Thanks, this fixes the broken examples I have here. Tested with GCC 12 in debug mode.

@aradi Before this is merged, can you also review, due to regression test structure?

PR #1053, opened 2022-06-14
Ise parser

bhourahine:
Anglicise keyword spellings, matching existing cases for other keywords (e.g. spin polarisation not polarization).
Parser and API versions increased.


awvwgk (approved):
Haha, I didn't think about this when creating the optimi~~z~~ser keywords.

For consistency we should fix those keywords.

PR #1058, opened 2022-06-21
Add X+Y data for NrOfExcitations to autotest.tag

thomas-niehaus:
PR to allow for overlapping atoms upon user request and write sensible overlap data. 
Necessary for interface with Newton-X.
Here: Implemented writing of X+Y data to autotest.tag, in this way only one file needs to be parsed by Newton-X. 
ToDo (?) Writing to results.tag instead of autotest.tag, but currently all excited state info is in autotest.tag. 
           (?) Remove writing to X+Y.DAT. The tag file can become quite large, redundant info.

### DIFF
--- a/src/dftbp/io/taggedoutput.F90
+++ b/src/dftbp/io/taggedoutput.F90
@@ -212,6 +212,9 @@ module dftbp_io_taggedoutput
     !> Atomic dipole moments
     character(lenLabel) :: dipoleAtom = 'atomic_dipole_moments'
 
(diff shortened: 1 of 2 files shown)


bhourahine:
Might be better to write to results.tag, as you mention, and use the scripting solution from #1051 to regression test that file instead of autotest.tag.

aradi:
Will be superseeded soon by a PR for writing this data into `results.tag`

PR #1059, opened 2022-06-25
Dipole derivatives wrt positions

bhourahine:
Finite difference derivatives of dipoles
- [x] molecular case tested as working
- [x] periodic case tested as working
- [x] documentation and regression tests
- [x] Clean up output

This matches between d mu/ dx and dF/dE for DFTB2. For periodic geometries, unfolded coordinates required for dipoles

Issues
=====
- External electric fields with xTB are broken.


awvwgk:
Can you please rebase this branch to see whether the failures are only related to the dptools issue?

awvwgk:
The derivative test `201 - dftb+_derivatives/gfn2_c6h6_dipole (Failed)` fails for Intel, any idea why it passes with GCC?

bhourahine:
> The derivative test `201 - dftb+_derivatives/gfn2_c6h6_dipole (Failed)` fails for Intel, any idea why it passes with GCC?

Yes. The intel workflow doesn't build TBLITE, but GCC does, and the test was running irrespective of TBLITE being included.

PR #1066, opened 2022-07-13
Add check for exceeding HSD line lengths

### LINKED ISSUE #1063: Problem reading poscar coordinates: no more than 384 atomes read
I just started to play with DFTB+, last stable version installed with anaconda on a CENTOS 7 dell R640 computer (48core, 392GB RAM)

I try to play with a 20Ang box of 266 water with a small organic molecule inside. I am using xtb hamiltonian.
I found the problem on geometry optimization, but it also happen for single point energy calculation: 

From vasp postcar coordinates file, only the 384 first atoms are used. 
No error message. Others atoms of the input are just ignored And calculation goes on.
It works fine for generic coordinates.

DFTB+ input file and vasp coordinate file are  in attachment.

Best,
Bruno Madebène

[detailed.zip](https://github.com/dftbplus/dftbplus/files/9073922/detailed.zip)

aradi:
HSD parser stops now if length of a line exceeds allowed limit (currently 1024 chars).

Closes #1063

### DIFF
--- a/src/dftbp/io/hsdparser.F90
+++ b/src/dftbp/io/hsdparser.F90
@@ -280,8 +280,8 @@ contains
         strLine = adjustl(residual)
         residual = ""
       else
-        read (fd, lineFormat, iostat=iostat) strLine
         curLine = curLine + 1
+        call getLine(fd, curFile, curLine, strLine, iostat)
         call convertWhitespaces(strLine)
         strLine = adjustl(strLine)
         !! If reading error (e.g. EOF) -> close current scope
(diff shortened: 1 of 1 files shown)


bhourahine (approved):
Seems fine, any chance of a unit test?

awvwgk:
> Seems fine, any chance of a unit test?

I think we currently cannot test failures in our testsuite (in general, yes for testing this!).

bhourahine:
> I think we currently cannot test failures in our testsuite (in general, yes for testing this!).

Unit should be able to, not regression

PR #1015, opened 2022-04-29
Restructuring of dptools and ptools

jubich:
- [ ] add unit-tests

PR #1068, opened 2022-07-16
Adding Typing support to python moduls (api, dptools and ptools)

jubich:
- [ ] Implementation of the [restructuring](https://github.com/dftbplus/dftbplus/pull/1015)
- [ ] Discussion about the supported Python versions because of the use of [typing](https://docs.python.org/3/library/typing.html)
- [ ]  Discussion about the handling of ctypes in [pythonapi](https://github.com/jubich/dftbplus/blob/PythonTyping/tools/python/api/src/dftbplus.py)


aradi:
Should be still considered.

PR #1073, opened 2022-07-27
Atomic energies output for time propagation

fbonafe:
Small PR to add an output (atom-resolved energy in real time), that was requested from a user.


src/dftbp/timedep/timeprop.F90
```diff
@@ -2241,12 +2248,23 @@ subroutine initTDOutput(this, dipoleDat, qDat, energyDat, populDat, forceDat, co
...
     end if
 
+    if (this%tWriteAtomEnergies) then
+      call openFile(this, atomEnergyDat, 'atomenergies.dat')
+      write(atomEnergyDat, "(A)", advance = "NO")"#             time (fs)      |"
+      write(atomEnergyDat, "(A)", advance = "NO")"   E total (H)  |"
+      write(atomEnergyDat, "(A)", advance = "NO")"   E (atom_1) (H)   |"
+      write(atomEnergyDat, "(A)", advance = "NO")"   E (atom_2) (H)   |"
+      write(atomEnergyDat, "(A)", advance = "NO")"        ...        |"
+      write(atomEnergyDat, "(A)", advance = "NO")"   E (atom_N) (H)   |"
```
awvwgk, line 2258: What happens if you have only one atom?

src/dftbp/timedep/timeprop.F90
```diff
@@ -2459,6 +2487,14 @@ subroutine writeTDOutputs(this, dipoleDat, qDat, energyDat, forceDat, coorDat, f
       end if
     end if
 
+    if (this%tWriteAtomEnergies) then
+        write(atomEnergyDat, "(2X,2F25.15)", advance="no") time * au__fs, energy%Etotal
+        do iAtom = 1, this%nAtom
+          write(atomEnergyDat, "(F25.15)", advance="no")this%energy%atomTotal(iAtom)
+        end do
+      write(atomEnergyDat, *)
```
awvwgk, line 2495: Wouldn't it be preferable to adopt tagged output here?

awvwgk (approved):
Generally okay to merge.

src/dftbp/timedep/timeprop.F90
fbonafe: Why tagged output?

bhourahine: So it's a) easily machine readable, b) can be checked for changes/bugs by the regression framework and c) if the tagged process ever gets co-opted to redirect through the API this automatically returns your data that way as well.

src/dftbp/timedep/timeprop.F90
bhourahine: Can we therefore make a routine to format this properly for both cases? Possibly as a separate PR to avoid holding this one up.

doc/dftb+/manual/dftbp.tex
```diff
@@ -6392,6 +6396,16 @@ \subsection{qsvst.dat}
 where N is the number of atoms, and the net total charge should always
 be zero (up to numerical presicion) for a neutral system.
 
+\subsection{atomenergies.dat}
+\index{atomenergies.dat}
+
+Atom-wise contribution to total energy (without nuclear kinetic energy)
```
bhourahine, line 6402: Is there a reason not to include Ekin in the output, or even give the energy with it included in a separate output?

src/dftbp/timedep/timeprop.F90
```diff
@@ -2459,6 +2487,14 @@ subroutine writeTDOutputs(this, dipoleDat, qDat, energyDat, forceDat, coorDat, f
       end if
     end if
 
+    if (this%tWriteAtomEnergies) then
```
bhourahine, line 2490: Can you add a checked example case to the regression test collection? These lines of code are not run with any current input, so are not protected against something breaking on refactoring.

bhourahine (changes requested):
Definitely should get regression working for these changes (i.e. tagged and a test case), but see other things for (separate future PRs?).

src/dftbp/timedep/timeprop.F90
fbonafe: I see, but I thought Sebastian meant to replace the current output file by a tagged one, is that what he meant? Because for regression test purposes I can just add an entry to autotest.tag with this info

awvwgk: `csv` is an easy machine readable format (which is also compatible with list directed reads).

bhourahine: I guess the question is how large this data is going to get, presumably its written at most iterations? For regression purposes, a snapshot of the final set would be OK.
But for machine readable data in real calculations, I think we need to set up a proper framework (including generating CSV or similar as an option or binary once things get big). That would be outside the scope of this PR.

src/dftbp/timedep/timeprop.F90
fbonafe: I just added a test and tagged output

doc/dftb+/manual/dftbp.tex
fbonafe: It can be calculated with the velocities printed to tdcoords.xyz, and the user that requested this feature does not seem to need the Ekin per atom, so I decided to leave it out for simplicity

src/dftbp/timedep/timeprop.F90
fbonafe: For this particular case I don't expect the data to be huge, it's only Natoms x Noutputsteps, and by default the output frequency is 50 steps, Noutputseps = Nsteps / 50. We have massive outputs like the bond-wise populations or energy that are stored in binary files, but for other outputs like orbital populations (which is Norbs x Noutputsteps) I guess it would be good to set such a framework.

src/dftbp/timedep/timeprop.F90
fbonafe: Yes, we can do it in a separate PR, I'll add it to the to do.

bhourahine:
OK, so follow up PRs for headers and abstracting the writing as a method of a derived type (so it can be switched from text, to csv, to ...).

PR #1077, opened 2022-08-03
Fix for uninitialised stress in sockets

bhourahine:
If boundary conditions don't support stress tensor, init to 0 in DFTB+ and skip printing in regression test.


awvwgk (approved):
Looks good to me.

PR #1080, opened 2022-08-07
Fix shell labelling for > 2 collisions

bhourahine:
Labels for repeated shells with the same l value were failing
above 2 cases, i.e. would get s s2 s2 .. s2.
This change correctly labels multiple instances (s, s2, s3, ...).

Co-authored-by: Berk Onat <B.Onat@warwick.ac.uk>


awvwgk (approved):
Good catch.

PR #1081, opened 2022-08-12
Update Python command line parsing to argparse

jubich:
This PR updates the command line parsing to argparse.

### DIFF
--- a/tools/misc/calc_timeprop_maxpoldir
+++ b/tools/misc/calc_timeprop_maxpoldir
@@ -11,45 +11,40 @@ Get direction of maximum polarizability from DFTB+ TD data
 '''
 
 import numpy as np
-import optparse
+import argparse
 from scipy import linalg, constants
 import sys
 hplanck = constants.physical_constants['Planck constant in eV s'][0] * 1.0E15
 
-USAGE = """usage: %prog [options]
+USAGE = """Reads output from TD calculation (after kick) and excitation energy
+(or wavelength) of interest and diagonalizes polarizability tensor, to
+calculate direction of maximum polarizability of that particular excitation.
 
-Reads output from TD calculation (after kick) and excitation energy (or wavelength)
-of interest and diagonalizes polarizability tensor, to calculate direction of maximum
-polarizability of that particular excitation.
-
-Needs mux.dat, muy.dat and muz.dat files in workind directory."""
+Needs mux.dat, muy.dat and muz.dat files in working directory."""
 
 def main():
-    parser = optparse.OptionParser(usage=USAGE)
-    parser.add_option("-d", "--damping", action="store", dest="tau",
-                      help="damping constant in fs")
-    parser.add_option("-e", "--energy", action="store", dest="energy",
-                      help="excitation energy in eV")
-    parser.add_option("-w", "--wavelength", action="store", dest="wvlength",
-                      help="wavelength of excitation in nm")
+    parser = argparse.ArgumentParser(description=USAGE)
+    parser.add_argument("tau", metavar="DAMPING",
+                        help="damping constant in fs")
 
-    (options, args) = parser.parse_args()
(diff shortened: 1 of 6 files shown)


tools/misc/calc_timeprop_maxpoldir
```diff
@@ -25,21 +25,21 @@ polarizability of that particular excitation.
 Needs mux.dat, muy.dat and muz.dat files in workind directory."""
```
vanderhe, line 25:
```suggestion
Needs mux.dat, muy.dat and muz.dat files in working directory."""
```

tools/misc/calc_timeprop_spectrum
```diff
@@ -11,24 +11,29 @@ Fourier transform of DFTB+ TD data
...
+    parser.add_argument("-d", "--damping", dest="tau",
+                        help="damping constant in fs")
+    parser.add_argument("-f", "--field", dest="field",
+                        help="field intensity in volts/angstrom")
+
+    options, args = parser.parse_known_args()
+
+    if len(args) != 0:
+        parser.error("You must specify exactly two argument, the damping constant in fs and "
+                     "the field intensity in volts/angstrom")
```
vanderhe, line 35: I know it's not your code-design, but since the damping constant and field intensity are required in any case, why make them optional in the first place?

tools/misc/calc_timeprop_maxpoldir
```diff
@@ -25,21 +25,21 @@ polarizability of that particular excitation.
...
-                      help="damping constant in fs")
-    parser.add_option("-e", "--energy", action="store", dest="energy",
-                      help="excitation energy in eV")
-    parser.add_option("-w", "--wavelength", action="store", dest="wvlength",
-                      help="wavelength of excitation in nm")
-
-    (options, args) = parser.parse_args()
+    parser = argparse.ArgumentParser(usage=USAGE)
+    parser.add_argument("-d", "--damping", dest="tau",
+                        help="damping constant in fs")
```
vanderhe, line 30: Same as comment below (sorry for confusing order). In my opinion the damping constant should not be optional in the first place.

tools/misc/convolve.py
```diff
@@ -6,51 +6,43 @@
...
+parser.add_argument("-b", "--broaden", dest="sigma", type=float,
+                    help="broadening width")
+parser.add_argument("-i", "--input", dest="infile",
+                    help="input file name")
+parser.add_argument("-u", "--unnorm", dest="norm", action="store_false",
+                    help="don't normalize by number of spins/kpoints")
+parser.add_argument("-w", "--weight", dest="weight", action="store_true",
+                    help="weight data present")
+
+options, args = parser.parse_known_args()
```
vanderhe, line 29: Would be nice if we could use the opportunity to add a useful 'usage-message' here.

vanderhe (commented):
Seems to work as intended, I've just added some comments that came to my mind when going through the changes made.

tools/misc/convolve.py
jubich: Do you have any suggestions for the message?

aradi:
@jubich Could you squash the two commits into one and rebase it to current main? Then we could merge it as a single commit on top of current main without the necessity of a merge commit.

jubich:
Sure, but before that I would like to add the 'usage-message' in convolve.py as suggested by @vanderhe. But I'm not sure how to formulate a good one. So I'm waiting for an answer/message to add it. But if it is not necessary/wanted, I can leave it out. It can then be added at a later time.

tools/misc/convolve.py
aradi:
```text
Broadens the eigenvalues with Gaussians of given width
```
Should be enough. Actually, this is an ancient leftover, probably we should remove it at some (later) point...

PR #1083, opened 2022-08-15
Add contact labels to parcheck in transport

bhourahine:


src/dftbp/poisson/parameters.F90
```diff
@@ -236,6 +237,20 @@ subroutine set_contdir(dir)
...
+  subroutine set_contlabels(names)
+
+    !> Names of contacts
+    character(mc), intent(in) :: names(:)
+
+    if (allocated(contnames)) then
+      deallocate(contnames)
+    end if
+    allocate(contnames(size(names)))
+    contnames(:) = names
```
awvwgk, line 250: Automatic LHS allocation does not work in this context, *i.e.* is buggy in our compilers?

bhourahine: Now changed to automatic.

PR #1085, opened 2022-08-15
Removes input for setting charge if no meaning

bhourahine:
Transport calculations can't set total charge in the device region.


awvwgk (approved):
Looks good to me.

PR #1090, opened 2022-08-25
Range separation +U combination enabled

bhourahine:
Also print eigen-decomposition of block populations for atoms.


awvwgk (approved):
Thanks for sharing. Looks good to me.

PR #1096, opened 2022-08-27
Proper cmake processing of dftbplus.h file

### LINKED ISSUE #1095: dftbplus.h preprocessing broken
**Describe the bug**
Macros with strings in the include file, as set by cmake, are broken.

**To Reproduce**
Generate dftbplus.h and examine contents

**Expected behaviour**
Correct macro expansions

bhourahine:
Only generated in build when WITH_API.
Contains correctly expanded API version macro.
Fixes bug introduced in c0f2a7dce

Closes #1095

### DIFF
--- a/src/dftbp/CMakeLists.txt
+++ b/src/dftbp/CMakeLists.txt
@@ -155,15 +155,18 @@ endif()
 
 if(INSTALL_INCLUDE_FILES)
   install(DIRECTORY ${includedir}/ DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/${INSTALL_MODULEDIR}")
-  CONFIGURE_FILE(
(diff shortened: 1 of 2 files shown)


awvwgk (approved):
Looks good to me. Maybe even more useful would be to install a stub header with an ```#error``` directive to show why the API is missing.

bhourahine:
@awvwgk are you proposing to install the dummy file as well? (Unless the API or shared library is built, there is no install path set up). Probably one for a separate PR in either case.

awvwgk:
> are you proposing to install the dummy file as well?

If we always installed the header even if the the API is not enabled, it might be better to install a stub header to communicate this change to projects consuming the DFTB+ C-API.

PR #1098, opened 2022-08-29
External model

bhourahine:
Importing model for energy/hamiltonian from external library.

- [x] Internal and keyword re-name of Hamiltonian to Model, as more descriptive of nature of total energy expressions (its not just an electronic Ham).
- [x] Model capability declarations
- [x] Model initialisation
- [x] Model internal state passed around
- [ ] Logic for returned cut-offs
- [ ] Ancillary model choice for obtaining total energy (i.e. ext model rep. + DFTB Ham; ext model ham + DFTB rep etc.)
- [ ] Documentation
- [ ] External header/module files

Then a future PR to use the example [external model ](https://github.com/bhourahine/dftbp-extmodel) as fallback if no actual external model specified, allowing regression testing.


bhourahine:
Replaced by #1101

PR #1099, opened 2022-09-03
Updates submodules for a fixed typo upstream

bhourahine:
Probably should look at some method for automatic notification of upstream changes on submodules.

In this case, a repeating CMake typo in several projects was fixed (plus other changes).


awvwgk (approved):
Sure, let's keep the *tblite* stack up-to-date with upstream.

PR #1102, opened 2022-09-06
Fix non-advancing I/O problems for incomplete lines

aradi:
Last line of a file without trailing newline is handled now correctly by the HSD parser.

### DIFF
--- a/src/dftbp/io/hsdparser.F90
+++ b/src/dftbp/io/hsdparser.F90
@@ -258,7 +258,7 @@ contains
     type(fnode), pointer :: childNode, dummy
     type(string) :: buffer
     integer :: newFile
-    integer :: iostat
+    integer :: iostat, newIoStat
     integer :: iType, sepPos
     integer :: newCurLine
     logical :: tTagClosed, tNewNodeCreated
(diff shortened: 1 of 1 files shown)


awvwgk (commented):
Why not check for [`is_iostat_end`](https://fortran-lang.org/en/learn/intrinsics/state/#is-iostat-end)?

aradi:
Unfortunately, under certain circumstances (file ending without newline character) some compilers report `iostat_end` and deliver valid content at the same time (see https://fortran-lang.discourse.group/t/inconsistent-eor-and-eof-conditions-with-non-advancing-i-o/4270)

awvwgk:
Maybe we can switch to stream access instead?

https://github.com/toml-f/toml-f/blob/4ea08ea7c1e15dbee660e3c5e97e5dbad8563f91/src/tomlf/utils/io.f90#L25-L52

aradi:
An elegant solution, but we need the file in the current parser line-by-line. I have now implemented a line-reader which does that (similar to your `read_whole_line()` routine, but it also makes sure, that you never get iostat /= 0 when valid content is returned). I'll make a PR for that later today.

PR #1103, opened 2022-09-08
Lift line length limitation in HSD parser

### LINKED ISSUE #1063: Problem reading poscar coordinates: no more than 384 atomes read

aradi:
Also fixes problems for incomplete lines at the end of files.

Closes #1102 
Closes #1063 

Fun fact: By using allocatable characters everywhere (instead of fixed length ones), parsing became four times faster! (Probably `trim()` works much more efficiently...)

### DIFF
--- a/doc/dftb+/manual/hsd.tex
+++ b/doc/dftb+/manual/hsd.tex
@@ -48,9 +48,6 @@ insensitive, so the assignments
 are pairwise identical. Quoted strings (specified either as a value
 for a property or as a file name), however, are case sensitive.
 
-Due to technical issues, the maximal line length is currently limited to 1024
-characters. Lines longer than this are chopped without warning.
-
 If a property, which should only appear once, is defined more than once, the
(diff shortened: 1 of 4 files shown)


awvwgk (approved):
Very cool.

PR #1105, opened 2022-09-08
Finalize mpi_wins in TNeighourList automatically

aradi:


bhourahine (approved):
LGTM

awvwgk (approved):
As long as we don't assign the neighbour list this should be fine.

aradi:
That's a valid point with the assignment. Probably, we should override assignment to stop if invoked.

PR #1109, opened 2022-09-15
Make building and linking of MUDPACK optional

aradi:
Due to bugs, MUDPACK was build and linked, even with WITH_POISSON=0


bhourahine (approved):
LGTM

PR #1114, opened 2022-09-20
Fix for not always initialised variable

bhourahine:


awvwgk (approved):
Good catch.

bhourahine:
> Good catch.

Valgrind

PR #1116, opened 2022-09-28
Update parser.F90

pecchia:
fix wrong units conversion of gate Potential for cylindrical gate


bhourahine (approved):
Ready to merge, but should possibly add a mention to the CHANGELOG.rst under the 'fixed' section (as this will change code behaviour for people using this feature).

PR #1119, opened 2022-10-03
Cosmo features

aradi:
Contains some Cosmo features
* [X] Calculate volume based on the divergence theorem
* [ ] Fix sign in dielectrics
* [ ] Fix the issue with the epsilon, which is calculated differently as in the original COSMO papers (and seem to contain a factor of 1/2)


aradi:
Still active, depends on centrally generated full neighbor lists.

PR #1123, opened 2022-10-20
Unused variables for spin-resolved energy

bhourahine:
Note that the current energy structure for RS will break for MPI parallelism (it's a bit of a hack at the moment).

### DIFF
--- a/src/dftbp/dftb/rangeseparated.F90
+++ b/src/dftbp/dftb/rangeseparated.F90
@@ -86,12 +86,6 @@ module dftbp_dftb_rangeseparated
     !> total long range energy
     real(dp) :: lrEnergy
 
-    !> spin up part of energy
-    real(dp) :: lrEnergyUp
-
-    !> spin down part of energy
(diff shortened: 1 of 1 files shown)


vanderhe (approved):
The deleted variables are clearly unused. The way colinear spin is implemented right now is indeed a hack. I've at least solved the issues with MPI in my local version (my neighbor-list based algorithm for the Gamma-implementation is already MPI-parallelised), but it is still a hack.

bhourahine:
We should compare code changes. #1004 needs some more work, but I'll hopefully push a simple mixing version that works for the pGEMM algorithm in the next few days.

vanderhe:
Sounds great, at the moment my MPI parallelism mostly throws all resources at the HFX routines, is very memory intensive (for that reason also quite fast) and does not even support distributed matrices. For now this enables me to calculate reasonably sized systems extremely fast (even with thousands of k-points) but I'm not sure whether we chose this strategy in the long term. Let's discuss next devel. meeting.

PR #1132, opened 2022-10-30
Remove redundant Mulliken analysis call

bhourahine:
As there is always a later Mulliken analysis for the real space density matrix, the square matrix analysis in the range separation call is not needed.


vanderhe (approved):
You're right, it's actually long gone in my [local version](https://github.com/vanderhe/dftbplus/blob/ed9d30f637a620e198d863e511876511e6bed005/src/dftbp/dftbplus/main.F90#L3187-L3192).

PR #1133, opened 2022-11-02
Check for interaction between atoms in contacts

bhourahine:
Transport specific check


pecchia:
This is adds an important sanity check that was missing. Perhaps I would change a bit the error message:
"Atom such and such interact with each other" or "... have non-zero interaction", instead of "... are too close" because with the latter one might think of a relaxation error or something similar.  I do not see other issues and it can be merged.

PR #1134, opened 2022-11-02
Checks if transport device should be periodic

bhourahine:
Also adds some missing comments

### DIFF
--- a/src/dftbp/transport/negfint.F90
+++ b/src/dftbp/transport/negfint.F90
@@ -113,8 +113,6 @@ contains
     integer, allocatable :: sizes(:)
     type(lnParams) :: params
     character(lc) :: errString
-    integer :: iCont
-    real(dp), allocatable :: contVectors(:, :), sigma(:), U(:, :), Vt(:, :)
 
 #:if WITH_MPI
     call negf_mpi_init(env%mpi%globalComm)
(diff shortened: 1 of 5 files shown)


src/dftbp/transport/negfvars.F90
```diff
@@ -249,6 +249,9 @@ module dftbp_transport_negfvars
     !> type of phonon modes (Longitudinal, transverse)
     integer :: typeModes
 
+    !> Tolerance for contact vectors being distorted
+    real(dp) :: contactlayertol
```
aradi, line 253:
```suggestion
    real(dp) :: contactLayerTol
```

src/dftbp/transport/negfint.F90
```diff
@@ -135,6 +142,20 @@ subroutine TNegfInt_init(this, transpar, env, greendens, tundos, tempElec, coord
...
+    if (isPeriodic .and. nCont > 2) then
+      allocate(sigma(min(nCont,3)))
+      allocate(contVectors(3, nCont))
+      allocate(U(3, 3))
+      allocate(Vt(nCont, nCont))
+      do iCont = 1, nCont
+        contVectors(:, iCont) = transpar%contacts(iCont)%lattice
+      end do
+      call gesvd(contVectors, U, sigma, Vt)
+      if (all(sigma > transpar%contactLayerTol**2)) then
```
aradi, line 154: Could you factor that out into a function ˙hasFullySurroundingContacts` or something similar? Then we would not need all the auxiliary variables in the local scope.

src/dftbp/transport/negfvars.F90
bhourahine: Fixed

src/dftbp/transport/negfint.F90
bhourahine: Sure, done.

PR #1138, opened 2022-11-07
Fix for GFN2 with spin

bhourahine:
* Fix for multipole mixing when spin is present
* Remove SCC option for xTB, as should always be SCC
* Fixed logic for shell resolved spin constants
* Add test case


src/dftbp/dftbplus/parser.F90
```diff
@@ -1880,7 +1880,7 @@ subroutine readXTBHam(node, ctrl, geo, poisson)
     call getChildValue(node, "ShellResolvedSCC", ctrl%tShellResolved, .true.)
 
     ! SCC parameters
-    call getChildValue(node, "SCC", ctrl%tSCC, .true.)
+    ctrl%tSCC = .true.
```
awvwgk, line 1883: While not recommended it is in principle possible to disable the SCC with xTB methods. I think we should leave this option in.

bhourahine: Unfortunately Scc=No segfaults due to a bug in the energy evaluation (something not being correctly allocated, even when Mulliken charges are requested). I'll try and fix it within this PR.

bhourahine: Non-scc re-enabled and spin for non-SCC now disabled (non-scc + spin model was segfaulting, but also combination doesn't make much sense in TB).

awvwgk (approved):
Thanks. Looks good to me now.

PR #1139, opened 2022-11-07
Add stacksize information/recommendation to standard output

vanderhe:
Inspired by the FHIaims code and our discussion during the developers meeting a while ago, I've added stacksize information to the standard output. Let's hope that this facilitates using the DFTB+ code and helps to avoid requests like [this](https://github.com/dftbplus/dftbplus/discussions/1092) one. However, undercutting the recommended stacksize does not stop the code, it solely prints a warning for now.

We might want to discuss if putting C sources into the same folder as Fortran files is a good idea.

### DIFF
--- a/src/dftbp/common/envcheck.F90
+++ b/src/dftbp/common/envcheck.F90
@@ -1,0 +1,76 @@
+!--------------------------------------------------------------------------------------------------!
+!  DFTB+: general package for performing fast atomistic simulations                                !
+!  Copyright (C) 2006 - 2023  DFTB+ developers group                                               !
+!                                                                                                  !
+!  See the LICENSE file for terms of usage and distribution.                                       !
+!--------------------------------------------------------------------------------------------------!
+
+#:include 'common.fypp'
+#:include 'error.fypp'
+
+!> Contains functionality to check environment settings.
+module dftbp_common_envcheck
+
+  use, intrinsic :: iso_c_binding, only : c_char, c_int
+  use dftbp_common_accuracy, only : dp
+  use dftbp_common_environment, only : TEnvironment
+  use dftbp_common_globalenv, only : stdOut
+  use dftbp_io_message, only : warning
+  implicit none
+
+  private
+  public :: checkStackSize
+
+
+  interface
+    ! Interface for checking stacksize through C-routine.
+    subroutine get_stacksize(cStack, iErr) bind(C, name="get_stacksize_c")
+
+      import :: c_char, c_int
+      implicit none
+
+      !> Current stacksize
+      integer(kind=c_int), intent(out) :: cStack
+
+      !> Error status
+      integer(kind=c_int), intent(out) :: iErr
+
+    end subroutine get_stacksize
+  end interface
+
+
+contains
+
+  !> Checks stacksize settings for optimal user experience/performance.
+  subroutine checkStackSize(env)
+
+    !> Environment settings
+    type(TEnvironment), intent(in) :: env
+
+    !! Current stacksize
+    integer :: cStack
+
+    !! Error status
+    integer :: iErr
+
+    call get_stacksize(cStack, iErr)
+
+    if (iErr /= 0) then
+      write(stdOut, "(A,':',T30,A,I0,A)") "Current stacksize", "N/A (error code: ", iErr, ")"
+    else
+      if (cStack == -1 .or. cStack == 0) then
+        write(stdOut, "(A,':',T30,A)") "Current stacksize", "unlimited"
+      else
+        write(stdOut, "(A,':',T30,I0,A)") "Current stacksize",&
+            & nint(real(cStack, dp) / 1024.0_dp**2), " [Mb] (recommended: unlimited)"
+        call warning("Current stacksize not set to unlimited or hard limit, which might cause"&
+            & // new_line("A") // "   random crashes (e.g. segmentation faults). It is advised to&
+            & unlimit the" // new_line("A") // "   stacksize by issuing 'ulimit -s unlimited'&
+            & (Linux) or setting it to the " // new_line("A") // "   hard limit by 'ulimit -s hard'&
+            & (Mac) in advance.")
+      end if
+    end if
+
+  end subroutine checkStackSize
+
+end module dftbp_common_envcheck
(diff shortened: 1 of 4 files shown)


vanderhe:
Seems the Intel CI just broke...

src/dftbp/common/envcheck.c
```diff
@@ -0,0 +1,27 @@
+/*------------------------------------------------------------------------------------------------*/
+/*  DFTB+: general package for performing fast atomistic simulations                              */
+/*  Copyright (C) 2006 - 2022  DFTB+ developers group                                             */
+/*                                                                                                */
+/*  See the LICENSE file for terms of usage and distribution.                                     */
+/*------------------------------------------------------------------------------------------------*/
+
+#include <sys/resource.h>
```
bhourahine, line 8: Might have a `portability` issue](https://www.gnu.org/software/gnulib/manual/html_node/sys_002fresource_002eh.html)

vanderhe: That's a pity, does it affect any of our supported/tested build environments?

bhourahine: No, but the BSD packaging guy will complain if we don't include the other two sys headers.

vanderhe: I have no problem with adding the additional headers.

aradi: If, we would only need `sys/types.h` on BSD. But, I'd leave it for the moment as it is, it probably also works on BSD.

src/dftbp/common/envcheck.F90
```diff
@@ -0,0 +1,93 @@
...
+      if (cStack == -1 .or. cStack == 0) then
+        write(stdOut, "(A,':',T30,A)") "Current stacksize:", "unlimited"
+      else
+        write(stdOut, "(A,':',T30,I0,A)") "Current stacksize: ",&
+            & nint(real(mStack, dp) / 1024.0_dp**2), " [Mb] (recommended: unlimited)"
+        if (cStack < minStacksize) then
+          call warning("Current stacksize undercuts recommended minimum value, which might cause"&
+              & // NEW_LINE("A") // "   random crashes (e.g. segmentation faults). It is advised to&
+              & unlimit the" // NEW_LINE("A") // "   stacksize by issuing 'ulimit -s unlimited'&
+              & (Linux) or 'ulimit -s hard' (Mac)" // NEW_LINE("A") // "   in advance.")
```
aradi, line 86: `new_line` should be small (it is a fortran intrinsic)

aradi (commented):
I'd suggest to implement a different user experience: We should only print the current stack size, and issue a warning if not unlimited. I don't see the point in printing the maximum available size. Also using a threashold for the warning is questionable, as it may run without problems for small systems below the threshold and crash for larger above.

I'd just report the current stack size and issue a warning if the stack is not unlimited.

src/dftbp/common/envcheck.F90
```diff
@@ -0,0 +1,75 @@
...
+
+  use, intrinsic :: iso_c_binding, only : c_char, c_int
+  use dftbp_common_accuracy, only : dp
+  use dftbp_common_environment, only : TEnvironment
+  use dftbp_common_globalenv, only : stdOut
+  use dftbp_io_message, only : warning
+  implicit none
+
+  private
+  public :: checkStacksize
```
aradi, line 22:
```suggestion
  public :: checkStackSize
```

src/dftbp/common/envcheck.F90
```diff
@@ -0,0 +1,75 @@
...
+
+
+contains
+
+  !> Checks stacksize settings for optimal user experience/performance.
+  subroutine checkStacksize(env)
```
aradi, line 28:
```suggestion
  subroutine checkStackSize(env)
```

src/dftbp/common/envcheck.F90
```diff
@@ -0,0 +1,75 @@
...
+        implicit none
+
+        !> Current stacksize
+        integer(kind=c_int), intent(out) :: cStack
+
+        !> Error status
+        integer(kind=c_int), intent(out) :: iErr
+
+      end subroutine get_stacksize
+    end interface
```
aradi, line 53: Please move that in the preamble of the module.

src/dftbp/dftbplus/initprogram.F90
```diff
@@ -2917,6 +2918,8 @@ subroutine initProgramVariables(this, input, env)
       write(stdOut, "(A,':',T30,I0)") "Specified random seed", iSeed
     end if
 
+    call checkStacksize(env)
```
aradi, line 2921:
```suggestion
    call checkStackSize(env)
```

aradi (changes requested):

src/dftbp/common/envcheck.F90
vanderhe: done

src/dftbp/common/envcheck.F90
vanderhe: done

src/dftbp/common/envcheck.F90
vanderhe: done

src/dftbp/dftbplus/initprogram.F90
vanderhe: done

src/dftbp/common/envcheck.F90
```diff
@@ -0,0 +1,76 @@
+!--------------------------------------------------------------------------------------------------!
+!  DFTB+: general package for performing fast atomistic simulations                                !
+!  Copyright (C) 2006 - 2022  DFTB+ developers group                                               !
```
aradi, line 3:
```suggestion
!  Copyright (C) 2006 - 2023  DFTB+ developers group                                               !
```

aradi (approved):
Perfect, could you make the year change and then squash all commits and rebase? I'd then merge it.

src/dftbp/common/envcheck.F90
vanderhe: done

bhourahine:
Is it worth adding a regression test for this?

PR #1142, opened 2022-11-15
Implementation of LC-OC-DFTB

inseonglee:
Since the onsite correction is derived from the XC functional, the long-range onsite correction terms are generated from the long-range HF exchange kernel.
The equations for energy, Hamiltonian, and gradients can be found in following doi: 10.1021/acs.jctc.2c00037.


src/dftbp/dftb/rangeseponscorr.F90
```diff
@@ -0,0 +1,679 @@
...
+
+            ! Find l-shell for jOrb
+            iSh2 = orb%iShellOrb(jOrb-ii+1,species(iAtom))
+
+            ! OC contribution except diagonal elements
+            if (iOrb /= jOrb) then
+              this%Omat0(iOrb,jOrb) = onSiteElements(iSh1,iSh2,3,species(iAtom))
+            end if
+
+            ! RI loss correction for p orbitals
```
bhourahine, line 148: Is there a way to generalise for other angular momentum? The original onsite code was coded for general l and worked for d and probably f. If it is not possible, there needs to be trapping for attempting to use the correction with un-supported angular momentum values.

src/dftbp/dftbplus/initprogram.F90
```diff
@@ -5453,6 +5477,97 @@ subroutine initRangeSeparated(this, nAtom, species0, hubbU, rangeSepInp, tSpin,
...
+      call error("Onsite correction with range separated calculations not currently&
+          & implemented for linear response excitation")
+    end if
+
+    if (allocated(solvation)) then
+      call error("Onsite correction with range separated calculations not currently&
+          & implemented for solvation")
+    end if
+
+    if (allocated(reks)) then
```
bhourahine, line 5563: Is the plan to add the REKS implementation described in the paper as a separate pull request?

bhourahine (commented):
There seems to be a loss of functionality, as it is now not possible to use the old style onsite corrections with range separation, only the new form. There needs to be the option to use the original on-site correction without the lc-onsites (or does setting the related constants to zero effectively do this? Is so this should be documented). There also needs to be test cases with the old and new forms, instead of replacing the old form.

Additionally, could you please add documentation in the manual for the new constants. The added keyword for the lc constants should be described with a reference to your paper, but also a table in an appendix of your fitted values to make it self-contained.

Please also update the CHANGELOG file

The LC onsite terms will  also need be available for the non-REKS cases, or to be blocked from being used. There is energy evaluation, but not the  hamiltonian contribution for the time propagation routine (timedep/timeprop.F90), the conventional hamiltonian perturbation (derivs/perturb.F90) also can use onsite and range separation corrections.

inseonglee:
> There seems to be a loss of functionality, as it is now not possible to use the old style onsite corrections with range separation, only the new form. There needs to be the option to use the original on-site correction without the lc-onsites (or does setting the related constants to zero effectively do this? Is so this should be documented). There also needs to be test cases with the old and new forms, instead of replacing the old form.

You mean that the user have to choose the option among the old and new onsite corrections regardless of rangeSep option.
Then, the old onsite correction (use uu, ud) with LC functional (use rangeSep) can be used (LC-frOC-DFTB). Also, the new onsite correction (use uu, ud, lc) with conventional functional (without rangeSep) can be used (frOC+lrOC-DFTB). Am I correct?

I think that if we consider the onsite correction formalism, then all possible onsite terms should be simultaneously included.
Since the onsite correction term is derived from the inclusion of onsite integrals to general four center integrals (full-range or long-range functionals), for the consistency of the level of theory, the upper cases (e.g. onsite correction: PBE, rangeSep: use LC) are somewhat strange. That's why I blocked the wrong use for LC and OC formalisms. In current commit, only OC-DFTB (PBE functional) or LC-OC-DFTB (LC functional for both LC and OC terms) can be used.

inseonglee:
> Additionally, could you please add documentation in the manual for the new constants. The added keyword for the lc constants should be described with a reference to your paper, but also a table in an appendix of your fitted values to make it self-contained.

> Please also update the CHANGELOG file

For the documentation and CHANGELOG, I will add new explanations and commits for this.

inseonglee:
> The LC onsite terms will also need be available for the non-REKS cases, or to be blocked from being used. There is energy evaluation, but not the hamiltonian contribution for the time propagation routine (timedep/timeprop.F90), the conventional hamiltonian perturbation (derivs/perturb.F90) also can use onsite and range separation corrections.

Current PR adds LC onsite terms, so I will check another compatibility for time propagation or perturbation.
At the moment, this PR is not relevant to REKS, so the use of both REKS and LC-OC is temporarily blocked.

test/app/dftb+/rangesep/Thymine-T1_onsite/dftb_in.hsd
```diff
@@ -81,24 +81,31 @@ Hamiltonian = DFTB {
...
+    Ouu = { 0.00000  0.07858
+            0.07858 -0.00665 }
+    Oud = { 0.00000  0.16395
+            0.16395  0.05179 }
+    Olc = { 0.00000  0.15181
+            0.15181 -0.01071 }
   }
 
   RangeSeparated = LC {
-    Screening = NeighbourBased {}
```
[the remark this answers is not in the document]
inseonglee: Since the current implementation for LC-onsite supports only matrix-based algorithm, so I changed the algorithm for this test.
Please notice that the same algorithm used in RangeSeparated block is exploited to evaluate the energy and Hamiltonian for LC-onsite term.

src/dftbp/dftb/rangeseponscorr.F90
inseonglee: Yes, there is a generalisation for other angular momentum. It can be extended to d or f orbitals, but the exact equations are not derived at the moment. I will temporarily add new constraint for the d or f angular momentum.

src/dftbp/dftbplus/initprogram.F90
inseonglee: Yes, after this PR is merged, then I will create new PR for REKS implementation with OC formalism as the paper described.

bhourahine:
> I think that if we consider the onsite correction formalism, then all possible onsite terms should be simultaneously included.

Unfortunately this causes problems for backward compatibility. The old (incorrect) behaviour should still be possible, but with a warning. Then in the longer term, it can be depreciated in a later release and then removed (using the oldcompat.F90 mechanism).

test/app/dftb+/rangesep/Thymine-T1_onsite/dftb_in.hsd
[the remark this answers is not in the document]
inseonglee: @bhourahine I reverted this test to original version (LC-frOC-DFTB), please see commit 2f83a86. Also, same thing is applied to deltadftb/H2O test.

inseonglee:
> Unfortunately this causes problems for backward compatibility. The old (incorrect) behaviour should still be possible, but with a warning. Then in the longer term, it can be depreciated in a later release and then removed (using the oldcompat.F90 mechanism).

To be able to use LC-frOC-DFTB as well as LC-OC-DFTB, I modify the code. Do I modify correctly for your intention? Please see commit 2f83a86.

inseonglee:
For time propagation and hamiltonian perturbation, I blocked these calculations with use of long-range OC. I just expressed `TODO` in the code for later implementation. If there are additional conditions that I have to consider, please let me know.

src/dftbp/dftb/rangeseponscorr.F90
inseonglee: @bhourahine The constraint for maximum angular momentum is added, please see the commit 2d7b4eb.

inseonglee:
@bhourahine Now, all comments are applied, please review this PR again.

inseonglee:
@bhourahine @aradi Please review this PR. After this PR is merged, then I will implement OC with SSR method.

inseonglee:
@bhourahine Long time for update on this PR! Because there are many updates on main branch (about hybrid functional), some missing points may exist in the source codes. Could you check whether the synchronization correctly work or not? Thanks in advance.

Additionally, I do not change the name of a new added class 'rsOnsCorr'. Should I change the name according to the change of the main branch such as 'hybridOnsCorr'?

inseonglee:
In my cluster (centOS, intel compiler), all regression tests are passed.

bhourahine:
@inseonglee I'm away from my office with limited internet connection this week. @aradi may be able to comment on the intel CI failure, but the gcc CI seems to be something else.

inseonglee:
After synchronization of main branch, the intel-CI successfully works but the gcc-CI still fails.

inseonglee:
@bhourahine @aradi Now this PR is ready to merge from my side. Could you check why the gcc-CI build fails? In my intel clusters, all regression tests are passed.

PR #1162, opened 2022-12-23
Add input flag to (de)activate writing charges to disc

### LINKED ISSUE #1088: It is necessary to save charges.bin at every step?
Not sure if it is a feature request. I noticed that the charges.bin file will be saved at the end of each SCF cycle. This is fine. But I just want to avoid the unnessary I/o load during the geometry optimization. Is there any keywords from `dftb_in.hsd` related to this?

I tried `RestartFrequency`, but it does not help for my case.

**Describe alternatives you've considered**

**Additional context**

vanderhe:
Adds option to control whether restart file (charges.{bin|dat}) shall be written to disc:
`Options = {
  WriteCharges = Yes|No
}`

Tasks:
- [x]  add input flag to (de)activate writing charges to disc
- [x] documentation

:negative_squared_cross_mark: regression tests (charges.{bin|dat} not tested)

Presumably closes #1088.

### DIFF
--- a/src/dftbp/dftbplus/parser.F90
+++ b/src/dftbp/dftbplus/parser.F90
@@ -3890,11 +3890,8 @@ contains
     call getChildValue(node, "WriteDetailedOut", ctrl%tWriteDetailedOut, tWriteDetailedOutDef)
 
     call getChildValue(node, "WriteAutotestTag", ctrl%tWriteTagged, .false.)
-    call getChildValue(node, "WriteDetailedXML", ctrl%tWriteDetailedXML, &
-        &.false.)
-    call getChildValue(node, "WriteResultsTag", ctrl%tWriteResultsTag, &
-        &.false.)
-
+    call getChildValue(node, "WriteDetailedXML", ctrl%tWriteDetailedXML, .false.)
+    call getChildValue(node, "WriteResultsTag", ctrl%tWriteResultsTag, .false.)
 
     if (.not.(ctrl%tMD.or.ctrl%isGeoOpt.or.allocated(ctrl%geoOpt))) then
       if (ctrl%tSCC) then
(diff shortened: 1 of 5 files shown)


vanderhe:
@aradi @bhourahine Would this extension of the `Options {}` block be something we could live with?

bhourahine:
Not sure how to regression test this one, you presumably could use the script option in the regression directories to run the input twice (once with and once without generation, just testing for file existence in the not, but re-loading in the true case into a final calculation).

src/dftbp/dftbplus/parser.F90
```diff
@@ -3925,7 +3922,13 @@ subroutine readOptions(node, ctrl)
     if (ctrl%tReadChrg) then
       call getChildValue(node, "ReadChargesAsText", ctrl%tReadChrgAscii, .false.)
     end if
-    call getChildValue(node, "WriteChargesAsText", ctrl%tWriteChrgAscii, .false.)
+
+    call getChildValue(node, "WriteCharges", ctrl%tWriteCharges, .true.)
+    call getChildValue(node, "WriteChargesAsText", ctrl%tWriteChrgAscii, .false., child=child)
```
bhourahine, line 3927: Should only need to check for this flag (and the next if block) if tWriteCharges == T

src/dftbp/dftbplus/parser.F90
```diff
@@ -3925,7 +3922,13 @@ subroutine readOptions(node, ctrl)
...
+    if (ctrl%tWriteChrgAscii .and. (.not. ctrl%tWriteCharges)) then
+      call detailedError(child, "Charges requested in Ascii format, but writing out charges is&
+          & disabled (contradictory input).")
+    end if
```
vanderhe, line 3931:
```suggestion
    call getChildValue(node, "WriteCharges", ctrl%tWriteCharges, .true.)
    if (ctrl%tWriteCharges) then
      call getChildValue(node, "WriteChargesAsText", ctrl%tWriteChrgAscii, .false.)
    end if
```

vanderhe: @bhourahine Like this?

bhourahine: @vanderhe sure, you could include the ascii if test that follows it as well, but that's harmless as is.

bhourahine (approved):
Do you wantt to squash to one commit (including the line 3926 suggestions) and merge? I guess we add a regresion test as a separate PR.

vanderhe:
Squashed and rebased on main, ready to get merged from my side.

PR #1167, opened 2023-01-13
Add RSH-DFTB to the list of publications to cite

vanderhe:
RSH-DFTB is added to the list of publications to cite (appendix).


bhourahine (approved):
Probably should work out a sensible ordering for the table entries at some stage.

PR #1169, opened 2023-01-20
Use workload-aware chunking according to num neighbors where possible

terminationshock:
Enhance chunking by considering that every MPI rank has a different workload due to the number of neighbours. Also try to use this chunking, where it is applicable.

### DIFF
--- a/src/dftbp/common/schedule.F90
+++ b/src/dftbp/common/schedule.F90
@@ -23,8 +23,7 @@ module dftbp_common_schedule
   implicit none
 
   private
-  public :: distributeRangeInChunks, distributeRangeInChunks2,&
-      & distributeRangeWithWorkload
+  public :: distributeRangeInChunks, distributeRangeInChunks2, distributeRangeWithWorkload
   public :: assembleChunks, getChunkRanges, getChunkIterWithWorkload
   public :: TChunkIterator
 
@@ -161,8 +160,8 @@ contains
     type(TChunkIterator), intent(out) :: chunkIter
 
   #:if WITH_MPI
-    call getChunkIterWithWorkload(env%mpi%groupComm%size, env%mpi%groupComm%rank, globalFirst, globalLast,&
-        & workload, chunkIter)
+    call getChunkIterWithWorkload(env%mpi%groupComm%size, env%mpi%groupComm%rank, globalFirst,&
+        & globalLast, workload, chunkIter)
   #:else
     call getChunkIterWithWorkload(1, 0, globalFirst, globalLast, workload, chunkIter)
   #:endif
@@ -228,7 +227,8 @@ contains
 
 
   !> Calculate the chunk ranges for a given MPI-communicator considerung different workload
-  subroutine getChunkIterWithWorkload(groupSize, myRank, globalFirst, globalLast, workload, chunkIter)
+  subroutine getChunkIterWithWorkload(groupSize, myRank, globalFirst, globalLast, workload,&
+        & chunkIter)
 
     !> Size of the group over which the chunks should be distributed
     integer, intent(in) :: groupSize
@@ -248,7 +248,7 @@ contains
     !> The chunk iterator
     type(TChunkIterator), intent(out) :: chunkIter
 
-    integer :: numIndices, rank, i
+    integer :: numIndices, rank, ii
     integer, allocatable :: rankWorkload(:), indices(:)
 
     allocate(rankWorkload(groupSize))
@@ -258,12 +258,12 @@ contains
     indices(:) = 0
     numIndices = 0
 
-    do i = globalFirst, globalLast
+    do ii = globalFirst, globalLast
       rank = minloc(rankWorkload, dim=1)
-      rankWorkload(rank) = rankWorkload(rank) + max(1, workload(i))
+      rankWorkload(rank) = rankWorkload(rank) + max(1, workload(ii))
       if (rank == myRank + 1) then
         numIndices = numIndices + 1
-        indices(numIndices) = i
+        indices(numIndices) = ii
       end if
     end do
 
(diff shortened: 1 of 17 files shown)


bhourahine:
Unfortunately the git actions don't test with more than one openMP thread, and various regression tests fail for assorted compilers when tested with -DTEST_OMP_THREADS=2 (either with or without MPI).

terminationshock:
@bhourahine In the OpenMP case, the `getNextIndex` didn't make much sense. I did a thread-safe implementation now, which passes all tests on my side.

src/dftbp/common/schedule.F90
```diff
@@ -177,4 +227,144 @@ subroutine getChunkRanges(groupSize, myRank, globalFirst, globalLast, localFirst
...
+    allocate(rankWorkload(groupSize))
+    allocate(indices(globalLast - globalFirst + 1))
+
+    rankWorkload(:) = 0
+    indices(:) = 0
+    numIndices = 0
+
+    do i = globalFirst, globalLast
+      rank = minloc(rankWorkload, dim=1)
+      rankWorkload(rank) = rankWorkload(rank) + max(1, workload(i))
```
aradi, line 263: Why `max(1, workload(i))`? Wouldn't that be equivalent to what we have without workload considerations, as `rankWorkload(rank)` will be increased by 1 only?

src/dftbp/common/schedule.F90
```diff
@@ -177,4 +227,144 @@ subroutine getChunkRanges(groupSize, myRank, globalFirst, globalLast, localFirst
...
+    !> Instance
+    class(TChunkIterator), intent(out) :: this
+
+    !> Index list
+    integer, intent(in) :: indices(:)
+
+    @:ASSERT(size(indices) > 0)
+    @:ASSERT(all(indices > 0))
+
+    allocate(this%indices(size(indices)))
```
aradi, line 287: This is line is superfluous. The next line uses the automatic reallocation of the LHS, so that `this%indices` will be deallocated and reallocated in the next line again. (Alternatively, you might use `this%indices(:) = indices` in the next line, but why using two lines for something, which can be written also in one....)

src/dftbp/common/schedule.F90
```diff
@@ -177,4 +227,144 @@ subroutine getChunkRanges(groupSize, myRank, globalFirst, globalLast, localFirst
...
+
+  end function TChunkIterator_getNextIndex
+
+
+  !> Returns a certain item from the underlying index array.
+  !> This routine is thread safe and can be used in OpenMP-parallelized loops.
+  function TChunkIterator_getIndex(this, item) result(nextIndex)
+
+    !> Instance
+    class(TChunkIterator), intent(inout) :: this
```
aradi, line 355:
```suggestion
    class(TChunkIterator), intent(in) :: this
```

src/dftbp/dftb/shortgamma.F90
```diff
@@ -743,45 +744,60 @@ subroutine buildShifts_(env, orb, species, iNeighbours, img2CentCell, hubb, nNei
...
+            do iSh2 = 1, maxShell2
+              iU2 = hubb%iHubbU(iSh2, iSp2)
+              shortGammaValue = shortGamma(iU2, iU1, iNeigh, iAt1)
+              shiftShell(iSh1, iAt1) = shiftShell(iSh1, iAt1)&
+                    & - shortGammaValue * deltaQShell(iSh2, iAt2f)
+              if (iAt2f /= iAt1) then
+                shiftShell(iSh2, iAt2f) = shiftShell(iSh2, iAt2f)&
+                    & - shortGammaValue * deltaQShellValue
+              end if
+            end do
```
aradi, line 796: I have some difficulties to understand theoretically, why the loop order had to be switched. We should definitely have a quick discussion on this.

aradi (commented):
Very nice idea! I have several remarks, and two questions we should discuss at the next meeting.

Additionally, we should discuss, whether it wouldn't be practical to use the thread-safe / functional programming version everywhere (with `do` instead of a `do while` loop). And maybe with slightly different naming?
```fortran
type(TChunkList) :: chunks
! integer, allocatable :: chunks(:)

call getChunkIterWithWorkload(..., chunks)

do iChunk = 1, chunks%size()
! do iChunk = 1, size(chunks)
  iAt = chunks(iChunk)
  ...
end do

File: src/dftbp/common/schedule.F90 [hunk 5]
```diff
@@ -177,4 +227,144 @@ subroutine getChunkRanges(groupSize, myRank, globalFirst, globalLast, localFirst
...
+    !> Last element of the range
+    integer, intent(in) :: globalLast
+
+    !> Workload for each item
+    integer, intent(in) :: workload(:)
+
+    !> The chunk iterator
+    type(TChunkIterator), intent(out) :: chunkIter
+
+    integer :: numIndices, rank, i
```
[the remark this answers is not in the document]
terminationshock: Thanks, I didn't know that convention.

src/dftbp/common/schedule.F90
terminationshock: I wanted to make sure that if the workload is zero that the chunked loop is executed anyway, because there might be some code parts not depending on the workload. Think about:
```
do while (chunkIter%hasNextIndex())
  iAtom1 = chunkIter%getNextIndex()
  do iNeigh = 1, nNeighbourSK(iAtom1)
    ! Assume that nNeighbourSK(iAtom1) is zero
  end do
  call someOtherRoutine()
end do
```

aradi: Sorry, I should not review PRs in the evening. I thought, it is `min()`. It all makes sense of course!

terminationshock:
@aradi As discussed yesterday, I have now removed the iterator and consistently use do loops everywhere.

PR #1171, opened 2023-01-20
Align Fortran arrays by calling posix_memalign

terminationshock:
Call `posix_memalign` to align arrays to the `n` byte boundary (usually used as 64 bytes on x86-64). Although the Intel Fortran compiler has a flag `-align array64byte`, which forces this alignment globally, nothing like this exists for the GNU compiler (and others). This implementation allows alignment independent of the compiler.

### DIFF
--- a/src/dftbp/common/memman.F90
+++ b/src/dftbp/common/memman.F90
@@ -5,14 +5,64 @@
 !  See the LICENSE file for terms of usage and distribution.                                       !
 !--------------------------------------------------------------------------------------------------!
 
+#:include 'common.fypp'
+
 !> Contains various constants for memory management
 module dftbp_common_memman
   use dftbp_common_accuracy, only : dp
+  use dftbp_io_message, only : error, warning
+  use, intrinsic :: iso_c_binding, only : c_ptr, c_null_ptr, c_f_pointer,&
+      & c_int, c_intptr_t, c_size_t
 
   implicit none
 
   private
-  public :: incrmntOfArray
+  public :: incrmntOfArray, TAlignedArray
+
+  !> Holds an allocated array being aligned in memory
+  type TAlignedArray
+
+    !> Actual array data
+    real(dp), pointer :: array(:)
+
+    !> Number of array elements
+    integer :: size = 0
+
+    !> Chosen alignment in bytes (default: 64 byted)
+    integer :: alignment = 64
+
+    !> Whether the array is allocated
+    logical :: isAllocated = .false.
+
+    !> Pointer to the allocated memory
+    type(c_ptr) :: memoryPointer = c_null_ptr
+
+  contains
+
+    procedure :: allocateAligned => TAlignedArray_allocateAligned
+    final :: TAlignedArray_finalize
+
+  end type TAlignedArray
+
+
+  !> Bound to 'posix_memalign' to allocate aligned memory
+  interface
+    function posixMemalign(ptr, alignment, size) result(error) bind(C, name="posix_memalign")
+      import c_ptr, c_intptr_t, c_int
+      type(c_ptr), intent(inout) :: ptr
+      integer(c_intptr_t), intent(in), value :: alignment, size
+      integer(c_int) :: error
+    end function
+  end interface
+
+
+  !> Bound to 'free' to deallocate memory again being previously allocated using 'posix_memalign'
+  interface
+    subroutine free(ptr) bind(C, name="free")
+      import c_ptr
+      type(c_ptr), value :: ptr
+    end subroutine
+  end interface
 
 
 contains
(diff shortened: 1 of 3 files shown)


src/dftbp/common/memman.F90
```diff
@@ -5,14 +5,64 @@
...
+  !> Holds an allocated array being aligned in memory
+  type TAlignedArray
+
+    !> Actual array data
+    real(dp), pointer :: array(:)
+
+    !> Number of array elements
+    integer :: size = 0
+
+    !> Chosen alignment in bytes (default: 64 byted)
```
bhourahine, line 31:
```suggestion
    !> Chosen alignment in bytes (default: 64 bytes)
```

src/dftbp/common/memman.F90
```diff
@@ -30,4 +80,59 @@ pure function incrmntOfArray(currentSize)
 
   end function incrmntOfArray
 
+
+  !> Allocate an array aligned to the n byte boundary (typically used for aligning to 64 byte)
```
bhourahine, line 84:
```suggestion
  !> Allocate an array aligned to the n byte boundary (typically used for aligning to 64 bytes)
```

test/src/dftbp/unittests/common/memman.F90
```diff
@@ -0,0 +1,73 @@
+!--------------------------------------------------------------------------------------------------!
+!  DFTB+: general package for performing fast atomistic simulations                                !
+!  Copyright (C) 2006 - 2022  DFTB+ developers group                                               !
```
bhourahine, line 3:
```suggestion
!  Copyright (C) 2006 - 2023  DFTB+ developers group                                               !
```

src/dftbp/common/memman.F90
```diff
@@ -30,4 +80,59 @@ pure function incrmntOfArray(currentSize)
...
+    integer, intent(in), optional :: alignment
+
+    if (this%isAllocated) then
+      call error("Aligned array is already allocated")
+    end if
+
+    @:ASSERT(size > 0)
+
+    if (present(alignment)) then
+      @:ASSERT(alignment >= dp)
```
aradi, line 103: You can not use `dp` as the number of bytes used to represent a given type, as it can have any compiler dependent positive value. (Many compilers choose 8 for doubles, but NAG uses 2 for example...). You could use `c_sizeof` instead.

src/dftbp/common/memman.F90
```diff
@@ -5,14 +5,64 @@
...
 
   private
-  public :: incrmntOfArray
+  public :: incrmntOfArray, TAlignedArray
+
+  !> Holds an allocated array being aligned in memory
+  type TAlignedArray
+
+    !> Actual array data
+    real(dp), pointer :: array(:)
```
aradi, line 26:
```suggestion
    real(dp), pointer :: array(:) => null()
```

src/dftbp/common/memman.F90
```diff
@@ -5,14 +5,64 @@
...
+
+    !> Number of array elements
+    integer :: size = 0
```
aradi, line 29: I'd leave it away. User can query `size(instance%array)` to find out the size.

src/dftbp/common/memman.F90
```diff
@@ -5,14 +5,64 @@
...
+
+    !> Chosen alignment in bytes (default: 64 bytes)
+    integer :: alignment = 64
+
+    !> Whether the array is allocated
+    logical :: isAllocated = .false.
```
aradi, line 35: Again, this is redundant IMO. Querying `associated(instance%array)` would do the same.

src/dftbp/common/memman.F90
```diff
@@ -5,14 +5,64 @@
...
+
+    !> Pointer to the allocated memory
+    type(c_ptr) :: memoryPointer = c_null_ptr
```
aradi, line 38: Let's make it `private`, we should only use the Fortran pointer, unless if possible.

src/dftbp/common/memman.F90
```diff
@@ -5,14 +5,64 @@
...
+
+  contains
+
+    procedure :: allocateAligned => TAlignedArray_allocateAligned
```
aradi, line 42: It is a matter of taste, and we can discuss, but I'd just call this `%allocate()`  instead of `%allocateAligned()`, as you can not allocate this array other than aligned.

src/dftbp/common/memman.F90
```diff
@@ -30,4 +80,59 @@ pure function incrmntOfArray(currentSize)
...
+    this%size = size
+    call c_f_pointer(this%memoryPointer, this%array, [size])
+
+    this%isAllocated = .true.
+
+  end subroutine TAlignedArray_allocateAligned
+
+
+  !> Clean-up an aligned array by deallocating it
+  subroutine TAlignedArray_finalize(this)
```
aradi, line 123: Note: if someone creates an array of `type(TAlignedArray)` , the destructor will not be called when it goes out of scope! You should make it `elemental impure` to have it invoked for any ranks.

aradi (commented):
Thanks a lot! A few remarks in addition to the comments so far:
- We should guard the type again assignment, either by deallocating and reallocating on assignment (which would be my preference) or by just throwing an error.
-  We should consider to give only indirect access to the Fortran pointer (via a getter) to make sure, it can not be changed from outside (which would end up in a desaster).
- One could think to drop the storage of the C-ptr as one can regain it back using `c_loc()`, if needed (as for the finalization):

terminationshock:
@aradi I reworked the implementation. To make it even cleaner, I call `c_f_pointer` when `getArray` is invoked, so the caller always get a pointer, which can be safely nullified (see the test). The array remains allocated and will be freed when getting out of scope. I also added an assignment procedure that re-allocates and copies the data. Please feel free to criticize if you do not agree.

src/dftbp/common/memman.F90
```diff
@@ -93,43 +90,76 @@ subroutine TAlignedArray_allocateAligned(this, size, alignment)
...
+
     this%size = size
-    call c_f_pointer(this%memoryPointer, this%array, [size])
 
-    this%isAllocated = .true.
+  end subroutine TAlignedArray_allocate
+
+
+  !> Fetch the pointer to the array data
+  subroutine TAlignedArray_getArray(this, array)
```
aradi, line 122: Yes, this is definitely the most robust solution. One stylistic remark (open for discussion): one could also implement this as function, so that on the caller side you had 
```
myArray => aligned%array()
```
instead of 
```
call aligned%getArray(myArray)
```.
I like the first one more from the optics, but it is more error prone, as if somebody writes `=` instead of `=>`, it will warrantedly crash at run-time. So, probably let's stay with the subroutine....

File: src/dftbp/common/memman.F90 [hunk 12]
```diff
@@ -5,14 +5,61 @@
...
+  type TAlignedArray
+
+    !> Number of array elements
+    integer :: size = 0
+
+    !> Chosen alignment in bytes (default: 64 bytes)
+    integer :: alignment = 64
+
+    !> Pointer to the allocated memory
+    type(c_ptr), private :: memoryPointer = c_null_ptr
```
aradi, line 32: I tend to add a `_` suffix to the private fields of a derived type.
```suggestion
    type(c_ptr), private :: memoryPointer_ = c_null_ptr
```

aradi (commented):
I like the solution a lot. I have only minor suggestions / remarks left.

src/dftbp/common/memman.F90
terminationshock: I was also thinking about that, because we are using it in ELPA. However, I do not like it, because of the exact reason you have mentioned: If the user just writes a `=`, then everything gets messed up. The subroutine is safer for sure.

PR #1173, opened 2023-01-23
Make contact shift reading/writing symmetric

aradi:
Binary stream I/O would break in original implementation as read/write are not symmetric. Also, make binary reading more robust by always reading the block charge block if present, as otherwise an eventually following Ef-read would deliver an incorrect value.


bhourahine (approved):
LGTM

PR #1181, opened 2023-01-30
Change ELPA mode to 2 to fix test

terminationshock:
The mode should be 2, otherwise the test is exactly equal to `Si384_ELPA1`.


bhourahine (approved):
OK to merge

PR #1182, opened 2023-01-30
Fix mpi-win finalization problem

aradi:
Seems to fix the error we get with the recent Intel compilers for MPI-applications, when `mpi_finalize()` is invoked: The mpi-windows will be freed explicitely before the call, allowing `mpi_finalize()` to run without any problems.

* [X] Test with buildbot (successful)

### DIFF
--- a/app/dftb+/dftbplus.F90
+++ b/app/dftb+/dftbplus.F90
@@ -22,7 +22,7 @@ program dftbplus
 
   type(TEnvironment) :: env
   type(TInputData), allocatable :: input
-  type(TDftbPlusMain), allocatable :: main
+  type(TDftbPlusMain), allocatable, target :: main
 
   call initGlobalEnv()
   call printDftbHeader(releaseName, releaseYear)
--- a/app/dftb+/dftbplus.F90
+++ b/app/dftb+/dftbplus.F90
@@ -22,17 +22,19 @@ program dftbplus
 
   type(TEnvironment) :: env
   type(TInputData), allocatable :: input
-  type(TDftbPlusMain) :: main
+  type(TDftbPlusMain), allocatable :: main
 
   call initGlobalEnv()
   call printDftbHeader(releaseName, releaseYear)
   allocate(input)
   call parseHsdInput(input)
   call TEnvironment_init(env)
+  allocate(main)
   call main%initProgramVariables(input, env)
   deallocate(input)
   call runDftbPlus(main, env)
   call main%destructProgramVariables()
+  deallocate(main)
 #:if WITH_MAGMA
   call magmaf_finalize()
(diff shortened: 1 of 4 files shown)


src/dftbp/dftbplus/initprogram.F90
```diff
@@ -1839,12 +1839,9 @@ subroutine initProgramVariables(this, input, env)
     else
       this%tCasidaForces = .false.
     end if
-    if (this%tSccCalc) then
-      this%forceType = input%ctrl%forceType
-    else
-      if (input%ctrl%forceType /= forceTypes%orig) then
-        call error("Invalid force evaluation method for non-SCC calculations.")
-      end if
+    this%forceType = input%ctrl%forceType
```
bhourahine, line 1842: Is this part of the changes needed for the PR, or a stray fix?

src/dftbp/io/hsdparser.F90
```diff
@@ -185,6 +185,7 @@ subroutine parseHSD_opened(initRootName, fd, file, xmlDoc)
     !> DOM-tree of the parsed input on exit
     type(fnode), pointer :: xmlDoc
 
+    logical, parameter :: parsedTypes(nSeparator) = .true.
```
bhourahine, line 188: Likewise, is this needed to resolve the finalizer problem, or should it be its own PR?

app/dftb+/dftbplus.F90
```diff
@@ -22,17 +22,19 @@ program dftbplus
...
   call printDftbHeader(releaseName, releaseYear)
   allocate(input)
   call parseHsdInput(input)
   call TEnvironment_init(env)
+  allocate(main)
   call main%initProgramVariables(input, env)
   deallocate(input)
   call runDftbPlus(main, env)
   call main%destructProgramVariables()
+  deallocate(main)
```
bhourahine, line 37: I suspect this will also help with finalizing other things, so definitely should get merged.

bhourahine (approved):
See comments, is this all one PR, or actually 3 separate ones together?

src/dftbp/dftbplus/initprogram.F90
aradi: It is needed, as tests fail without it. It was always buggy, but by accident `this%forceType` was always zero. By putting the `TDftbPlus` instance on the heap rather than on the stack the uninitialized integer became usually non-zero, causing several non-scc tests to fail.

app/dftb+/dftbplus.F90
aradi: Actually, we still have memory leaks (despite this), part of that will be fixed in a followup PR.

PR #1187, opened 2023-02-10
Nacv only

thomas-niehaus:
Non-adiabatic coupling vectors


thomas-niehaus:
Closed this pull request because main was too difficult to merge + numerical issues on some platfor
ms that need to be fixed first.

PR #1188, opened 2023-02-13
Fix dp_dos for non-collinear cases

aradi:


tools/dptools/src/dptools/bandout.py
```diff
@@ -62,35 +65,35 @@ def fromfile(cls, fobj):
...
             tmp = np.array(vals.split(), dtype=float)
-
-            # Only keep last two data columns
-            # (optional first column, if present, contains sequential numbering)
             nrows = vals.strip().count('\n') + 1
             ncols = len(tmp) // nrows
-            tmp.shape = (nrows, ncols)
-            tmp = tmp[:, ncols - 2 : ncols]
 
+            # Optional first row might contain sequential numbering.
```
charlymedrano7, line 80: First row or first column? I am not sure if I understand correctly the following block. For me it makes sense if the first *column* would be sequential numbers. If is that the case and I understood correctly, the script just skip that first column, right? My second question is, when do we have those sequential numbering? Is a matter of previous versions output?

charlymedrano7 (commented):
I tested already in some non-collinear systems and seems to work fine. When I artificially created a pdos.out with 7 columns it shows me the proper error.
My only question is in a comment made in the bandout.py that I am not sure if I understood correctly.

tools/dptools/src/dptools/bandout.py
aradi: That's a typo, I meant first column of course. Thx, corrected.

As for the sequential numbering: Eigenlevels in `band.out` are numbered for easier interpretation by humans. In the `ProjectedStates` output files, however, this enumeration is missing. Therefore, the script must be able to handle both cases.

charlymedrano7: You are welcome. Understood. Thanks, Bálint!

PR #1189, opened 2023-02-13
Ensure data dealloc in TNeighbourList without MPI-windows

aradi:


terminationshock (approved):
Thanks for this bugfix. I came across a similar thing in another code recently...
This Commit will cause a merge conflict in my merge request https://github.com/dftbplus/dftbplus/pull/1170, so I will rebase this one.

aradi:
OK, it needs some more refactoring in order to avoid exposition of internal details of the neighbor list. I am working on it.

aradi:
@terminationshock In order to prevent exposing the internal details of the neighbor list update, I have moved the actual `setNeighbourList()` into the same module, where `TNeighbourList` is defined. As far as I see, I did not break anything. Nevertheless, could you eventually check, whether the API works for your integration?

src/dftbp/dftb/periodic.F90
```diff
@@ -592,6 +607,187 @@ subroutine updateNeighbourList(coord, img2CentCell, iCellVec, neigh, nAllAtom, c
...
+
+      call distributeRangeInChunks(env, nAtom + 1, nAllAtom, iAtFirst, iAtLast)
+
+      do iAtom = iAtFirst, iAtLast
+        iImage = img2CentCell(iAtom)
+        diff(:) = coord0(:,iImage) - coord(:,iAtom)
+        do iCell = 1, nCellVec
+          iCellFlat = 3 * (iCell - 1)
+          ! Note that we need an explicit loop unrolling here to assist vectorization.
+          ! Using the pointer rCellVec helps the compiler to assume continuous memory access.
```
terminationshock, line 717:
```suggestion
          ! Using the pointer rCellVecFlat helps the compiler to assume continuous memory access.
```

terminationshock (approved):
As the neighbour list handling grows and becomes more elaborate, it could make sense to move all that even to a seperate module instead of `periodic`. Something like `neighbours`. This is not in scope of this merge request, though.

terminationshock:
> @terminationshock In order to prevent exposing the internal details of the neighbor list update, I have moved the actual `setNeighbourList()` into the same module, where `TNeighbourList` is defined. As far as I see, I did not break anything. Nevertheless, could you eventually check, whether the API works for your integration?

Let me check out the code and do a test...

aradi:
> As the neighbour list handling grows and becomes more elaborate, it could make sense to move all that even to a seperate module instead of `periodic`. Something like `neighbours`. This is not in scope of this merge request, though.

I fully agree. I indeed want to do it in a follow-up PR with some additional refactorings.

terminationshock:
@aradi My test was successful. The API works as expected.

aradi:
Cool, thx, I'll merge it then.

PR #1192, opened 2023-02-15
Fix testing

aradi:
Fixes two issues with testing:
- One example which can not run with 8 MPI procs was not disabled for this case,
- autotest2 output on failing test was confusing / wrong. The collected stderr and tagdiff files do not make much sense when testing with CTest, which runs autotest2 separately for each test case (and might execute several tests in parallel).


bhourahine (approved):
LGTM

PR #1193, opened 2023-02-15
Fix partial born and born derivatives generation

aradi:
Partial born and born derivatives calculations were never triggered due to wrong checking conditions.

I could not make a test case for this bug unfortunately, as currently we can not run complex scenarios invoking  both dftb+ and modes within one test case. :disappointed:


bhourahine (approved):
Seems ready to merge. I guess we need to talk about more general regression framework.

PR #1197, opened 2023-02-24
Fix transport io

aradi:


bhourahine (approved):
Would be nice if the test_open_files was somehow used in the CI in debug mode.

PR #1198, opened 2023-02-25
Refactoring of SCC loop

bhourahine:
Moving content into subroutines, re-using pontential update. Also some minor comment fixes for scc error variable.


vanderhe (approved):
Refactoring as is would vastly facilitate the CDFTB integration, therefore approved from my side.

PR #1202, opened 2023-03-01
Refactor and extend cubemanip

aradi:
* [x] Add `posx()`, `posy()` and `posz()` functions
* [x] Refactor cubemanip and make source PEP 8 compatible

### DIFF
--- a/tools/misc/cubemanip
+++ b/tools/misc/cubemanip
@@ -319,7 +319,7 @@ def _get_pos_funcs(origin, ngrids, gridvecs):
         return _calc_positions(1)
 
     def posz():
-        "Returns the z-position of every grid point"
+        "Returns the z-position of every grid point allowing for the use  of a coordinate origin"
         return _calc_positions(2)
 
     def _calc_positions(ind):
--- a/tools/misc/cubemanip
+++ b/tools/misc/cubemanip
@@ -311,7 +311,7 @@ def _get_pos_funcs(origin, ngrids, gridvecs):
     "Delivers three functions which return x, y & z position of the grid points"
 
     def posx():
-        "Returns the x-position of every grid point"
+        "Returns the x-position of every grid point, allowing for the use of a coordinate origin"
         return _calc_positions(0)
 
     def posy():
--- a/tools/misc/cubemanip
+++ b/tools/misc/cubemanip
@@ -39,7 +39,7 @@ data array is created by evaluating the expression. The number of cube files
 passed as arguments must be higher than the maximal reference index in the
 expression.
 
-Apart of the data arrays, following variables and functions can be used in the
+Apart from the data arrays, following variables and functions can be used in the
 expression:
 
   origin:  origin of the grid in the first input file as (3,) shaped array
--- a/tools/misc/cubemanip
+++ b/tools/misc/cubemanip
@@ -1,280 +1,365 @@
 #!/usr/bin/env python3
-#------------------------------------------------------------------------------#
-#  DFTB+: general package for performing fast atomistic simulations            #
-#  Copyright (C) 2006 - 2023  DFTB+ developers group                           #
-#                                                                              #
-#  See the LICENSE file for terms of usage and distribution.                   #
-#------------------------------------------------------------------------------#
+#--------------------------------------------------------------------------------------------------#
+#  DFTB+: general package for performing fast atomistic simulations                                #
+#  Copyright (C) 2006 - 2023  DFTB+ developers group                                               #
+#                                                                                                  #
+#  See the LICENSE file for terms of usage and distribution.                                       #
+#--------------------------------------------------------------------------------------------------#
 
-from numpy import oldnumeric as num
-from numpy import linalg as linal
-import math
-import sys
-import re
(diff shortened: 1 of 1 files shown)


tools/misc/cubemanip
```diff
@@ -1,280 +1,365 @@
...
+using the functions found in the numpy and numpy.linalg packages and results in
+a float or a float array. Unless the '-d' option is passed, the result must have
+the same shape, as the data arrays in the cube files. The references in the
+expression are substituted by the data arrays of the appropriate cube files. The
+resulting cube file will have the same header as the first input file, while its
+data array is created by evaluating the expression. The number of cube files
+passed as arguments must be higher than the maximal reference index in the
+expression.
+
+Apart of the data arrays, following variables and functions can be used in the
```
bhourahine, line 42:
```suggestion
Apart from the data arrays, following variables and functions can be used in the
```

tools/misc/cubemanip
```diff
@@ -1,280 +1,365 @@
...
+Examples:
+
+  Weighted sum of three cube files:
+  {SCRIPT_NAME} '0.25 * ($1) + 0.25 * ($2) + 0.5 * ($3)' f1.cube f2.cube f3.cube
+
+  Integral of the values in a cube file:
+  {SCRIPT_NAME} -d 'sum(($1))' file1.cube
+
+  Squared norm of the data in a cube file:
+  {SCRIPT_NAME} -d 'sum(($1)**2) * det(gridvecs)' file1.cube
```
bhourahine: It looks like this syntax might be able to handle circular polarisation (in addition to the linear case you've already demonstrated). Is that correct?

tools/misc/cubemanip
```diff
@@ -1,280 +1,365 @@
...
+import sys
+import numpy as np
+import numpy.linalg as la
 
 SCRIPT_NAME = "cubemanip"
 
-# Tolerance for positions to be treated as equal
-tolerance = 1.0e-5
+# tolerance for positions to be treated as equal
+TOLERANCE = 1.0e-5
```
bhourahine, line 21: Comment probably outside scope of this PR: Any reason for the specific value, and why not a user supplied argument with a default?

tools/misc/cubemanip
```diff
@@ -1,280 +1,365 @@
...
+        if self.ngrids.shape != (3,):
+            raise ValueError(_invalid_shape_msg("ngrids", (3,), self.ngrids.shape))
+
+        if self.data.shape != tuple(self.ngrids):
+            raise ValueError(_invalid_shape_msg("data", self.ngrids, self.data.shape))
+
+
+    @classmethod
+    def from_file(cls, fname):
+        """Creates a Cube instance from a file.
```
bhourahine, line 114:
```suggestion
        """Creates a cube instance from a file.
```
Or is this a proper name?

tools/misc/cubemanip
```diff
@@ -1,280 +1,365 @@
...
 
 
-class cube:
-    "Empty type to carry information about cube headers"
-    pass
+def _get_pos_funcs(origin, ngrids, gridvecs):
+    "Delivers three functions which return x, y & z position of the grid points"
 
+    def posx():
+        "Returns the x-position of every grid point"
```
bhourahine, line 314:
```suggestion
        "Returns the x-position of every grid point, allowing for the use of a coordinate origin"
```

tools/misc/cubemanip
```diff
@@ -1,280 +1,365 @@
...
+        return _calc_positions(0)
 
+    def posy():
+        "Returns the y-position of every grid point"
```
bhourahine, line 318:
```suggestion
        "Returns the y-position of every grid point, allowing for the the use of a coordinate origin"
```

tools/misc/cubemanip
```diff
@@ -1,280 +1,365 @@
...
+        return _calc_positions(1)
 
-def getCubeRepr(fname):
-    """Reads in a cube file with a given name"""
+    def posz():
+        "Returns the z-position of every grid point"
```
bhourahine, line 322:
```suggestion
        "Returns the z-position of every grid point allowing for the use  of a coordinate origin"
```

bhourahine (approved):
There is probably an argument for documenting does and don'ts, particularly for periodic cases if the coordinates are used in a supplied formula. (i.e. <i|x|j> doesn't mean much in a supercell). In the longer term, if we can get a consistent phase definition set up, something similar for k could also be done as for x.

tools/misc/cubemanip
aradi: It is the name of the class, which is capitalized. :wink:

tools/misc/cubemanip
aradi: Sure, we could do of course. The usual case is, however, that you combine cube-files from the same waveplot calculations, where the tolerance above should be OK-ish.

aradi:
You comment above the origin is true. However, note, that the cube file is not aware of periodicity, so that would be probably out of the scope of this simple script. (What we could do, though, is to allow for a user specified origin...)

PR #1203, opened 2023-03-01
Fix backward compatibility of argparse

jubich:
This should fix the backward compatibility problems when switching from optparse to argparse (#1081).

I would suggest inserting a test in `calc_timeprop_maxpoldir` so that at least two parameters (either -d and -w or -d and -e) are set before the files (mux.dat, muy.dat and muz.dat) are imported.

### DIFF
--- a/tools/misc/calc_timeprop_maxpoldir
+++ b/tools/misc/calc_timeprop_maxpoldir
@@ -16,30 +16,29 @@ from scipy import linalg, constants
 import sys
 hplanck = constants.physical_constants['Planck constant in eV s'][0] * 1.0E15
 
-USAGE = """Reads output from TD calculation (after kick) and excitation energy
+DESCRIPTION = """Reads output from TD calculation (after kick) and excitation energy
 (or wavelength) of interest and diagonalizes polarizability tensor, to
 calculate direction of maximum polarizability of that particular excitation.
-
 Needs mux.dat, muy.dat and muz.dat files in working directory."""
 
 def main():
-    parser = argparse.ArgumentParser(description=USAGE)
-    parser.add_argument("tau", metavar="DAMPING",
-                        help="damping constant in fs")
+    parser = argparse.ArgumentParser(description=DESCRIPTION)
+    parser.add_argument("-d", "--damping", dest="tau", required=True,
+                        type=float, help="damping constant in fs")
 
     group = parser.add_mutually_exclusive_group(required=True)
-    group.add_argument("-e", "--energy", dest="energy",
-                        help="excitation energy in eV")
-    group.add_argument("-w", "--wavelength", dest="wvlength",
-                        help="wavelength of excitation in nm")
+    group.add_argument("-e", "--energy", dest="energy", type=float,
+                       help="excitation energy in eV")
+    group.add_argument("-w", "--wavelength", dest="wvlength", type=float,
+                       help="wavelength of excitation in nm")
 
     args = parser.parse_args()
 
     if args.energy is not None:
-        energy = float(args.energy)
+        energy = args.energy
     elif args.wvlength is not None:
         cspeednm = constants.speed_of_light * 1.0e9 / 1.0e15
-        energy = hplanck * cspeednm / float(args.wvlength)
+        energy = hplanck * cspeednm / args.wvlength
 
     mux = np.loadtxt('mux.dat')
     muy = np.loadtxt('muy.dat')
@@ -87,7 +86,7 @@ def main():
         mu[2,1] += factor * muz[:,5]
         mu[2,2] += factor * muz[:,6]
 
-    damp = np.exp(-time/float(args.tau))
+    damp = np.exp(-time/ args.tau)
     energsev = np.fft.rfftfreq(10*time.shape[0],time[1]-time[0]) * hplanck
     idx = (np.abs(energsev-energy)).argmin()
 
--- a/tools/misc/calc_timeprop_spectrum
+++ b/tools/misc/calc_timeprop_spectrum
@@ -15,15 +15,15 @@ import argparse
 import numpy as np
 from scipy import constants
 
-USAGE = """Reads output from TD calculation (after kick) and produces
-absorption spectra. Needs mux.dat, muy.dat and muz.dat files in working directory."""
+DESCRIPTION = """Reads output from TD calculation (after kick) and produces absorption spectra.
+Needs mux.dat, muy.dat and muz.dat files in working directory."""
 
 def main():
-    parser = argparse.ArgumentParser(description=USAGE)
-    parser.add_argument("tau", metavar="DAMPING",
-                        help="damping constant in fs")
-    parser.add_argument("field", metavar="FIELD",
-                        help="field intensity in volts/angstrom")
+    parser = argparse.ArgumentParser(description=DESCRIPTION)
+    parser.add_argument("-d", "--damping", dest="tau", required=True,
+                        type=float, help="damping constant in fs")
+    parser.add_argument("-f", "--field", dest="field", required=True,
+                        type=float, help="field intensity in volts/angstrom")
 
     args = parser.parse_args()
 
@@ -38,7 +38,7 @@ def main():
     else:
         spinpol = False
 
-    damp = np.exp(-mux[:,0]/float(args.tau))
+    damp = np.exp(-mux[:,0]/ args.tau)
     if not spinpol:
         average = ((mux[:,1]-mux[0,1]) + (muy[:,2]-muy[0,2]) + (muz[:,3]-muz[0,3]))/3.0
     elif (spintype == 's'):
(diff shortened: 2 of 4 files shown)


vanderhe:
What about resorting to `required=True` for flags that are actually non-optional to omit some of the manual checks? Of course in case of mutually exclusive flags (like excitation energy/wavelength in [calc_timeprop_maxpoldir](https://github.com/dftbplus/dftbplus/pull/1203/files#diff-4adb58d108684cc44b9bc0bb2afa7a96c6e873b1ce733f5a90e7a406f1b14742)) we cannot indicate both as being required.

tools/misc/calc_timeprop_spectrum
```diff
@@ -15,17 +15,24 @@ import argparse
...
-    parser.add_argument("tau", metavar="DAMPING",
+    parser = argparse.ArgumentParser(usage=USAGE)
+    parser.add_argument("-d", "--damping", dest="tau",
                         help="damping constant in fs")
-    parser.add_argument("field", metavar="FIELD",
+    parser.add_argument("-f", "--field", dest="field",
                         help="field intensity in volts/angstrom")
 
-    args = parser.parse_args()
+    options, args = parser.parse_known_args()
```
aradi, line 31: Why not using `parser_args()` instead? It would automatically take care to check, that no positional arguments are provided.

tools/misc/calc_timeprop_spectrum
```diff
@@ -15,17 +15,24 @@ import argparse
...
+
+    if len(args) != 0:
+        parser.error("You must specify exactly two argument, the damping constant in fs and "
+                     "the field intensity in volts/angstrom")
```
aradi, line 35: Should be superfluous, if `parse_args()` is used above.

tools/misc/calc_timeprop_spectrum
```diff
@@ -15,17 +15,24 @@ import argparse
...
 
 def main():
-    parser = argparse.ArgumentParser(description=USAGE)
-    parser.add_argument("tau", metavar="DAMPING",
+    parser = argparse.ArgumentParser(usage=USAGE)
+    parser.add_argument("-d", "--damping", dest="tau",
                         help="damping constant in fs")
-    parser.add_argument("field", metavar="FIELD",
+    parser.add_argument("-f", "--field", dest="field",
                         help="field intensity in volts/angstrom")
```
aradi, line 29: Please specify argument type (`float`) here. Then you don't need the conversion to floats further down.

tools/misc/calc_timeprop_spectrum
```diff
@@ -38,7 +45,7 @@ def main():
     else:
         spinpol = False
 
-    damp = np.exp(-mux[:,0]/float(args.tau))
+    damp = np.exp(-mux[:,0]/float(options.tau))
```
aradi, line 48:
```suggestion
    damp = np.exp(-mux[:,0]/ options.tau)
```

tools/misc/calc_timeprop_spectrum
```diff
@@ -54,7 +61,7 @@ def main():
     cspeednm = constants.speed_of_light * 1.0e9 / 1.0e15
     energsev = np.fft.rfftfreq(10*mux.shape[0], mux[1,0]-mux[0,0]) * hplanck
     frec = np.fft.rfftfreq(10*mux.shape[0], (mux[1,0] - mux[0,0]) * 1.0E-15 )
-    absorption = -2.0 * energsev * spec.imag / np.pi / float(args.field)
+    absorption = -2.0 * energsev * spec.imag / np.pi / float(options.field)
```
aradi, line 64:
```suggestion
    absorption = -2.0 * energsev * spec.imag / np.pi / options.field
```

tools/misc/calc_timeprop_maxpoldir
```diff
@@ -16,30 +16,37 @@ from scipy import linalg, constants
...
+    parser.add_argument("-d", "--damping", dest="tau",
                         help="damping constant in fs")
-
-    group = parser.add_mutually_exclusive_group(required=True)
-    group.add_argument("-e", "--energy", dest="energy",
+    parser.add_argument("-e", "--energy", dest="energy",
                         help="excitation energy in eV")
-    group.add_argument("-w", "--wavelength", dest="wvlength",
+    parser.add_argument("-w", "--wavelength", dest="wvlength",
                         help="wavelength of excitation in nm")
```
aradi, line 34: Specifiy argument type to float to avoid manual conversion later (and to generate useful error messages, if something else was provided)

tools/misc/calc_timeprop_spectrum
```diff
@@ -15,17 +15,24 @@ import argparse
 import numpy as np
 from scipy import constants
 
-USAGE = """Reads output from TD calculation (after kick) and produces
-absorption spectra. Needs mux.dat, muy.dat and muz.dat files in working directory."""
+USAGE = """%(prog)s -d xx
```
aradi, line 18: That's not a very useful "usage" message. Use the one here, which one would get, if argumentparser had generated the message automatically (when `usage=USAGE` is used)

aradi (changes requested):
Thanks a lot for the quick fix. I have a few requests which should be considered.

tools/misc/calc_timeprop_spectrum
jubich: I changed it to `parse_args()`. My idea was simply to keep the message.

jubich:
> What about resorting to `required=True` for flags that are actually non-optional to omit some of the manual checks? Of course in case of mutually exclusive flags (like excitation energy/wavelength in [calc_timeprop_maxpoldir](https://github.com/dftbplus/dftbplus/pull/1203/files#diff-4adb58d108684cc44b9bc0bb2afa7a96c6e873b1ce733f5a90e7a406f1b14742)) we cannot indicate both as being required.

I added the `required=True` flags in `convolve.py`, `calc_timeprop_spectrum` and `calc_timeprop_maxpoldir` as suggested. As for the mutually exclusive arguments, I reintroduced `parser.add_mutually_exclusive_group(required=True)` which does just that. 

The reason for the required corrections was simply a misunderstanding on my part. I thought by
> I know it's not your code-design, but since the damping constant and field intensity are required in any case, why make them optional in the first place?

 in the original PR you meant switching to positional arguments because I didn't think of the `required=True` flag.

tools/misc/calc_timeprop_maxpoldir
```diff
@@ -16,30 +16,32 @@ from scipy import linalg, constants
...
 hplanck = constants.physical_constants['Planck constant in eV s'][0] * 1.0E15
 
-USAGE = """Reads output from TD calculation (after kick) and excitation energy
+USAGE = """%(prog)s [options]
+
+Reads output from TD calculation (after kick) and excitation energy
 (or wavelength) of interest and diagonalizes polarizability tensor, to
 calculate direction of maximum polarizability of that particular excitation.
 
 Needs mux.dat, muy.dat and muz.dat files in working directory."""
```
aradi, line 25:
```suggestion
DESCRIPTION = """Reads output from TD calculation (after kick) and excitation energy
(or wavelength) of interest and diagonalizes polarizability tensor, to
calculate direction of maximum polarizability of that particular excitation.
Needs mux.dat, muy.dat and muz.dat files in working directory."""
```

tools/misc/calc_timeprop_maxpoldir
```diff
@@ -16,30 +16,32 @@ from scipy import linalg, constants
...
 
 def main():
-    parser = argparse.ArgumentParser(description=USAGE)
-    parser.add_argument("tau", metavar="DAMPING",
-                        help="damping constant in fs")
+    parser = argparse.ArgumentParser(usage=USAGE)
```
aradi, line 28:
```suggestion
    parser = argparse.ArgumentParser(description=DESCRIPTION)
```

tools/misc/calc_timeprop_spectrum
```diff
@@ -15,15 +15,16 @@ import argparse
 import numpy as np
 from scipy import constants
 
-USAGE = """Reads output from TD calculation (after kick) and produces
-absorption spectra. Needs mux.dat, muy.dat and muz.dat files in working directory."""
+USAGE = """Reads output from TD calculation (after kick) and produces absorption spectra.
+
+Needs mux.dat, muy.dat and muz.dat files in working directory."""
```
aradi, line 20:
```suggestion
DESCRIPTION = """Reads output from TD calculation (after kick) and produces absorption spectra. Needs mux.dat, muy.dat and muz.dat files in working directory."""
```

tools/misc/calc_timeprop_spectrum
```diff
@@ -15,15 +15,16 @@ import argparse
...
 
 def main():
-    parser = argparse.ArgumentParser(description=USAGE)
-    parser.add_argument("tau", metavar="DAMPING",
-                        help="damping constant in fs")
-    parser.add_argument("field", metavar="FIELD",
-                        help="field intensity in volts/angstrom")
+    parser = argparse.ArgumentParser(usage=USAGE)
```
aradi, line 23:
```suggestion
    parser = argparse.ArgumentParser(description=DESCRIPTION)
```

aradi (approved):
I think, that's OK now. Please use `description=` instead of `usage=` (as suggested below), so that the generated help messages are more meaningful.

aradi:
I've merged it manually as squash-merge would have introduced some weird email address in the commit log (`88139524+jubich@users.noreply.github.com`), not sure why...

PR #1206, opened 2023-03-09
Minor typo in routine comment

bhourahine:
Preprocessor macro applied for correctly label precision in a loop


vanderhe (approved):
Works.

PR #1207, opened 2023-03-09
Fix indentation errors and convert comments to uppercase

vanderhe:
Had some idle time and wrote a script that fixes some style related things. While white spaces for visual alignment are discouraged in the style guide, having FORD comments start with upper case characters is more of an undocumented convention I guess (and not even close to being consistently implemented).

The script is not yet robust enough to apply it to the whole code base, but I might work on improving it so that it could potentially end up supplementing the tools of srcmanip at some point.


bhourahine:
Interesting question as to whether there is a convention about caps in variable comments... You might want to check the script in, or otherwise make available.

bhourahine (approved):
Ready to merge

PR #1212, opened 2023-03-24
Missing left side array dimensioning in k-pts

bhourahine:
Mostly the helical cases, where (:) = are missing.


vanderhe (approved):
Looks good. I think if we would systematically search for these cases we would find a whole bunch of them all over the codebase.

PR #1213, opened 2023-03-28
Nacv fixed numerics ii

thomas-niehaus:
I know its stupid to create another pull request for the same code change, but I had massive problems with the submodules. I commited the 'modified' submodules before issuing ' git submodule update --recursive' With git revert I could not come back to a version without errors. So here we are. 

The old PR  'Nacv fixed numerics'  contains the history of Balints code change requests. This PR has those changes (or most of them).


aradi:
OK, I merged it, but then reverted the merge again due to the issue with the fake GitHub email address. 

@thomas-niehaus Could you please create the same PR again after you made your email address public on GitHub (in case you are willing to do so)? Then the single commit, resulting of the squash-merge I intend to do, would be associated with your proper name and email address (hopefully)...

PR #1214, opened 2023-03-29
Fixed commensurate k-test for helical

bhourahine:
And added high order rotation regression case


src/dftbp/dftbplus/initprogram.F90
```diff
@@ -1403,7 +1404,7 @@ subroutine initProgramVariables(this, input, env)
       this%kWeight(:) = input%ctrl%kWeight / sum(input%ctrl%kWeight)
       if (this%tHelical) then
         if (any(abs(this%kPoint(2,:)*nint(this%latVec(3,1))&
-            & -nint(this%kPoint(2,:)*nint(this%latVec(3,1)))) > epsilon(1.0_dp))) then
+            & -nint(this%kPoint(2,:)*nint(this%latVec(3,1)))) > epsilon(1.0_rsp))) then
```
aradi, line 1407: Wouldn't it be more meaningful not to tie it to `epsilon` but to an arbitrary tolerance value we are willing to accept (e.g. `1e-6_dp` or so)?

PR #1215, opened 2023-03-30
Temporary: Turn some of the T-formatting into functions

aradi:


bhourahine:
@aradi Do you want to add this to the 23.1 release?

PR #1216, opened 2023-03-30
Remove unsupported function in dftbplus.h file

jubich:


jubich:
Sure, that should be changed now.

PR #1219, opened 2023-04-08
API update of q0

bhourahine:
This PR enable updating for reference occupations of atoms for the non-shell-resolved DFTB model.

- [X] q0 get and set from Fortran API (atom resolved)
- [X] Recalculate on q0 update
- [X] Serial tester code
- [x] MPI tester
- [x] C bindings
- [x] Test cases


terminationshock:
So far, the layout of the API looks reasonable to me.

bhourahine:
@terminationshock We should probably discuss at the dev meeting:
1.  This updates the orbital reference, but there is an argument for instead working with the shell average reference.
2. The C API will be a bit messy as q0 is a multi-dimensional array, so extra calls for the dimensioning will be needed (and should it be passed as a 1D array or some attempt as a 2/3D array?).
3. In the case of original non-shell resolved DFTB, only a single number matters (effective Z), can we have some simpler way to deal with this special case?
Finally, for purposes like MD, we need to discuss driver strategies.

bhourahine:
Now non-shell resolved model calls

bhourahine:
@terminationshock should be done now, just on test. The tester in test/src/dftbp/api/mm/testers/test_mpic.c shows the layout of a C driven MPI calculation.

src/dftbp/api/mm/dftbplus.h
```diff
@@ -386,6 +406,48 @@ void dftbp_get_gross_charges(DftbPlus *instance, double *charges);
...
+
+
+/**
+ * Sets the net charge of the QM system.
+ *
+ * \param[inout] instance Handler of the DFTB+ instance.
+ *
+ * \param[in] charge of the system (+ve value being electron deficient)
+ */
+void dftbp_set_net_charge(DftbPlus *instance, double *charge);
```
terminationshock, line 436: What is the benefit in passing a pointer to the charge instead of the charge itself?

src/dftbp/api/mm/dftbplus.h
```diff
@@ -386,6 +406,48 @@ void dftbp_get_gross_charges(DftbPlus *instance, double *charges);
...
+/**
+ * Sets the net charge and spin of the QM system.
+ *
+ * \param[inout] instance Handler of the DFTB+ instance.
+ *
+ * \param[in] charge net charge of the system (+ve value being electron deficient)
+ *
+ * \param[in] spin unpaired spins of the system (+ve value being electron deficient)
+ */
+void dftbp_set_net_chargespin(DftbPlus *instance, double *charge, double *spin);
```
terminationshock, line 448: Same question here for `charge` and `spin`.

src/dftbp/api/mm/mmapi.F90
```diff
@@ -634,6 +648,101 @@ subroutine TDftbPlus_getCM5Charges(this, atomCharges)
...
+
+    call this%checkInit()
+
+    if (this%main%uniqHubbU%mHubbU > 1) then
+      call error("Reference charge call unsupported for shell resolved models")
+    end if
+    mOrb = this%main%orb%mOrb
+    nAtom = nrOfAtoms(this%main)
+    nSpin = this%main%nSpin
+    allocate(q0(mOrb, nAtom, nspin), source=0.0_dp)
```
terminationshock, line 671: Is it really necessary to initialize with zeros here?

test/src/dftbp/api/mm/testers/test_mpic.c
```diff
@@ -0,0 +1,135 @@
+/*------------------------------------------------------------------------------------------------*/
+/*  DFTB+: general package for performing fast atomistic simulations                              */
+/*  Copyright (C) 2006 - 2023  DFTB+ developers group                                             */
+/*                                                                                                */
+/*  See the LICENSE file for terms of usage and distribution.                                     */
+/*------------------------------------------------------------------------------------------------*/
+
+/**
+ * MPI parallel calculation via the API and checks the result
```
terminationshock, line 9: Thank you for this example!

src/dftbp/api/mm/mmapi.F90
bhourahine: Not really, but again the pattern is becoming common. The call below is intent out, so it's definitely irrelevant (I'll remove)
```suggestion
    allocate(q0(mOrb, nAtom, nspin))
```

test/src/dftbp/api/mm/testers/test_mpic.c
bhourahine: It doesn't do anything with coordinates, but the call prototype is in dftbplus.h

bhourahine:
@terminationshock changes from the comments now made (and since the net charge part of the API is removed, this also deals with the pointer comment).

src/dftbp/dftbplus/initprogram.F90
```diff
@@ -4150,16 +4150,30 @@ end subroutine initializeCharges
...
+
+    !> Reference 'neutral' charges for each atom orbital
+    real(dp), allocatable, intent(out) :: q0(:,:,:)
+
+    !> Shell resolved neutral reference
+    real(dp), allocatable, intent(out) :: qShell0(:,:)
+
     type(TWrappedInt1), optional, intent(in) :: customOccAtoms(:)
+
     real(dp), optional, intent(in) :: customOccFillings(:,:)
```
terminationshock, line 4174: It is probably worth documenting these optional arguments as well.

test/src/dftbp/api/mm/testers/test_intcharges.f90
```diff
@@ -0,0 +1,173 @@
...
+  real(dp) :: atomCharges(nAtom), cm5Charges(nAtom), atomMasses(nAtom)
+  real(dp) :: q0(nAtom)
+  type(fnode), pointer :: pRoot, pGeo, pHam, pDftb, pMaxAng, pSlakos, pAnalysis, pCm5
+  type(fnode), pointer :: pParserOpts
+
+  character(:), allocatable :: DftbVersion
+  integer :: major, minor, patch
+  integer :: ii
+
+  !integer :: devNull
```
terminationshock, line 55: Here are some commented-out lines in this file, which should be removed from my perspective.

test/src/dftbp/api/mm/testers/test_intcharges.f90
```diff
@@ -0,0 +1,173 @@
...
+
+  call getDftbPlusBuild(DftbVersion)
+  write(*,*)'DFTB+ build: ' // "'" // trim(DftbVersion) // "'"
+  call getDftbPlusApi(major, minor, patch)
+  write(*,"(1X,A,1X,I0,'.',I0,'.',I0)")'API version:', major, minor, patch
+
+  ! Note: setting the global standard output to /dev/null will also suppress run-time error messages
+  !open(newunit=devNull, file="/dev/null", action="write")
+  !call TDftbPlus_init(dftbp, outputUnit=devNull)
```
terminationshock, line 64: see other comment

terminationshock (commented):
Thanks again for these changes. Just some minor comments from my side.

src/dftbp/dftbplus/initprogram.F90
bhourahine: Done

test/src/dftbp/api/mm/testers/test_intcharges.f90
bhourahine: Done,  but were presumably originally intended as hints for implementers (they are preserved in test_extcharges.f90)

test/src/dftbp/api/mm/testers/test_intcharges.f90
bhourahine: Done

PR #1222, opened 2023-04-27
Steepest descent optimizer for new interface

vanderhe:
Introduces the steepest descent optimizer as fallback option, integrated into the new `GeometryOptimization{}` interface.


src/dftbp/geoopt/steepdesc.F90
```diff
@@ -7,199 +7,107 @@
...
-    !> minimizer object
-    type(TSteepDesc), intent(inout) :: this
+    @:ASSERT(nVar > 0)
+    @:ASSERT(size(mass) * 3 == nVar)
 
-    !> Point to start from
-    real(dp), intent(in) :: x0(:)
+    this%nVar = nVar
+    allocate(this%weight(nVar))
+    this%weight(:) = 0.5_dp * input%stepSize**2 / reshape(spread(mass, 1, 3), [nVar])
```
bhourahine, line 71: Any reason not to make the StepSize**2 / mass the optimizer argument? It's giving a step distance as length/force, but actually what probably beter is a quantity dimensioned as length/(dE/dx). There is also a question as to why scaling the coordinate system by inverse mass would be a better choice than mass or no scaling at all (the inverse mass means H atoms move further than heavier atoms in each step).

bhourahine (commented):
Some things to discuss...

src/dftbp/geoopt/steepdesc.F90
bhourahine: Possibly the simplest thing would be to set the compliance (i.e. inverse of stiffness) for the driver (i.e. proportionality between displacement and force). This quantity is dimensioned as length / force, so would  not be in mass weighted coordinates.

test/app/dftb+/geoopt/H2O_sd_old/dftb_in.hsd
```diff
@@ -0,0 +1,50 @@
...
+ O H
+     1    1    0.00000000000E+00  -0.10000000000E+01   0.00000000000E+00
+     2    2    0.00000000000E+00   0.00000000000E+00   0.78306400000E+00
+     3    2    0.00000000000E+00   0.00000000000E+00  -0.78306400000E+00
+}
+
+Driver = SteepestDescent {
+  StepSize = 100.0
+  MaxForceComponent = 1.0E-08
+  MaxSteps = 100
```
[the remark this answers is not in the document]
vanderhe: On my machine it needs 57 geometry steps to converge. Shall I loosen the convergence criterion then?

doc/dftb+/manual/dftbp.tex
```diff
@@ -458,6 +458,31 @@ \subsubsection{Optimiser: FIRE}
...
+  along the forces. The displacement $\delta x_i$ along the
+  $i^\mathrm{th}$ coordinate is given for each atom as $\delta x_i =
+  \frac{f_i}{2m}\delta t^2$, where $f_i$ is the appropriate force
+  component and $m$ is the mass of the atom.
+\end{description}
+
+\begin{verbatim}
+Driver = GeometryOptimization {
+  Optimizer = SteepestDescent {
+    StepSize = 100.0
```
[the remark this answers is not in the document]
vanderhe: Atomic time units I guess. The StepSize parameter is identical to the old SteepestDescent{} driver.

test/app/dftb+/geoopt/H2O_sd_old/dftb_in.hsd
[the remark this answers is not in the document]
bhourahine: It doesn't have to converge, only write a tag file which can be tested. Provided the quantities are robust to ~1e-6 between different machines (unless this is over-ridden) you could even just set the max number of iterations to something low.

src/dftbp/dftbplus/input/geoopt.F90
```diff
@@ -184,6 +189,24 @@ subroutine readOptTolerance(node, input)
...
+    type(fnode), pointer, intent(in) :: node
+
+    !> Control structure to be filled
+    type(TSteepDescInput), intent(out) :: input
+
+    type(fnode), pointer :: field
+    type(string) :: modifier
+
+    call getChildValue(node, "StepSize", input%stepSize, 100.0_dp, modifier=modifier, child=field)
+    call convertUnitHsd(char(modifier), timeUnits, field, input%stepSize)
```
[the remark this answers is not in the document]
vanderhe: The only reason I kept it is the old driver. We should discuss if we want to keep this StepSize or switch to something less cumbersome, especially since calculating the weights requires the masses and messes up the function signature(s).

test/app/dftb+/geoopt/H2O_sd_old/dftb_in.hsd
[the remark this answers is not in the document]
vanderhe: I see, I will adapt it.

src/dftbp/dftbplus/input/geoopt.F90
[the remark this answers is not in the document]
bhourahine: One for the next dev meeting. My quick test suggested the unweighted coordinates required fewer steps for H2O2, but the weighted ones were better for H2O.

src/dftbp/geoopt/steepdesc.F90
vanderhe: This would effectively be a single scalar, scaling the gradients right? Or a 1d array of weights?

test/app/dftb+/geoopt/H2O_sd_old/dftb_in.hsd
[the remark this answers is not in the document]
vanderhe: Number of steps reduced to ~20.

src/dftbp/geoopt/steepdesc.F90
vanderhe: Switched to single scaling factor.

vanderhe:
Rebased on main...

doc/dftb+/manual/dftbp.tex
```diff
@@ -458,6 +458,27 @@ \subsubsection{Optimiser: FIRE}
...
+  \kw{Scaling} & r &  & 1.0 & \\
+\end{ptable}
+\begin{description}
+\item[\is{Scaling}] Scaling factor of gradients.
+\end{description}
+
+\begin{verbatim}
+Driver = GeometryOptimization {
+  Optimizer = SteepestDescent {
+    Scaling = 1.0
```
bhourahine, line 476: Possible 'ScalingFactor' is clearer, or the old one uses 'StepSize'. Which of the three do you think is better?

bhourahine (commented):
Generally OK, but a couple of comments to discuss.

doc/dftb+/manual/dftbp.tex
vanderhe: From those three options I would probably choose ''ScalingFactor". I will adapt it.

doc/dftb+/manual/dftbp.tex
```diff
@@ -458,6 +458,27 @@ \subsubsection{Optimiser: FIRE}
 \end{verbatim}
 
 
+\subsubsection{Optimiser: SteepestDescent}
+\label{sec:dftbp.Optimiser.SteepestDescent}
+
+Steepest descent optimiser, that proposes displacements according to the (scaled) gradients.
```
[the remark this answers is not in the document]
vanderhe: Fixed

doc/dftb+/manual/dftbp.tex
```diff
@@ -458,6 +458,27 @@ \subsubsection{Optimiser: FIRE}
...
+
+\begin{ptable}
+  \kw{Scaling} & r &  & 1.0 & \\
+\end{ptable}
+\begin{description}
+\item[\is{Scaling}] Scaling factor of gradients.
```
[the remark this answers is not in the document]
vanderhe: Fixed

doc/dftb+/manual/dftbp.tex
vanderhe: Done

doc/dftb+/manual/dftbp.tex
```diff
@@ -458,6 +458,28 @@ \subsubsection{Optimiser: FIRE}
 \end{verbatim}
 
 
+\subsubsection{Optimiser: SteepestDescent}
+\label{sec:dftbp.Optimiser.SteepestDescent}
+
+Steepest descent optimiser, that proposes displacements according to the (scaled) gradient value.
+
+\begin{ptable}
+  \kw{ScalingFactor} & r &  & 1.0 & \\
```
bhourahine, line 467: Coded as a different value
```suggestion
  \kw{ScalingFactor} & r &  & 0.1 & \\
```

vanderhe: Yes, sorry for that. According to my testing (geoopt) a value of 1.0 performs quite well, so I would suggest to consistently switch it to 1.0 then. Let me fix that real quick.

vanderhe: Should be consistently set to 1.0 now.

bhourahine (approved):
LGTM

PR #1225, opened 2023-05-03
Density Matrix Construction with GPU MAGMA

robagl9695:
Added the construction of the Density Matrix on the GPU employing MAGMA BLAS subroutines for complex and real cases.

### DIFF
--- a/src/dftbp/dftb/densitymatrix.F90
+++ b/src/dftbp/dftb/densitymatrix.F90
@@ -464,7 +464,6 @@ contains
   !> Make a regular density matrix for the real wave-function case in the GPU
   !> Note: In order to save memory, the eigenvectors (which should be intent(in) parameters) are
   !> overwritten and then restored again
-
   subroutine makeDensityMtxRealGPU(dm, eigenvecs, filling)
 
     !> the resulting nOrb*nOrb density matrix
@@ -593,10 +592,10 @@ contains
 
   end subroutine makeDensityMtxRealGPU
 
+
   !> Make a regular density matrix for the complex wave-function case in the GPU
   !> Note: In order to save memory, the eigenvectors (which should be intent(in) parameters) are
   !> overwritten and then restored again
-
   subroutine makeDensityMtxCmplxGPU(dm, eigenvecs, filling)
 
     !> the resulting nOrb*nOrb density matrix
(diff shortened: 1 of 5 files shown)


bhourahine (commented):
Could you add an entry to the CHANGELOG file, plus any appropriate edit to the AUTHORS.RST file.
Let me know if the kind(Int64) suggestions from @aradi  works, as this would make the changes a little simpler (and should work for any standard compliant compiler, instead of needing customisation for each one separately).

src/dftbp/dftb/densitymatrix.F90
```diff
@@ -446,4 +455,277 @@ end subroutine makeDensityMtxCplxBlacs
...
+    endif
+
+    call magmaf_queue_destroy( queue );
+
+  end subroutine makeDensityMtxRealGPU
+
+  !> Make a regular density matrix for the complex wave-function case in the GPU
+  !> Note: In order to save memory, the eigenvectors (which should be intent(in) parameters) are
+  !> overwritten and then restored again
+
```
bhourahine, line 599:
```suggestion
```

src/dftbp/dftb/densitymatrix.F90
```diff
@@ -446,4 +455,277 @@ end subroutine makeDensityMtxCplxBlacs
...
+#:if WITH_MAGMA
+!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+!!!!!  MAGMA GPU routines    !!!!!                                                   
+!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+  !> Make a regular density matrix for the real wave-function case in the GPU
+  !> Note: In order to save memory, the eigenvectors (which should be intent(in) parameters) are
+  !> overwritten and then restored again
+
```
bhourahine, line 467: Doxygen might have problems with the blank line between the comment and the associated routine
```suggestion
```

bhourahine (commented):
Just some trivial source formatting stuff

bhourahine:
I've just tested the regressions with intel 2021, and ctest -R GPU fails for the SiC_512_GPU case (but not the non-self-consistent one). I think this is an issue with the complex density matrix routine (as running it with a real matrix works, but making the problem slightly complex fails).

robagl9695:
Resubmitting pull request to include only real valued DM.

bhourahine:
Is the plan to open a new pull request? If so, could you squash the changes into one commit and also re-base on the main repository (git rebase in both cases).

robagl9695:
Hi Ben,

Yes, that's the idea. We have been trying to fix the convergence issues for
complex-valued DM.

However, real-valued DM cases work correctly. We plan to open a new pull
request for just the real-valued DM modification and then another pull
request with the complex-valued, once we fix it.

I will follow your instructions to do so.

Kind regards,

Roberto G.

PR #1200, opened 2023-02-27
Electronic Constraints

vanderhe:
Enables to impose constraints on the electronic ground state.

- [x] general CDFTB infrastructure
- [x] available constraints: atomic Mulliken populations
- [x] atomic gradients/stress tensor
- [x] re-check analytic against numerical gradients of artificial potential
- [x] documentation
- [x] possible barriers for unsupported extensions
- [x] wait for #1226
- [x] wait for #1222
- [x] regression tests
- [x] re-check atomic gradients/stress tensor

Caveats of current implementation:
- SCC-DFTB Hamiltonian re-build in every micro-iteration, although not necessary
- LC-DFTB Hamiltonian re-build in every micro-iteration, although not necessary
- allows for impossible targets, e.g. those that would violate the total number of electrons in the system

Adding constraints, e.g. non-colinear atomic shell spin, atomic shell orbital angular momentum, atomic shell projection of orbital angular momentum, atomic shell total angular momentum, atomic shell projection of angular momentum, ..., is straightforward.

PR #1226, opened 2023-05-04
Reset functionality for new optimizers

vanderhe:
Adds reset functionality to new optimizers, as required by #1200.

### DIFF
--- a/src/dftbp/geoopt/rationalfunc.F90
+++ b/src/dftbp/geoopt/rationalfunc.F90
@@ -5,6 +5,8 @@
 !  See the LICENSE file for terms of usage and distribution.                                       !
 !--------------------------------------------------------------------------------------------------!
 
+#:include 'common.fypp'
+
 !> Implementation of a rational function optimization procedure.
 !>
 !> The optimization problem is solved by determining the lowest eigensolution of the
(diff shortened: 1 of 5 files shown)


src/dftbp/geoopt/rationalfunc.F90
```diff
@@ -130,6 +135,30 @@ subroutine step(this, val, grad, displ)
...
 
+  !> Reset optimizer
+  subroutine reset(this)
+
+    !> Instance of geometry optimization driver
+    class(TRationalFunc), intent(inout) :: this
+
+    integer :: ii, nvar1, npvar, npvar1
+
+    nvar1 = this%nvar + 1
```
bhourahine, line 146: Could you add a comment as to why/what this is doing

bhourahine (approved):
Seems ready (see comment)

src/dftbp/geoopt/rationalfunc.F90
vanderhe: Better not, otherwise I would have to write something like "superfluous temporary integers". Jokes aside, I have removed the unnecessary lines, including the same copy&paste error in the step() function.

bhourahine:
I'll merge in a minute. At some point we'll need another PR to return errors upwards to main instead of calling error() internally.

PR #1234, opened 2023-05-18
Implement a generalized input and structure for constrained DFTB

aradi:
I think, with not too much effort, we could make the input and the features for constrained DFTB more flexible. Before making a coded up suggestion, I'd like to ask you about your opinion on it:

```
ElectronicConstraints {

  Optimiser = SteepestDescent {
    ScalingFactor = 0.25
  }
  ConstrTolerance = 1.0e-08
  MaxConstrIterations = 100
  ConvergentConstrOnly = Yes

  Constraints {

    # Setting the total population on a group of atoms
    TotalPopulation {
      Atoms = 1:10
      Population = 20  
    }

    # Setting individual populations on atoms (same number for all atoms)
    # Populations field contains only a single number
    AtomPopulations {
      Atoms = C
      Populations = 4
    }

    # Setting individual populations on atoms (same number for all atoms)
    # Populations field contains only a single number
    AtomPopulations {
      Atoms = C
      SpinChannels = 2
      # Specifying q/m values for each atom. (Or should we go for q_up/q_down in the colinear case?)
      Populations = 4.0 0.5
    }

    # Setting individual populations on atoms (different number for each atom)
    # Populations field contains as many numbers as atoms in the Atoms field.
    AtomPopulations {
      Atoms = 1:10
      Populations = 4.2 3.8 4.2 3.8 4.2 3.8 4.2 3.8 4.2 3.8
    }

    # Setting shell resolved population on atoms
    # (same shell population for each atom, negative number means that no constraint
    # is set for a given shell -- maybe that is somewhat confusing?)
    # Populations 
    ShellPopulations {
      Atoms = Si
      Populations = 2.0 -1.0 0.0
    }

    # Setting shell resolved population on atoms
    # (different shell populations for each atom, negative number means that no constraint
    # is set for a given shell -- maybe that is somewhat confusing?)
   # Populations field contains nShells x nAtoms fields.
    ShellPopulations {
      Atoms = 1:2
      Populations = 2.0 -1.0 2.0  1.8 -1.0 2.2
    }
  }
}
```
It would allow a more flexible individual setting for atoms, if needed (e.g. user gets individual atomic / shell-wise populations from some embedding model). If I understand correctly, this all can be implemented with the current constrained formalism easily.


aradi:
* Yes, I think, every constraint should its own block. If the given constraint requires a list of atoms to specify its target, it would have an `Atoms` option. If it needs any other specification (e.g. geometrical region), it should have an other option inside instead. So, the suggestion is to have the constraint type as outer block, and its target and further details as options within. (Instead of specifying the target region as outer block...)

* The `SpinChannels` is indeed somewhat tricky, but I could not come up with something better so far. And yes, we should make sure, that m can be specified without q. (Maybe we should allow for `NaN` in the input to signalize unspecified value? :wink: )

* As for angular momentum & co: 
   ```
   # Sets individual angular momenta for a list of atoms
   AngularMomentum {
     L2 = ...
     Sz = ...
     Atoms = ...
   }

   # Sets the total angular momentum for a list of atoms
   TotalAngularMomentum {
      Lz = ...
      Atoms = ...
   ```

* But, probably we could also condense things a bit with respect of total vs. individual, by allowing their specification in the same block
   ```
   Population {
      Atoms = 1:10
     # Individual settings, if needed
     Populations = 3.8 4.2 3.8 4.2 3.8 4.2 3.8 4.2 3.8 4.2 
     # Alternatively using charges
     Charges = -0.2 0.2 -0.2 0.2 -0.2 0.2 -0.2 0.2 -0.2 0.2 
     # Constraint on total population 
     TotalPopulation = 40
     # Alternatively with charges
     TotalCharges = 0.0
   }
   ```

bhourahine:
@aradi lots of conflicts for this branch, is it worth trying to keep this in sync for the moment?
SpinChannels = 2 still puzzles me.

Also the syntax should be naturally extensible, so any thoughts how to deal with non-Lagrangian based versions, i.e., things like the quadratic-type terms of the form v * (x - x_0)^2 where the v is user provided, not solved as a Lagrangian problem.

aradi:
Here a more condensed suggestion for the constraint syntax:
```
ElectronicConstraints {

  # Individual population on every C atom
  Population {
    Atoms = C
    Populations = 4
    # Also possible:
    # Charges = 0
  }

  # Total population of every C atom
  Population {
    Atoms = C
    TotalPopulation = 12
    # Also possible:
    # TotalCharge = 0
 }

  # Population on individual shells
  Population {
     Atoms = C
     Shells = 1:-1
     Populations = 2.0 2.0  2.0 2.0  2.0 2.0
     # Also possible:
    # Charges = 0.0 0.0  0.0 0.0  0.0 0.0
  }
}

bhourahine:
Definitely getting there. Instead of Populations, would IndividualPopulations (in contrast to TotalPopulation) be clearer? Also, for the shell populations is the same logic applying (i.e. TotalPopulation is all of the s shells for these atoms) ?

bhourahine:
@aradi Due to #1677 do you want to close this one?

aradi:
Sure, makes sense to close it.

PR #1235, opened 2023-05-18
NACV compatibility tests happening earlier

bhourahine:
Changes for the code to trap for cases that won't run before performing much work.

### DIFF
--- a/src/dftbp/timedep/linresp.F90
+++ b/src/dftbp/timedep/linresp.F90
@@ -122,7 +122,7 @@ contains
 
 
   !> Initialize an internal data type for linear response excitations
-  subroutine LinResp_init(this, ini, nAtom, nEl, onSiteMatrixElements)
+  subroutine LinResp_init(this, ini, nAtom, nEl, nSpin, onSiteMatrixElements)
 
     !> data structure for linear response
     type(TLinResp), intent(out) :: this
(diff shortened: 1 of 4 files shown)


src/dftbp/timedep/linresp.F90
thomas-niehaus, line 1: Sometimes the key StateCoupling is used, sometimes StateCouplings. Should be the same everywhere and identical to use in manual.

src/dftbp/timedep/linrespgrad.F90
thomas-niehaus, line 1: Well, Balint told me to suppress explicit array dimension and write
nacv = 0
instead of nacv(:,:,:) = 0. I find the second one more informative, but what is the current politics on this issue?

thomas-niehaus (commented):
Useful changes!

src/dftbp/timedep/linrespgrad.F90
bhourahine: I think we discuss at the next development meeting. The case without the right hand side dimensioning looks more sensible if it is an automatic allocation, but not for an already allocated variable.

aradi: As an example from the code above, I prefer to write
```
shiftPerAtom(:) = shiftPerAtom + shiftPerL(1,:)
```
Explicit index placeholders on the LHS (as no reallocation should occur), but on the RHS only for those terms, where an actual sub-selection should be made, but not, if the entire arrray is meant.

src/dftbp/timedep/linresp.F90
bhourahine: Now changed to be consistent.

PR #1253, opened 2023-06-15
Real-valued DM on GPU using MAGMA-BLAS

robagl9695:


bhourahine (commented):
Did you test the suggestion for using the appropriate int type instead of using the compiler option?

bhourahine:
Pull request #1289 contains these changes and is up to date with the main codebase.

PR #1255, opened 2023-06-18
Fixes and generalises header print

bhourahine:
The setupgeom code also uses this and was outdated. Mechanism ready for other codes (waveplot, modes) for a future PR.  Also moved type for release map used in parser calls to a global location, so can be used by other Fortran tools in the future.


vanderhe:
@bhourahine Will review tomorrow, I'm technically on vacation today.

vanderhe (approved):
Works as intended, as far as I can tell.

PR #1257, opened 2023-06-21
ChIMES repulsives for non-periodic systems

vanderhe:
Adds a workaround to enable ChIMES repulsives for non-periodic systems.

The molecule is placed in a large box, whereas the padding is specified from the users' perspective. If it were possible to query the largest occurring ChIMES cutoff via the interface, this parameter would not be necessary, but that does not seem to be implemented at the moment. In the future, it is of course preferable to directly support non-periodic systems in the ChIMES calculator and lsq fitting tool.

- [x] ChIMES repulsives for non-periodic systems
- [ ] documentation
- [ ] regression tests

@aradi @bhourahine Could you have a quick look at the changes and provide feedback if this workaround is acceptable? In that case I would proceed with documenting and testing it.


bhourahine:
@vanderhe did it turn out that the pair/triplet API call not solve this without padding?

vanderhe:
@bhourahine I did not yet find time to look into it again, will try to do so soon.

bhourahine:
Should use API for CHiMES directly for these geometries.

PR #1261, opened 2023-06-30
Fix manual links in README

### LINKED ISSUE #1258: fix manual pdf link in README.rst
The links to the manual in [README](https://github.com/dftbplus/dftbplus/blob/fc3770abefb4242caa152d621dea395ae85904d9/README.rst) (http://www.dftbplus.org/fileadmin/DFTBPLUS/public/dftbplus/latest/manual.pdf) returns [404 Not Found](https://web.archive.org/web/20230622161113/https://dftbplus.org/fileadmin/DFTBPLUS/public/dftbplus/latest/manual.pdf)!

vanderhe:
Fixes #1258.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -121,7 +121,7 @@ Consult following resources for documentation:
   <http://dftbplus-recipes.readthedocs.io/>`_
 
 * `Reference manual describing all features (DFTB+ Manual)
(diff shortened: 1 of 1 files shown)


bhourahine (approved):
Ready to merge. 

Very minor thing for the future would be to think about a way to pin the manual link to the last code release tag, instead of only the latest version. That way, anyone with an outdated source tree will get the relevant manual. I suspect the only way to do that would be via something like utils/srcmanip/set_source_headers

PR #1274, opened 2023-07-06
Td mpi

thomas-niehaus:
First attempt of an MPI version of Casida. Currently supports Arpack solver and can do spectra for singlets/triplets/temperature/windows/spin polarized/onsite corrected. Based on the MPI version (not BLACS) of arpack-ng. 
To compile, compile first arpack-ng (works out of the box) and then include the path to the lib via CMAKE_PREFIX_PATH.

- The Arpack section of the code is fully parallel, not only the matrix-vector product. 
- The code seems to scale for a couple of processors. Could not test it nProcs > 6, because our HPC center doesn't have the compilers dftb+ requires ;-) UPDATE: Code now compiles on the HPC with ifort 2021, but spin-polarized calculations crash. Issue of MPI communicators?
- Since I use the arpack MPI library, I make no use of the new BLACS excVectorGrid feature. Distribution is done in the routine localSizeCasidaVectors.
- The bare routine transq can no longer be called under MPI (it is simply too costly to evaluate it with global indices). Consequence: CachedCharges is mandatory for MPI. The routine calcTransitionDipoles was using transq, therefore had to call TTransCharges_init earlier, which makes a second call to TTransCharges_init necessary, in case windowing is used.
- Please have a look how code can be made more elegant, there are to my taste too many preprocessor directives. Example in linrespgrad.F90 line 639.  This is the first time I write MPI code so my approach might be cumbersome/laughable.
- I have started to parallelize rangeSep, which should pose no problems. It needs more memory though, need nAtoms x nOcc x nVir global array. Currently the ground state is not MPI parallel, so I will not continue right now. 
- Can you have a look at transcharges.F90 line 182?
- getAtomicGammaMatrixBlacs is maybe not what you want.
- local2GlobalBlacsArray should maybe go to a different module. 
- The tests contain some MPI cases

Let me know what you think.....


bhourahine:
Looks like the arpack-ng finder isn't working with a CMAKE_MODULE_PATH pointing to the location of arpackng-config.cmake. Probably should also cope with old style parpack install, as that's more likely to be in the usual deb/rpm installs.

bhourahine:
> But you can compile on dpdevel? Works for me using env_dftb+/gcc12-openmpi. I get yet another error (compared to my HPC center) with the NO example. Will look into this.

Exactly which cmake options are you using? As with that environment on buildbot, I get undefined references to pdsaupd_

thomas-niehaus:
Now I also have a test with more cpu. Looks reasonable, I think. 
[plot.pdf](https://github.com/dftbplus/dftbplus/files/14162648/plot.pdf)

bhourahine:
Hi Thomas,

Does

module add env/_whatever_compiler_

(i.e. the gnu/intel/whatever build environment)

not put this into the path?

Regards,

Ben

thomas-niehaus:
But you had some problems with missing symbols, didn't you? I don't see arpack in the module list, but remember vaguely that Balint had arpack installed.

src/dftbp/dftb/scc.F90
```diff
@@ -526,20 +542,64 @@ subroutine getAtomicGammaMatrix(this, gammamat, iNeighbour, img2CentCell)
...
+    integer, intent(in) :: img2CentCell(:)
+
+    integer :: ii, jj, iLoc, jLoc, rSrc, cSrc
 
+    @:ASSERT(this%tInitialised)
+    @:ASSERT(all(shape(gammamat) == [ this%nAtom, this%nAtom ]))
+    @:ASSERT(this%elstatType == elstatTypes%gammaFunc)
+
+    gammamat(:,:) = 0.0_dp
+    do ii = 1, this%coulomb%descInvRMat_(M_)
```
bhourahine, line 588: Is this copying the distributed matrix to a local matrix?  The current structure behaves like a serial part of the code, as every process loops over the global matrix sizing.The gemr2d routine might be faster, or alternatively, loop over the local size of the invRMat and use scalafx_indxl2g to find the global location (see calls in coulomb.F90).

cmake/Modules/FindCustomArpack.cmake
bhourahine, line 1: Any reason to drop the finder? This could cope with both arpack and arpackng.

src/dftbp/timedep/linrespcommon.F90
```diff
@@ -858,7 +1241,7 @@ subroutine actionAminusB(spin, wij, win, nocc_ud, nvir_ud, nxoo_ud, nxvv_ud, nxo
 
     ! orb. energy difference diagonal contribution
     vout(:) = vout + wij * vin
-
+
```
bhourahine, line 1244:
```suggestion

```

src/dftbp/timedep/linrespcommon.F90
```diff
@@ -1460,7 +1877,7 @@ subroutine getExcSpin(Ssq, nmatup, getIA, win, eval, xpy, filling, ovrXev, grndE
     integer, allocatable :: TDvin(:)
     logical :: ud_ia, ud_jb
     real(dp) :: s_iaja, s_iaib, s_iajb, tmp
-
+
```
bhourahine, line 1880:
```suggestion

```

src/dftbp/timedep/linrespcommon.F90
```diff
@@ -1405,23 +1825,20 @@ subroutine calcTransitionDipoles(coord0, win, nmatup, getIA, iAtomStart, ovrXev,
     integer :: indm, ii, jj, ss
     real(dp), allocatable :: qij(:)
     logical :: updwn
-
+
```
bhourahine, line 1828:
```suggestion

```

src/dftbp/timedep/linrespcommon.F90
```diff
@@ -689,15 +703,381 @@ subroutine actionAplusB(spin, wij, sym, win, nocc_ud, nvir_ud, nxoo_ud, nxvv_ud,
...
+      end if
+
+    else !--------------spin-polarized systems--------
+
+      call hemv(gtmp, gamma, otmp)
+
+      otmp(:) = 0.0_dp
+
+      !$OMP PARALLEL DO DEFAULT(SHARED) PRIVATE(ia,ss,qij) &
```
bhourahine, line 913: Does OMP make sense inside a MPI region of code?

src/dftbp/timedep/transcharges.F90
```diff
@@ -113,18 +135,156 @@ subroutine TTransCharges_init(this, iAtomStart, sTimesGrndEigVecs, grndEigVecs,
...
+      desc(:) = denseDesc%blacsOrbSqr
+      allocate(workGlobal(nOrb,nOrb,nSpin))
+      nLocRow = size(grndEigVecs,dim=1)
+      nLocCol = size(grndEigVecs,dim=2)
+      allocate(workLocal(nLocRow,nLocCol))
+      allocate(maskMat(nLocRow,nLocCol))
+
+      do iSpin = 1, nSpin
+        do iAtom = 1, this%nAtom
+          workGlobal = 0.0_dp
```
bhourahine, line 195:
```suggestion
          workGlobal(:,:,:) = 0.0_dp
```
Likewise for the next two lines.

src/dftbp/timedep/transcharges.F90
```diff
@@ -332,29 +510,39 @@ pure subroutine TTransCharges_qMatVec(this, iAtomStart, sTimesGrndEigVecs, grndE
...
+    fGlb = iOff + size(vector)
+
     if (this%tCacheChargesOccVir) then
 
-      qProduct(:) = qProduct + matmul(this%qCacheOccVir, vector)
+      qProduct(:) = qProduct + matmul(this%qCacheOccVir(:,iGlb:fGlb), vector)
 
     else
-
+
```
bhourahine, line 534:
```suggestion

```

src/dftbp/timedep/transcharges.F90
```diff
@@ -391,26 +583,37 @@ pure subroutine TTransCharges_qVecMat(this, iAtomStart, sTimesGrndEigVecs, grndE
...
+    fGlb = iOff + size(qProduct)
+
     if (this%tCacheChargesOccVir) then
 
-      qProduct(:) = qProduct + matmul(vector, this%qCacheOccVir)
+      qProduct(:) = qProduct + matmul(vector, this%qCacheOccVir(:,iGlb:fGlb))
 
     else
-
+
```
bhourahine, line 607:
```suggestion

```

src/dftbp/timedep/linrespgrad.F90
```diff
@@ -1239,7 +1401,7 @@ subroutine buildAndDiagExcMatrixStratmann(tSpin, subSpaceFactor, wij, sym, win,
     allocate(evecR(memDim, nExc))
     allocate(workArray(3 * memDim + 1))
     allocate(vecNorm(2 * memDim))
-
+
```
bhourahine, line 1404:
```suggestion

```

src/dftbp/timedep/linrespgrad.F90
```diff
@@ -1259,13 +1421,13 @@ subroutine buildAndDiagExcMatrixStratmann(tSpin, subSpaceFactor, wij, sym, win,
       if (prevSubSpaceDim > 0) then
 
         ! Extend subspace matrices:
-        do ii = prevSubSpaceDim + 1, subSpaceDim
-          call actionAplusB(tSpin, wij, sym, win, nocc_ud, nvir_ud, nxoo_ud, nxvv_ud, nxov_ud,&
-            & nxov_rd, iaTrans, getIA, getIJ, getAB, iAtomStart, ovrXev, grndEigVecs, filling,&
+        do ii = prevSubSpaceDim + 1, subSpaceDim
```
bhourahine, line 1424:
```suggestion
        do ii = prevSubSpaceDim + 1, subSpaceDim
```

src/dftbp/timedep/linrespgrad.F90
```diff
@@ -2065,23 +2233,23 @@ subroutine solveZVectorPrecond(rhs, tSpin, wij, win, nocc_ud, nvir_ud, nxoo_ud,
...
     do kk = 1, nxov**2
 
-      ! action of matrix on vector
+       ! action of matrix on vector      
       call actionAplusB(tSpin, wij, 'S', win, nocc_ud, nvir_ud, nxoo_ud, nxvv_ud, nxov_ud,&
-         & nxov_rd, iaTrans, getIA, getIJ, getAB, iAtomStart, ovrXev, grndEigVecs, occNr, sqrOccIA,&
+         & nxov_rd, iaTrans, getIA, getIJ, getAB, env, denseDesc, ovrXev, grndEigVecs, occNr, sqrOccIA,&
          & gammaMat, species0, spinW, onsMEs, orb, .true., transChrg, pkm1, apk, tRangeSep, lrGamma)
-
+
```
bhourahine, line 2252:
```suggestion

```

src/dftbp/timedep/linrespgrad.F90
```diff
@@ -2065,23 +2233,23 @@ subroutine solveZVectorPrecond(rhs, tSpin, wij, win, nocc_ud, nvir_ud, nxoo_ud,
...
+    
     rkm1(:) = rhs - rkm1
     zkm1(:) = P * rkm1
     pkm1(:) = zkm1
 
     ! Iteration: should be convergent in at most nxov steps for a quadradic surface, so set higher
     do kk = 1, nxov**2
 
-      ! action of matrix on vector
+       ! action of matrix on vector
```
bhourahine, line 2248:
```suggestion
       ! action of matrix on vector
```

bhourahine (commented):
The biggest issue I'm noticing is that the cmake export for the arpack/parpack is now needed, so probably requires them to be built to get this on systems using packaging.

src/dftbp/dftb/scc.F90
thomas-niehaus: Yes, this is to get a global copy of Gamma. The partitioning of RPA vectors is not done with PBLAS, hence there is no point in distributing gamma. Can you come up with a nicer way to build it? I tried a lot of different things. This solution worked at least.

cmake/Modules/FindCustomArpack.cmake
thomas-niehaus: Maybe Balint can comment?

bhourahine: @aradi any views?

bhourahine:
@thomas-niehaus Running on buildbot with 2 MPI procs fails for several tests (block sizes are too small), but also timedep/C66O10N4H44_Ewindow fails for exc_oscillator strength. I used the intel22 MPI with your arpack-ng libraries.

src/dftbp/timedep/transcharges.F90
```diff
@@ -113,18 +135,156 @@ subroutine TTransCharges_init(this, iAtomStart, sTimesGrndEigVecs, grndEigVecs,
...
+      this%tCacheChargesSame = .false.
+      
+    end if
+    
     if (tStoreOccVir) then
+      
+      if (tFirstCall) then
+        @:ASSERT(.not.allocated(this%qCacheOccVir))
+      else
+        ! A simple deallocate gives me an error, why is this%qCacheOccVir
```
bhourahine, line 180: Ans. It's due to the second call to TTransCharges_init from after nxov_rd is redefined due to oscillator windowing in LinRespGrad_old()

src/dftbp/timedep/transcharges.F90
```diff
@@ -113,18 +135,156 @@ subroutine TTransCharges_init(this, iAtomStart, sTimesGrndEigVecs, grndEigVecs,
...
+          do mLoc = 1, size(grndEigVecs, dim=1)
+            mGlb = scalafx_indxl2g(mLoc, desc(MB_), env%blacs%orbitalGrid%myrow, desc(RSRC_), &
+                & env%blacs%orbitalGrid%nrow)
+            if (mGlb >= denseDesc%iAtomStart(iAtom) .and. &
+                & mGlb < denseDesc%iAtomStart(iAtom + 1)) then
+              maskMat(mLoc,:) = sTimesGrndEigVecs(mLoc,:,iSpin)
+            end if
+          end do
+          
+          call pblasfx_pgemm(grndEigVecs(:,:,iSpin), denseDesc%blacsOrbSqr, maskMat, &
```
bhourahine, line 208: As this is inside a loop over atoms, doesn't this make the routine O(N^3) scaling? (~constant number of matrix vector products * natoms)

src/dftbp/timedep/linrespcommon.F90
```diff
@@ -1938,4 +2529,78 @@ subroutine incMemStratmann(memDim, workDim, vecB, vP, vM, mP, mM, mH, mMsqrt, mM
...
+    !> Environment settings
+    type(TEnvironment), intent(in) :: env
+
+    !> Dense matrix descriptor
+    type(TDenseDescr), intent(in) :: denseDesc
+
+    !> Distributed array
+    real(dp), intent(in) :: locArray(:,:)
+
+    !> Global array
```
bhourahine, line 2547:
```suggestion
    !> Global array, i.e. the whole array, but replicated on each member of the comm
```

src/dftbp/timedep/linrespcommon.F90
```diff
@@ -1938,4 +2529,78 @@ subroutine incMemStratmann(memDim, workDim, vecB, vP, vM, mP, mM, mH, mMsqrt, mM
...
+  !> Note: should maybe go into different module and be more general
+  subroutine local2GlobalBlacsArray(env, denseDesc, locArray, glbArray)
+
+    !> Environment settings
+    type(TEnvironment), intent(in) :: env
+
+    !> Dense matrix descriptor
+    type(TDenseDescr), intent(in) :: denseDesc
+
+    !> Distributed array
```
bhourahine, line 2544:
```suggestion
    !> Local part of the distributed array
```

src/dftbp/timedep/linrespcommon.F90
```diff
@@ -1938,4 +2529,78 @@ subroutine incMemStratmann(memDim, workDim, vecB, vP, vM, mP, mM, mH, mMsqrt, mM
...
+    real(dp), intent(in) :: locArray(:,:)
+
+    !> Global array
+    real(dp), intent(out) :: glbArray(:,:)    
+    
+    integer :: desc(DLEN_), iLoc, jLoc, iGlb, jGlb, nLocalRows, nLocalCols, ierr 
+
+    desc(:) = denseDesc%blacsOrbSqr
+
+    call scalafx_getlocalshape(env%blacs%orbitalGrid, denseDesc%blacsOrbSqr, nLocalRows,&
```
bhourahine, line 2554:
```suggestion
    call scalafx_getlocalshape(env%blacs%orbitalGrid, desc, nLocalRows, nLocalCols)
```

src/dftbp/timedep/linrespcommon.F90
```diff
@@ -1938,4 +2529,78 @@ subroutine incMemStratmann(memDim, workDim, vecB, vP, vM, mP, mM, mH, mMsqrt, mM
...
+        & nLocalCols)
```
bhourahine, line 2555:
```suggestion
```

src/dftbp/timedep/linrespcommon.F90
```diff
@@ -1938,4 +2529,78 @@ subroutine incMemStratmann(memDim, workDim, vecB, vP, vM, mP, mM, mH, mMsqrt, mM
...
+
+    glbArray = 0.0_dp
```
bhourahine, line 2557:
```suggestion
    glbArray(:,:) = 0.0_dp
```

src/dftbp/timedep/linrespcommon.F90
thomas-niehaus: You don't think it will be executed OMP?

src/dftbp/timedep/transcharges.F90
thomas-niehaus: ok!

src/dftbp/timedep/transcharges.F90
```diff
@@ -113,18 +135,156 @@ subroutine TTransCharges_init(this, iAtomStart, sTimesGrndEigVecs, grndEigVecs,
...
+            if (ss == iSpin) then
+              if (kk <= nMatUp) then
+                this%qCacheOccVir(iAtom, ia) = 0.5_dp * (workGlobal(ii,aa,1)+workGlobal(aa,ii,1))
+              else
+                this%qCacheOccVir(iAtom, ia) = 0.5_dp * (workGlobal(ii,aa,2)+workGlobal(aa,ii,2))
+              end if
+            end if 
+          enddo
+
+          if (tStoreSame .and. tFirstCall) then
```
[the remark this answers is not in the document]
thomas-niehaus: These arrays are currently not affected by windowing. In principle the set of required occ-occ, vir-vir is reduced by windowing, which means that at the current stage LC-DFTB could profit from that. Let's keep that in mind, would require some major changes.

thomas-niehaus:
> @thomas-niehaus Running on buildbot with 2 MPI procs fails for several tests (block sizes are too small), but also timedep/C66O10N4H44_Ewindow fails for exc_oscillator strength. I used the intel22 MPI with your arpack-ng libraries.

Is adding a small block size in the tests a solution? Not sure if the serial tests would accept that. I actually would like to use the same test cases for MPI/serial. Otherwise we would need to double the tests to capture MPI problems. Will look into the osc strength/ window problem.

bhourahine:
@thomas-niehaus Something got broken for both the xlbomd and derivatives tests, which appears for 2 processors:
cmake -DTEST_MPI_PROCS=2 ..; ctest -R deriv && ctest -R xlbomd

bhourahine:
> @thomas-niehaus Something got broken for both the xlbomd and derivatives tests, which appears for 2 processors: cmake -DTEST_MPI_PROCS=2 ..; ctest -R deriv && ctest -R xlbomd

Might be specific to one test machine, I'll get back to you about this.

bhourahine:
@thomas-niehaus Branch changes now merged in. I noticed that dipole intensities for propadiene_OscWindow-Stratmann are OK for 1 processor, but differ for 2.

bhourahine:
@thomas-niehaus the N2_onsite test (and a few surrounding ones) doesn't have a trap for the number of processors. In that specific one, 16 processors fails (probably by that time some matrix isn't distributed properly). Can you add a MPI_PROCS <= 16 or similar to all of timedep/2CH3-Temp, timedep/2CH3-Triplet-Temp, timedep/NO, timedep/NO_onsite, timedep/N2_onsite, timedep/propadiene_OscWindow and have a look why the N2_onsite the the only one failing on 16 procs ?

bhourahine:
> @thomas-niehaus I've just checked in a limit to the tests to <=8 procs and will see if I can find out why the N2_onsite was different.

The error is inside the call to pdsaupd, so probably not much we can do.

thomas-niehaus:
All tests (serial/parallel) pass locally, the buildbot errors seem to be due to the CMAKE arpack-ng issue. @Balint: could you have a look at it? Otherwise the PR seems to be ready.

thomas-niehaus:
One final thing. If i compile the serial version with ifort and all checks on,  I get a runtime error  "Attempt to fetch from allocatable variable LOCSIZE when it is not allocated". In the serial version the locSize array is never allocated, used or touched. It is just passed (unallocated) to buildAndDiagExcMatrixArpack as integer, intent(in) :: locSize(:). This is sufficient to throw the error. Any ideas why this happens?

bhourahine:
@thomas-niehaus That's a missing allocatable from the variable definition in the affected subroutine, as intel tests this in more recent versions. I'll push changes to affected variables in a couple of minutes.

src/dftbp/timedep/linrespcommon.F90
bhourahine: If it's executed as MPI with current structures, probably not at the moment. But if in future we start using node level parallelism, possibly.

src/dftbp/timedep/transcharges.F90
bhourahine: We can look at this again after the branch is merged.

src/dftbp/timedep/transcharges.F90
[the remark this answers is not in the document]
bhourahine: OK,  so possible further pull request after this one...

bhourahine:
> > @thomas-niehaus Something got broken for both the xlbomd and derivatives tests, which appears for 2 processors: cmake -DTEST_MPI_PROCS=2 ..; ctest -R deriv &.& ctest -R xlbomd
> 
> Might be specific to one test machine, I'll get back to you about this.

Yes, machine specific, so not a problem for your changes (will need to investigate why the main branch was failing for these tests)

aradi:
Funcionality already added in #1493

PR #1275, opened 2023-07-07
Move file wfn printout above convergence terminator

### LINKED ISSUE #1273: Warning for property calculations if SCC fails
**Describe the bug**
There is no indication that various property calculations are not performed if ConvergentSCCOnly=Yes and the SCC fails. If old files are left in the directory, this can confuse users with mismatching data.

**To Reproduce**
KLines calculation with SccIterations=1 and some property (eigenvectors, projection of states, ...) requested.

**Expected behaviour**
Blanking various property files and/or printing warnings that later stages of the calculation requested in the input are not performed.

bhourahine:
Hence if these are requested for a non-converged calc, they are still output.

Closes #1273

### DIFF
--- a/src/dftbp/dftbplus/main.F90
+++ b/src/dftbp/dftbplus/main.F90
@@ -1442,12 +1442,33 @@ contains
       end if
     end if
 
+    call env%globalTimer%startTimer(globalTimers%eigvecWriting)
+    if (this%tPrintEigVecs) then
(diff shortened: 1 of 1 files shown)


bhourahine:
Will put an alternative solution to #1273 in a separate branch, an we can decide which is better.

bhourahine:
WF coefficients according to the recipes can't currently be written without this change, so escalating to a bug.

PR #1223, opened 2023-05-03
General periodic hybrid-DFTB formalism

vanderhe:
General, periodic hybrid-DFTB formalism, as well as a variety of related improvements.

Main features (+ a lot of minor improvements):
- generalization of hybrid-DFTB ground-state Hamiltonian to general CAM functionals
(non-periodic systems)
- MPI-parallelization of matrix-based hybrid-DFTB ground-state Hamiltonian construction
(non-periodic systems)
- introduction of highly optimized matrix-based energy gradient evaluation
(non-periodic systems)
- neighbour-list based hybrid-DFTB ground-state Hamiltonian and energy gradients for general CAM functionals
(periodic, Gamma-only, MPI-parallel version restricted to Hamiltonian construction)
- matrix-based hybrid-DFTB ground-state Hamiltonian and energy gradients for general CAM functionals
(periodic, Gamma-only, MPI-parallel versions available)
- neighbour-list based hybrid-DFTB ground-state Hamiltonian and energy gradients for general CAM functionals
(periodic, general k-points, MPI-parallel version restricted to Hamiltonian construction)
- related infrastructure, including but not limited to new (parallel) dense Mulliken analysis, q0 subtraction, real <--> dual space density matrix transformation, ...

Known issues/missing:
- periodic systems limited to spin-restricted calculations
(routines do support co-linear spin-polarization throughout but haven't been extensively tested yet)
- licensing of Wigner-Seitz related routine needs discussion
- energy gradients of general k-point implementation not yet optimized/MPI-parallelized
- neighbour-list based, Gamma-only Hamiltonian construction could probably be more efficient
- possible incompatibility with other extension due to generalization to general CAM functionals
(hybrid TD-DFTB, ...)
- due to the lack of database, a lot of SK-files checked in for regression testing
- general k-point implementation does not yet utilize MPI-windows and stores Hamiltonian/density matrices on all MPI ranks, leading to relatively high memory consumption
- limited support/testing of restarting calculations
- limitations in k-point sampling, needs to be MP-like with zero shift
- TCI cutoff search might be inefficient, however not performance critical in the slightest
- stress tensor not implemented for all available periodic, hybrid-DFTB ground-state calculations
- slightly inconsistent variable naming in hybridXc module
- general k-point implementation does not yet provide energy gradients for the MIC algorithm
- possible duplicate of ordinary gamma function by full-range HF gamma
- ...

Possible reviewing strategy:
[...]

PR #1280, opened 2023-07-13
Ordering agnostic SK-parsing

### LINKED ISSUE #1269: Sequence dependent parsing of hybrid SK extra-tag
**Describe the bug**
The current iteration of the SK-file parser is sensitive to the order in which the repulsive and hybrid-DFTB extra-tag appear. As long as the repulsive is in front of the extra-tag, parsing behaves as expected. The other way around, however, absolutely random behavior with potentially wrong results (if the code does not accidentally halt e.g. due to a negative range-separation parameter) occurs.

(Kudos to Peter Broqvist, who pointed this out, when testing #1223)

**To Reproduce**
Run a hybrid-DFTB calculation and reverse the order of aforementioned entries in the SK-files.

**Expected behaviour**
I would suggest to make the parser ordering-agnostic in this regard.

vanderhe:
Fixes #1269.

Remark: While it seems like this fix is automatically tested by updating the testparams commit ID, it is not. I verified it by (manually) applying the patch to #1223, however, proper regression testing will only be in place as soon as #1223 is merged.

### DIFF
--- a/src/dftbp/type/oldskdata.F90
+++ b/src/dftbp/type/oldskdata.F90
@@ -182,12 +182,9 @@ contains
       end if
     end do
 
-    if (.not. present(splineRepIn)) then
-      call closeFile(file)
-      return
+    if (present(splineRepIn)) then
+      call readSplineRep(file%unit, fileName, splineRepIn, iSp1, iSp2)
     end if
-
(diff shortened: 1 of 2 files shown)


bhourahine (approved):
LGTM

PR #1281, opened 2023-07-19
Fix for incorrect pointers

### LINKED ISSUE #1266: Memory leak when iterating in DFTB+
**Describe the bug**
I have now tried installing DFTB+ from the current developement version and a number of different compiler, lapack libraries, machines, and DFTB+ complation flags.
So far I have found in all cases that running the GeometryOptimization for several iterations leads to a growing memory consumption.
This does however not cause a crash since it is only some hundreds of kB every cycle, 
but I am intending to use the library function of DFTB+, and it is therefore important that the program doesn't slowly accumulate memory as it will crash a calculation running for more than 1000 DFTB+ steps.

**To Reproduce**
I have tried many different setups so currently I have verified it to be the case for:

MKL 2018.3, 2023.1.0, Lapack.
gfortran 9, 11, and 12.
Turning on an off thread support in DFTB didn't change anything either.
Turning on and off the D3 dispersion library and inputs didn't change anything either.

**Expected behaviour**
The memory consumption remains flat over the run.

**Additional context**
I have also tried running the code through Valgrind, but there are no obvious unreachable memory that were growing even after running with more and more steps in the calculation.

bhourahine:
Note that other MPI-only parts of the window code probably should be inside pre-processor blocks, so probably need a further PR to refact that.

Closes #1266

### DIFF
--- a/src/dftbp/dftb/periodic.F90
+++ b/src/dftbp/dftb/periodic.F90
@@ -898,16 +898,6 @@ contains
       call neigh%neighDist2Win_%unlock()
     #:endif
     else
-      if (associated(neigh%iNeighbourMem_)) then
-        deallocate(neigh%iNeighbourMem_)
-      end if
-      if (associated(neigh%neighDist2Mem_)) then
-        deallocate(neigh%neighDist2Mem_)
-      end if
-
-      allocate(neigh%iNeighbour(0:maxNeighbour, 1:nAtom))
-      allocate(neigh%neighDist2(0:maxNeighbour, 1:nAtom))
-
       neigh%iNeighbour(1:,:) = iNeighbour(1:maxNeighbour,:)
       neigh%neighDist2(1:,:) = neighDist2(1:maxNeighbour,:)
     end if


src/dftbp/dftb/periodic.F90
```diff
@@ -898,11 +898,11 @@ subroutine fillNeighbourArrays(neigh, iNeighbour, neighDist2, startAtom, endAtom
...
+      if (associated(neigh%iNeighbour)) then
+        deallocate(neigh%iNeighbour)
       end if
-      if (associated(neigh%neighDist2Mem_)) then
-        deallocate(neigh%neighDist2Mem_)
+      if (associated(neigh%neighDist2)) then
+        deallocate(neigh%neighDist2)
       end if
 
       allocate(neigh%iNeighbour(0:maxNeighbour, 1:nAtom))
```
terminationshock, line 908: I would argue that we could delete the entire deallocation-allocation from this routine, because there is `allocateNeighbourArrays` being called right before it. Could you please check that?

terminationshock (changes requested):

src/dftbp/dftb/periodic.F90
bhourahine: Now removed.

terminationshock (approved):

PR #1282, opened 2023-07-19
Use common header code

bhourahine:
Follow-up to #1255


vanderhe (approved):
Ready to go from my side.

PR #1284, opened 2023-07-24
Fixed window locks called unitialised

bhourahine:
Also moved more of type inside of MPI conditional block and commented up why the logical has to be available in serial, despite referring to MPI. This variable could be factored out of the serial path, but would need a duplicate non-windowed neighbor case for MPI, since the parser calls this as well for Slater-Kirkwood dispersion before the computational enviroment is setup for MPI.

Also changed isParallel use in relevant if() blocks over to the type's MPI window variable, just in case that ever becomes an optional switch.

### DIFF
--- a/src/dftbp/dftb/periodic.F90
+++ b/src/dftbp/dftb/periodic.F90
@@ -64,27 +64,33 @@ module dftbp_dftb_periodic
     !> whether the neighbour list has been set by an API call
     logical :: setExternally = .false.
 
-    ! Whether memory should be allocated via MPI-windows (or directly via allocate() otherwise)
+    !> Whether memory should be allocated via MPI-windows (or directly via allocate() otherwise)
+    !!
+    !! Note: this variable cannot be inside the MPI block below with current code, as the parser
+    !! uses this data structure before the MPI environment is set up (in addition to the main code
+    !! use). This leads to a fall back to the usual distributed case if MPI is enabled, instead of
+    !! the shared window in that case.
     logical, private :: useMpiWindows_ = .false.
 
-    ! memory allocated for the iNeighbour array
+  #:if WITH_MPI
+
+    !> memory allocated for the iNeighbour array
     integer, pointer, private :: iNeighbourMem_(:) => null()
 
-    ! memory allocated for the neighDist2 array
+    !> memory allocated for the neighDist2 array
     real(dp), pointer, private :: neighDist2Mem_(:) => null()
 
-  #:if WITH_MPI
-    ! MPI shared memory window handler for iNeighbour
+    !> MPI shared memory window handler for iNeighbour
     type(mpifx_win), private :: iNeighbourWin_
 
-    ! MPI shared memory window handler for neightDist2
+    !> MPI shared memory window handler for neightDist2
     type(mpifx_win), private :: neighDist2Win_
-
-  #:endif
 
   contains
 
     final :: TNeighbourList_final
+
+  #:endif
 
   end type TNeighbourList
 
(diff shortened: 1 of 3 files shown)


src/dftbp/dftb/periodic.F90
```diff
@@ -116,31 +122,24 @@ subroutine TNeighbourList_init(this, nAtom, nInitNeighbour)
...
         call this%neighDist2Win_%free()
       end if
-    #:endif
-    else
-      if (associated(this%iNeighbourMem_)) then
-        deallocate(this%iNeighbourMem_)
-      end if
-      if (associated(this%neighDist2Mem_)) then
-        deallocate(this%neighDist2Mem_)
-      end if
```
terminationshock, line 140: I think, we need this routine even if MPI windows are not used. `*Mem_` are independent of the window stuff.

src/dftbp/dftb/periodic.F90
```diff
@@ -64,28 +64,34 @@ module dftbp_dftb_periodic
...
+    !> Whether memory should be allocated via MPI-windows (or directly via allocate() otherwise)
+    !!
+    !! Note: this variable cannot be inside the MPI block below with current code, as the parser
+    !! uses this data structure before the MPI environment is set up (in addition to the main code
+    !! use). This leads to a fall back to the usual distributed case if MPI is enabled, instead of
+    !! the shared window in that case.
     logical, private :: useMpiWindows_ = .false.
 
-    ! memory allocated for the iNeighbour array
+  #:if WITH_MPI
```
terminationshock, line 75: See my other comment. I think, we need the `*Mem_` also if MPI windows is not used.

src/dftbp/dftb/periodic.F90
```diff
@@ -874,7 +875,7 @@ subroutine fillNeighbourArrays(neigh, iNeighbour, neighDist2, startAtom, endAtom
     integer :: maxNeighbourLocal
   #:endif
 
-    if (isParallel) then
+    if (neigh%useMpiWindows_) then
```
terminationshock, line 878: If `isParallel` is indeed not required, we could also remove it from the routine's signature.

terminationshock:
[isParallel.zip](https://github.com/dftbplus/dftbplus/files/12191967/isParallel.zip)
I have created a patch to your commit a82e4e6b (zipped, because github does not like patch files). This removes the `isParallel` keyword. The logic is otherwise really difficult to understand.
It is important that `useMpiWindows_` is set to `.false.` if `WITH_MPI` is `0`.

terminationshock:
And I revoke my comments that `*Mem_` is required in the non-parallel case.

bhourahine:
> [isParallel.zip](https://github.com/dftbplus/dftbplus/files/12191967/isParallel.zip) I have created a patch to your commit [a82e4e6](https://github.com/dftbplus/dftbplus/commit/a82e4e6ba203a72f65793b49b8fb975eba7edab6) (zipped, because github does not like patch files). This removes the `isParallel` keyword. The logic is otherwise really difficult to understand. It is important that `useMpiWindows_` is set to `.false.` if `WITH_MPI` is `0`.

Thanks, I'll work through and incorporate as a co-authored PR. For future similar cases, would it be simpler to make a PR into the PR, which we can then squash down?

terminationshock:
Sure that would also work for me. The patch was also meant as a proposal, so feel free to edit it again.

bhourahine:
@terminationshock looks like the API tests will now also need to be adapted.

bhourahine:
> @terminationshock looks like the API tests will now also need to be adapted.

OK, think I've fixed it (but had to make the assumption that the MPI window is set up for the API call case, DFTB+ also needs a tester for setting neighbours under that situation).

terminationshock:
I tested this commit with Lammps and it seems to be fine. I agree, a regression test would be nice. Let's discuss this in the next meeting.

bhourahine:
Together with #1281 and #1283 this should fully close the MPI window issue in #1266

PR #1286, opened 2023-07-26
Test for memory accumulation in longer MD

bhourahine:
Custom run script for an indefinite length MD on a small system. It checks for fraction of memory use and whether the md can be cleanly stopped with an interupt file.


bhourahine:
The alternative would be to get the binary to self-document its memory use via reading data from /proc/. Will need some C code added though.

PR #1287, opened 2023-07-27
Revise quantities printed in md.out

### LINKED ISSUE #1272: Saving forces and stress in each MD step to use the results for MLFF training
I am trying to use DFTB+ MD results to train Machine Learning Force Field (MLFF) model to obtain interatomic potential to be used with e.g. LAMMPS (See, e.g., [KLIFF](https://github.com/openkim/kliff), [MAML](https://github.com/materialsvirtuallab/maml), [GAP](https://libatoms.github.io/), [MLIP](https://mlip.skoltech.ru/), [SNAP](https://github.com/FitSNAP/FitSNAP), [PANNA](https://gitlab.com/PANNAdevs/panna), [AMP](https://bitbucket.org/andrewpeterson/amp), [DeePMD](https://github.com/deepmodeling/deepmd-kit), [SchNet](https://github.com/atomistic-machine-learning/schnetpack), [NequIP](https://github.com/mir-group/nequip), [FLARE](https://github.com/mir-group/flare), ......).

However, this requires writing the following at each MD step:
-forces
-stress (per structure)
-energy (per structure)
-structure

I know that DFTB+ can write forces using the `CalculateForces` flag. However, it writes forces for only the last step. Same for stress and energy.

DFTB+ also writes the structure every `MDRestartFrequency` steps, but for periodic systems, it writes only xyz without unit cell axes. The axes are written only at the final step to the .gen file.

So, is there a method to save forces, stress, and energy at each MD step?
Is it possible to write the full periodic structure (positions+cell vectors) after each MD step?

vanderhe:
Remark: In order to print atomic forces to `md.out`, the HSD block
```
Analysis {
  CalculateForces = Yes
}
```
needs to be set explicitly. Lattice vectors and the total stress tensor are printed as well, atomic positions however are only written to the `.xyz` trajectory file, not `md.out`.

- [x] documentation (does not seem to be affected)

Possibly closes #1272.

### DIFF
--- a/src/dftbp/dftbplus/mainio.F90
+++ b/src/dftbp/dftbplus/mainio.F90
@@ -4028,17 +4028,17 @@ contains
       do ii = 1, size(derivs, dim=2)
         write(fd, "(3E24.8)") -derivs(:, ii)
       end do
+      if (hasStress) then
+        write(fd, "(A)") "Total stress (au)"
+        do ii = 1, size(totalStress, dim=2)
+          write(fd, "(3E24.8)") totalStress(:,ii)
+        end do
+      end if
     end if
     if (withBarostat) then
       write(fd, "(A)") "Lattice vectors (A)"
       do ii = 1, size(latVec, dim=2)
         write(fd, "(3E24.8)") latVec(:,ii) * Bohr__AA
-      end do
-    end if
-    if (hasStress) then
-      write(fd, "(A)") "Total stress (au)"
-      do ii = 1, size(totalStress, dim=2)
-        write(fd, "(3E24.8)") totalStress(:,ii)
       end do
     end if
     if (isPeriodic) then
--- a/src/dftbp/dftbplus/mainio.F90
+++ b/src/dftbp/dftbplus/mainio.F90
@@ -3356,7 +3356,7 @@ contains
         write(fd, format2U) 'Energy DFTB+U', energy%Edftbu, 'H', energy%Edftbu * Hartree__eV, 'eV'
       end if
       if (tOnSite) then
-        write (fd,format2U) 'Energy onsite', energy%eOnSite, 'H', energy%eOnSite*Hartree__eV, 'eV'
+        write(fd,format2U) 'Energy onsite', energy%eOnSite, 'H', energy%eOnSite*Hartree__eV, 'eV'
       end if
     end if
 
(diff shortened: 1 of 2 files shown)


bhourahine (commented):
The barostat logic was due to the need to print the lattice vectors as they are changing. But switching to always printing them is safe, but uncessesarily makes the file a bit bigger when there is no barostat.

vanderhe:
@aradi I have reverted the changes concerning the barostat logic, so that the lattice vectors are only printed if a barostat is used (restores previous behavior). Ready for review from my side.

src/dftbp/dftbplus/mainio.F90
```diff
@@ -4014,60 +4023,75 @@ subroutine writeMdOut2(fd, tStress, tPeriodic, tBarostat, isLinResp, eField, tFi
...
-            & eField%absEField * au__V_m, 'V/m'
+        write(fd, format1U1e) "External E field", eField%absEField, "au",&
+            & eField%absEField * au__V_m, "V/m"
       end if
     end if
-    if (tFixEf .and. tPrintMulliken) then
-      write(fd, "(A, F14.8)") 'Net charge: ', sum(q0(:, :, 1) - qOutput(:, :, 1))
+
+    if (fixEf .and. printMulliken) then
+      write(fd, "(A, F14.8)") "Net charge: ", sum(q0(:, :, 1) - qOutput(:, :, 1))
```
bhourahine, line 4083: The coverage report that there is no fixed Fermi test case reaching this line, so is it worth adding a short one?

src/dftbp/dftbplus/mainio.F90
```diff
@@ -4014,60 +4023,75 @@ subroutine writeMdOut2(fd, tStress, tPeriodic, tBarostat, isLinResp, eField, tFi
     integer :: ii
     character(lc) :: strTmp
 
-    if (tStress) then
-      if (tBarostat) then
-        write(fd, "(A)") 'Lattice vectors (A)'
+    if (isPeriodic) then
```
bhourahine, line 4026: What happens if helical (the boundary derivatives could be useful there as well)? Might be better to pass in the TboundaryConditions variable in main and test for boundaryCond%iBoundaryCondition .not. boundaryConditions%cluster or only test withBarostat and print lattice vectors then (assuming withBarostat is set for any future helical case).

src/dftbp/dftbplus/mainio.F90
```diff
@@ -4014,60 +4023,75 @@ subroutine writeMdOut2(fd, tStress, tPeriodic, tBarostat, isLinResp, eField, tFi
...
       end if
-      if (tPeriodic) then
-        write(fd, format2Ue) 'Pressure', cellPressure, 'au', cellPressure * au__pascal, 'Pa'
-        if (pressure /= 0.0_dp) then
-          write(fd, format2U) 'Gibbs free energy', energy%EGibbs, 'H',&
-              & Hartree__eV * energy%EGibbs,'eV'
-          write(fd, format2U) 'Gibbs free energy including KE', energy%EGibbsKin, 'H',&
-              & Hartree__eV * energy%EGibbsKin, 'eV'
-        end if
+      if (printForces) then
```
bhourahine, line 4033: Should this be inside the isPeriodic block, as forces for molecular cases can be requested.

bhourahine (commented):
See comments.

src/dftbp/dftbplus/mainio.F90
```diff
@@ -5743,7 +5767,7 @@ subroutine writeReksDetailedOut1(fd, nGeoSteps, iGeoStep, tMD, tDerivs,&
       write(fd, format2U) 'Energy SCC', energy%ESCC, 'H', energy%ESCC * Hartree__eV, 'eV'
       write(fd, format2U) 'Energy SPIN', energy%Espin, 'H', energy%Espin * Hartree__eV, 'eV'
       if (t3rd) then
-        write (fd,format2U) 'Energy 3rd', energy%e3rd, 'H', energy%e3rd*Hartree__eV, 'eV'
+        write(fd,format2U) 'Energy 3rd', energy%e3rd, 'H', energy%e3rd*Hartree__eV, 'eV'
```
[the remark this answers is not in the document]
vanderhe: Would certainly not hurt to have one. Of course it would only test if it runs, not that anything reasonable is printed to {detailed|md}.out. Separate PR?

src/dftbp/dftbplus/mainio.F90
vanderhe: See comment above.

src/dftbp/dftbplus/mainio.F90
vanderhe: No it shouldn't, thanks for the fix.

src/dftbp/dftbplus/mainio.F90
vanderhe: Not entirely sure about the helical case and which quantities are affected. I guess the way the lattice vectors are printed (and converted unit-wise) in case of helical boundary conditions is already broken. Also the cell volume is always zero? What about all the other quantities, do you have a suggestion how to fix it elegantly?

bhourahine: We don't change the helical vectors during a calculation at the moment, so can just leave this case out and fix later. 
For reference, it is a (3,1) array, where the first two values are a distance and an angle, while the last value is an integer stored as a real.
The longer term solution is probably to put a writer routine as a method for each boudary condition.

src/dftbp/dftbplus/mainio.F90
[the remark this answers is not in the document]
bhourahine: Separate PR :+1:

src/dftbp/dftbplus/mainio.F90
vanderhe: I have now changed two things:

1. the total stress is only printed if printing forces is requested, as of your suggestion
2. the block that prints the lattice vectors moved into the `isPeriodic` block, as an extra layer of protection to prevent printing something with wrong unit descriptor (and utilized conversion factor) in case of helical boundary conditions.

Do you agree with those changes or did I overlook something?

bhourahine:
Can you please rebase on main and I'll then merge it.

vanderhe:
> Can you please rebase on main and I'll then merge it.

@bhourahine Done.

bhourahine (approved):
Ready to merge

PR #1288, opened 2023-07-27
Correct keyword in setupgeom section

bhourahine:
Also remove tab characters.


vanderhe (approved):
Looks good, in line with [#58](https://github.com/dftbplus/recipes/pull/58).

PR #1289, opened 2023-07-30
GPU density matrix

bhourahine:
Density matrix build on GPUs.
Needs:
- [x] Extra test case for non-positive definite fillings (MP fill routine for example)
- [X] Density matrix build switch between algorithms
- [X] Real DM and EDM
- [X] Complex DM case
- [x] Complex EDM
- [X] Documentation


bhourahine:
Incorporates #1253 with additional changes

PR #1290, opened 2023-08-02
Shell script copyright dates

bhourahine:
Trivial PR for dates in scripts

Issue #13, opened 2017-06-23
Dumping HSD with large text blocks fails


aradi:
When compiled with certain compilers (especially ifort) the executables fail to dump the hsd tree when it contains large text blocks (huge geometry, Hessian matrix, etc.) due to record overflow.

PR #1291, opened 2023-08-02
Update script path for sh headers

bhourahine:
Follow-up to #1290

### DIFF
--- a/utils/srcmanip/set_source_headers
+++ b/utils/srcmanip/set_source_headers
@@ -43,6 +43,8 @@ FILE_TYPES = [
     ('src/**/*.h', 'c'),
     ('src/**/*.fypp', 'fypp'),
     ('src/**/make.*', 'makefile'),
+    ('test/app/dftb+/*/*.sh', 'shell'),
+    ('test/app/dftb+/*/*/*.sh', 'shell'),
     ('test/app/dftb+/bin/*', 'shell'),
     ('test/app/modes+/bin/*', 'shell'),
     ('test/app/phonons/bin/*', 'shell'),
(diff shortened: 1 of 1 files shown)


vanderhe (commented):
Is it a problem that e.g. in the testrun.sh files the shebang is below the copyright-header after executing `set_source_headers`? Might be a stupid question, just wondered...

bhourahine:
> Is it a problem that e.g. in the testrun.sh files the shebang is below the copyright-header after executing `set_source_headers`? Might be a stupid question, just wondered...

That will run as a script with whatever the default shell is as interpreter, so real comments should definitely be after the shebang.

vanderhe:
> > Is it a problem that e.g. in the testrun.sh files the shebang is below the copyright-header after executing `set_source_headers`? Might be a stupid question, just wondered...
> 
> That will run as a script with whatever the default shell is as interpreter, so real comments should definitely be after the shebang.

I see. What about the command line wrappers in `dptools/bin`, there the shebang is above the copyright header, even after executing the script and therefore not consistent with the `testrun.sh` files. Do we care about that?

bhourahine:
The shebang should always be at the top. I didn't write this script so haven't worked out why it is chewing these up. To be honest, this looks over engineered and fragile (a suitable regex replacement on the DFTB+ copyright string with a range of quote marks in the match, plus the unix find, would work and not need the list of paths to be maintained).

vanderhe:
> The shebang should always be at the top. I didn't write this script so haven't worked out why it is chewing these up. To be honest, this looks over engineered and fragile (a suitable regex replacement on the DFTB+ copyright string with a range of quote marks in the match, plus the unix find, would work and not need the list of paths to be maintained).

@bhourahine See my suggestion of [#13](https://github.com/bhourahine/dftbplus/pull/13).

bhourahine:
@vanderhe shall we just merge this?

PR #1299, opened 2023-08-09
Dipole correction

aradi:
Dipole correction for slabs.

Currently only works for neutrals systems, slab has to be aligned along x/y with vacuum along z.

Todo:
* [X] Correction for neutral systems
* [ ] More detailed/meaningful test of the energy expression (potential seems to behave correctly)
* [ ] Generalize for charged cases
* [ ] Allow for x- or y- as orthogonal axis in user input (internally the axis is parameterized already)


aradi:
@terminationshock You might give it a try at its current stage, the buildbot tests pass.

terminationshock:
Thank you, @aradi . I have installed that version on our colleagues's VM. I will let you know when their tests are done.

aradi:
Ping to keep it alive.

terminationshock:
Should we also merge this soon? I would say, the feature is well-tested by the colleagues.

bhourahine:
@terminationshock now merged

PR #1301, opened 2023-08-13
Periodic contacts

bhourahine:
Fixes transport calculation periodic geometries

- [x] Correct lattice points for omitted directions
- [x] Correct coordinate folding for ommitted directions
- [ ] Pseudo-inversion for sub-sets of directions
- [ ] Correct test for open boundary periodic directions
- [ ] Boundary condition structure updated for contact information (WIP)
- [ ] Identify contact directions
- [ ] Replicate atoms of the PLs of contacts out to the cutoff distance
- [ ] Indexing for contact atoms


bhourahine:
Will deffer extending contacts to the contact distance to another PR.

bhourahine:
@aradi @pecchia I think this is now final (but a follow up PR to apply the same fixes to the phonon code is also needed at some point, depending on what is happening about the neighbour lists in that code).

bhourahine:
Seems to be an impact on one of the ELSI tests according to the buildbot collection.

PR #1303, opened 2023-08-15
Move from deprecated icc to icx Intel compiler

vanderhe:
Updates GitHub workflow to use the ICX compiler, instead of the deprecated and soon to be removed ICC.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -188,7 +188,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         fc: [ifort]
-        cc: [icc]
+        cc: [icx]
     env:
       FC: ${{ matrix.fc }}
       CC: ${{ matrix.cc }}


bhourahine (commented):
We probably should discuss before merging, but good to get a clang based compiler into the mix. Questions I have:
1. Why not also ifx?
2. Why not keep icc as well as a separate workflow for a while?
3. Is mixing ifort/icx a good idea?

vanderhe:
> We probably should discuss before merging, but good to get a clang based compiler into the mix. Questions I have:
> 
>     1. Why not also ifx?
> 
>     2. Why not keep icc as well as a separate workflow for a while?
> 
>     3. Is mixing ifort/icx a good idea?

Yes, I agree. I did not expect that this PR gets merged immediately, just wanted to trigger a discussion, since the deprecation warning is around for quite some time now.

bhourahine:
> Yes, I agree. I did not expect that this PR gets merged immediately, just wanted to trigger a discussion, since the deprecation warning is around for quite some time now.

Now added to the next agenda.
The intel22 ifx doesn't build the code at the moment. The first error I hit is having a procedure in the interface for dftb/coordnumber.F90:getCoordinationNumber().

bhourahine (approved):
As per discussion, icx is ready. At a later stage we need to look at ifx again.

PR #1307, opened 2023-09-05
Generalised Born sym neighbours

bhourahine:
Uses symmetric neighbour lists from #1223 so will need to be rebased on that branch once it is in turn ready for merging.

- [x] call Born solvent model with symmetric tests
- [x] recalculate affected tests
- [ ] verify correctness of results (???)


bhourahine:
Only 750ee90 relates to the solvent changes.

PR #1310, opened 2023-09-15
multiple calls to setExternalCharges in qm/mm

tjgiese:
Working on the amber/dftb+ QM/MM implementation. I call setExternalCharges at each MD time step to inform dftb+ about the nearby MM charges within the direct-space cutoff; however, the dftb+ allocations only occur on the first MD step. If subsequent steps encounter more nearby MM charges, then setExternalCharges will write passed the end of the arrays. This commit checks the allocated sizes and reallocates if necessary.

### DIFF
--- a/src/dftbp/dftb/extcharges.F90
+++ b/src/dftbp/dftb/extcharges.F90
@@ -25,7 +25,7 @@ module dftbp_dftb_extcharges
     private
 
     !> Number of point charges
-    integer, public :: nChrg
+    integer :: nChrg
 
     !> Number of atoms
     integer :: nAtom
@@ -80,6 +80,9 @@ module dftbp_dftb_extcharges
     !> Copy Q * inverse R contribution for the point charges
     procedure :: copyInvRvec
 
+    !> Returns the number of external charges
+    procedure :: getNumCharges
+    
   end type TExtCharges
 
 
@@ -316,5 +319,18 @@ contains
 
   end subroutine copyInvRvec
 
+  
+  !> Returns the number of external charges
+  function getNumCharges(this) result(n)
+    
+    !> Instance of SCC calculation
+    class(TExtCharges), intent(in) :: this
 
+    integer :: n
+    
+    n = this%nChrg
+    
+  end function getNumCharges
+
+  
 end module dftbp_dftb_extcharges
--- a/src/dftbp/dftb/extcharges.F90
+++ b/src/dftbp/dftb/extcharges.F90
@@ -25,7 +25,7 @@ module dftbp_dftb_extcharges
     private
 
     !> Number of point charges
-    integer :: nChrg
+    integer, public :: nChrg
 
     !> Number of atoms
     integer :: nAtom
--- a/src/dftbp/dftb/populations.F90
+++ b/src/dftbp/dftb/populations.F90
@@ -533,6 +533,8 @@ contains
     integer :: iAt, iSp, iSh
     integer :: nAtom, nSpin
 
+    allocate( qq(size(qorb,1),size(qorb,2),size(qorb,3)) )
+    
     if (present(qRef)) then
       qq = qOrb - qRef
     else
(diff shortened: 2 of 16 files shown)


src/dftbp/dftb/populations.F90
```diff
@@ -533,6 +533,8 @@ subroutine getChargePerShell(qOrb, orb, species, chargePerShell, qRef)
     integer :: iAt, iSp, iSh
     integer :: nAtom, nSpin
 
+    allocate( qq(size(qorb,1),size(qorb,2),size(qorb,3)) )
```
bhourahine, line 536: Lines 539 or 541 already does this through automatic allocation (as there is no dimensioning on the left), or are you getting problems with a specific compiler or need it on the heap?

src/dftbp/dftb/extcharges.F90
```diff
@@ -25,7 +25,7 @@ module dftbp_dftb_extcharges
     private
 
     !> Number of point charges
-    integer :: nChrg
+    integer, public :: nChrg
```
bhourahine, line 28: It is safer to return this value via a function instead of making it a public variable (otherwise it can be changed outside the module). The function should be a procedure for the type, so caries the same allocation status etc. as the instance of the type.

bhourahine (commented):
Thank you for the bug fix, please see comments about some changes and possible reversion of

bhourahine:
I've just made a PR into your branch (https://github.com/tjgiese/dftbplus/pull/1) which fixes another bug  I discovered while looking setting up tests for your changes. Could you please merge this, so it will incorporate here.

src/dftbp/dftb/extcharges.F90
tjgiese: I made the requested changes. The allocation was originally added because there was a segmentation fault and valgrind identified the stack allocation as the culprit, but I think that report was merely an artifact of having written pass the end of the arrays elsewhere. Now that the issue elsewhere has been resolved, the explicit allocation no longer appears to be necessary.

src/dftbp/dftb/extcharges.F90
```diff
@@ -316,5 +319,18 @@ subroutine copyInvRvec(this, qInvR)
 
   end subroutine copyInvRvec
 
+  
+  !> Returns the number of external charges
+  function getNumCharges(this) result(n)
```
bhourahine, line 324:
```suggestion
  pure function getNumCharges(this) result(n)
```

src/dftbp/dftb/populations.F90
```diff
@@ -533,6 +533,7 @@ subroutine getChargePerShell(qOrb, orb, species, chargePerShell, qRef)
     integer :: iAt, iSp, iSh
     integer :: nAtom, nSpin
 
+
```
bhourahine, line 536:
```suggestion
```
As otherwise this file has an update of extra blank space.

src/dftbp/dftb/scc.F90
```diff
@@ -488,6 +488,12 @@ subroutine setExternalCharges(this, chargeCoords, chargeQs, blurWidths)
     !> Spatial extension of external charge distribution
     real(dp), intent(in), optional :: blurWidths(:)
 
+    if ( allocated(this%extCharges) ) then
+       if ( this%extCharges%getNumCharges() /= size(chargeQs) ) then
+          deallocate(this%extCharges)
+       end if
+    end if
+
```
bhourahine, line 496:
```suggestion

```
Trivial white space removal.

bhourahine (approved):
Just a few trivial suggestions. If you are OK with this, I will merge them, them squash all of the changes for this branch into one commit.

bhourahine:
The gcc-mac-os regression fail is a known problem and not due to this branch. It non-deterministically fails due to a dependency problem in the cmake for that OS. I'll re-run it once the other cases complete.

src/dftbp/dftb/scc.F90
tjgiese: Sure. Sounds good.

bhourahine:
Now squashed and merged to main as commit c6ce461

PR #1311, opened 2023-09-15
Slight revision of modes codebase

vanderhe:
A suggestion for slightly revising the MODES codebase.

PR #1316, opened 2023-09-21
Modes eigenvector output bugfix

vanderhe:
At the moment the fields `eigenmodes` and `eigenmodes_scaled` in _vibrations.tag_ are always identical.

No need to recompute regression test, as there are none that cover this functionality. I will create another PR that decouples the calculation (and writing) of eigenvectors from animating selected modes. This PR will then further contain regression tests, at least that would be my suggestion.

To be merged after #1311.

### DIFF
--- a/app/modes/modes.F90
+++ b/app/modes/modes.F90
@@ -31,7 +31,7 @@ program modes
 
   integer :: ii, jj, kk, ll, iMode, iAt, iAtMoved, nAtom, nTrans
   integer :: iCount, jCount
-  real(dp), allocatable :: eigenValues(:), displ(:,:,:)
+  real(dp), allocatable :: eigenValues(:), eigenModesScaled(:,:), displ(:,:,:)
   real(dp), allocatable :: transDip(:), degenTransDip(:), transPol(:), degenTransPol(:)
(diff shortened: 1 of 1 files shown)


vanderhe:
@bhourahine The bug fix is now living in this PR.

bhourahine (commented):
Please update on main (suspect you'll get conflicts on comment changes from #1311 )

vanderhe:
> Please update on main (suspect you'll get conflicts on comment changes from #1311 )

@bhourahine Done, ready for review.

bhourahine (approved):
Looks fine as is. But is it worth adding a test case to get the coverage up for this? This might need a little work, as the eigenvector phases aren't portable between lapack implementations, so probably should be a separate PR.

vanderhe:
> Looks fine as is. But is it worth adding a test case to get the coverage up for this? This might need a little work, as the eigenvector phases aren't portable between lapack implementations, so probably should be a separate PR.

I'm already working on it (in a separate branch that also decouples eigenvector printing from mode animation), thanks for the phase-reminder, I think I would have forgotten to deal with this issue.

PR #1317, opened 2023-09-25
Re-order a cmake to reduce if blocks

bhourahine:


aradi:
I am fine with the block merging. But is the line width change (100 -> 80, I guess) intended?

bhourahine:
> I am fine with the block merging. But is the line width change (100 -> 80, I guess) intended?

Would you prefer the minimal change? I'd just used the line-wrap for emacs, which presumably is 80 in the CMake mode as I have it set up.

bhourahine:
@aradi now wrapped at 100 characters.

PR #1319, opened 2023-09-30
Marks external neighbour set for recomputation

bhourahine:
Also adds a variable to the main type to prevent DFTB+ updating the neighbourlist if it has been set externally.


bhourahine:
>  am fine with your changes, but propose to remove the `neigh%setExternally`.

We should probably discuss at the next devel meeting. Both solutions seem not to be fully blocking adjustment of the neighbour list and image coordinates after they are externally set.

aradi (approved):
I think, this is OK so. It should be probably rebased to current main before merging.

PR #1321, opened 2023-09-30
Check neighbour set

bhourahine:
Tests the setNeighbour data when a debug build. Existing testers for this function provide a non-compliant neighbour list and associated data.


bhourahine:
Is on top of #1319

bhourahine:
We don't want to forget about this.

PR #1322, opened 2023-09-30
Add ability to set xticks and xticklabels in plotxy

jubich:
Another PR will follow that will allow to read the command line arguments from a file, making this change even more useful. The idea behind this PR is to add the possibility to recreate the plots in the recipes with plotxy and thus reduce the dependency on third-party tools for plotting.

### DIFF
--- a/tools/misc/plotxy
+++ b/tools/misc/plotxy
@@ -11,6 +11,16 @@ import argparse
 import numpy as np
 import matplotlib.pylab as plt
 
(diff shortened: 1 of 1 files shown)


bhourahine (approved):
Looks ready to merge. When you make the next pull request for the command line, please reference back to this on in the comments.
There is a problem with the CI system today, so the fails are not important.

PR #1323, opened 2023-10-02
Require Magma only if WITH_MAGMA in GPU build

terminationshock:
When DFTB+ is built `WITH_GPU`, it enforces to have a custom Magma, although the user uses ELPA. From my point of view, it should only be enforced if `WITH_MAGMA` has been set.

### DIFF
--- a/utils/export/dftbplus-config.cmake.in
+++ b/utils/export/dftbplus-config.cmake.in
@@ -49,7 +49,7 @@ if(NOT TARGET DftbPlus::DftbPlus)
     find_dependency(elsi)
   endif()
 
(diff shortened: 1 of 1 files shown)


bhourahine (approved):
Agreed.
We should also discuss getting a testing pipeline for ELPA with GPU enabled.

aradi:
Note, I think the solution is not 100%-ly correct, as `DftbPlus_WITH_MAGMA` is not defined in the export file (must be created explicitly, there is now automatic way to transfer internal config options into the export file). I've have fixed that in #1326.

PR #1327, opened 2023-10-06
Automatic and manual alignment options for dp_dos

vanderhe:
Adds automatic and manual alignment options to the dp_dos script of the dptools, following the same principle as #1293.

Additionally, moves the check for incompatible optional arguments in dp_bands to the parser routine, as I have found this to be more intuitive having looked at it a second time.


tools/dptools/src/dptools/scripts/dp_dos.py
```diff
@@ -71,7 +71,8 @@ def parse_arguments(cmdlineargs=None):
         cmdlineargs: List of command line arguments. When None, arguments in
             sys.argv are parsed (Default: None).
     '''
-    parser = argparse.ArgumentParser(description=USAGE)
+    parser = ArgumentParser(
+        description=USAGE, formatter_class=RawTextHelpFormatter)
```
aradi, line 75: Is there any reason to use the `RawTextHelpFormatter` class? We would then have to format each option ourself, it won't be adaptive to the terminal width, etc.

aradi (commented):
Thx, see my comment.

Please also rebase to current main, to prevent false CI-alerts.

tools/dptools/src/dptools/scripts/dp_dos.py
vanderhe: I used the `RawTextHelpFormatter` class to format the `--auto-align` option, so that there is a visual distinction between the integer and fractional occupation case.

vanderhe: I've undone the switch to `RawTextHelpFormatter` (also for `dp_bands`, which has already been merged). Ready to go from my side.

PR #1330, opened 2023-10-06
Allow plotxy to read args from file

jubich:
This is the follow-up PR mentioned in #1322.


aradi (commented):
Fine, but please add documentation to the class and its public method.

jubich:
@aradi Documentation added

bhourahine (approved):
Ready to merge

PR #1332, opened 2023-10-09
Fixes the incompatibility problem for range separated input

### LINKED ISSUE #1331: Range separated input can not handle None {}
**Describe the bug**
Setting
```
RangeSeparated = None {}
```
in the `Hamiltonian` block triggers

* an error message about non-rs SK-files, when used with conventional SK-files
* an error message about unknown rs-functional "none", when used with rs-SK-files (which should not be done).

**To Reproduce**

Take any example and set `RangeSeparated = None {}`.

**Expected behaviour**

Code should accept it.

**Additional context**

This worked in 22.2 and was broken in 23.1, when the check about the SK-files was introduced. Unfortunately, as 22.2 created `RangeSeparated = None {}` in the input, this prevents users to use `dftb_pin.hsd` of earlier runs as input.

aradi:
Closes #1331 

However,  #1223 might fix this (as the input was changed there), so do not merge it yet.

### DIFF
--- a/src/dftbp/dftbplus/parser.F90
+++ b/src/dftbp/dftbplus/parser.F90
@@ -7782,11 +7782,11 @@ contains
     !! File name of representative SK-file to read
     character(lc) :: fileName
 
-    !! Range-separated extra tag in SK-files, if allocated
-    integer :: rangeSepSkTag
-
     !! Range-separated functional type of user input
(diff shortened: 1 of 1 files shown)


bhourahine:
@aradi Is it worth merging for the 23.2 release then reverting for the hybrids merge prior to 24.1?

PR #1334, opened 2023-10-17
Commenting and unused variables in timeprop

bhourahine:


vanderhe:
@bhourahine Shall we use this opportunity to additionally run my script to standardize the capitalization of comments?

bhourahine:
Might as well. Did you check the script in somewhere? In utils/srcmanip would make sense.

vanderhe:
> Might as well. Did you check the script in somewhere? In utils/srcmanip would make sense.

Not yet, it really isn't that robust. I can create a PR against your branch this evening and also share the script with you so that we can think about whether it is good enough for utils/srcmanip or not.

bhourahine:
Sounds good. I'd suggest making the script a separate PR itself. Then it can get iteratively improved.

fbonafe:
Thanks for the changes Ben! This is really appreciated. At the moment we are touching several subroutines with @charlymedrano7 and @mberdakin for future PR, we'll ask you for help later if we have some issues with the rebasing, hopefully not many! But this is long overdue.

charlymedrano7:
Thanks a lot Ben! We were discussing yesterday with @fbonafe about how would be the best way to clean the code in the future and now we see this PR. I can already feel the pain of the rebase of our current branches in the future PRs. Out of joke, thanks a lot for this work!

PR #1217, opened 2023-04-04
Added API functions for ASI API implemenation. Including import and e…

PavelStishenko:
This PR is a rebased version of the PR #1093 .
It includes all the work done to enable ASI API implementation. Mainly C API extension to include export and import of the Hamiltonian, Overlap, and Density matrices.

PR #1335, opened 2023-10-22
ASIrebase

bhourahine:
Contains changes from #1217 rebased on top of #1223

Needs
- [] Blocking of non-functioning combinations when ASI enabled
- [] Integration of k over suplied matrices to provide overlap/H elsewhere in the code


bhourahine:
@PavelStishenko this is now rebased on the current main (only two conflicts, so not so bad this time). Note that until there is a signaling mechanism for whether the overlap has been modified via ASI, it is unsafe to run SCC calculations.

bhourahine:
Superseded by changes in #1485

PR #1337, opened 2023-10-26
Fix block hermitian code and clean up

bhourahine:
* Moves the dense matrix symmetrization routines to a distinct module
* Cleans out un-used routines and redundant routines
* Rationalises routine names to make it clear that a triangle is copied
* Adds unit tests for serial case
* Adds BLACS matrix cases

### DIFF
--- a/src/dftbp/dftb/hybridxc.F90
+++ b/src/dftbp/dftb/hybridxc.F90
@@ -22,7 +22,7 @@ module dftbp_dftb_hybridxc
   use dftbp_type_commontypes, only : TOrbitals, TParallelKS
   use dftbp_dftb_periodic, only : TSymNeighbourList, getCellTranslations, cart2frac
   use dftbp_dftb_nonscc, only : buildS
-  use dftbp_math_matrixops, only : triangleCopySquareMatrix
+  use dftbp_math_matrixops, only : adjointLowerTriangle
   use dftbp_math_simplealgebra, only : determinant33
   use dftbp_common_parallel, only : getStartAndEndIndex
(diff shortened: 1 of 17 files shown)


bhourahine:
Temporarily switch to draft, as want to add unit tests and simplify one routine name.

aradi (commented):
I like the idea a lot to clean up the ambiguous naming we have. I have two comments on naming:

- We so far tried to use module names, which each member in the module path is not longer than 15 chars. `matrixoperations` is just one character too long. (What about `matrixops` instead?)
- It is not really a triangle copy in the comlex case. What about calling the interface `adjungateLowerTriangle()` or `transjugateLowerTriangle()`, which would also document, which triangle is the source, and would be valid for both, real and complex cases.

bhourahine:
>     * It is not really a triangle copy in the comlex case. What about calling the interface `adjungateLowerTriangle()` or `transjugateLowerTriangle()`, which would also document, which triangle is the source, and would be valid for both, real and complex cases.
https://en.wikipedia.org/wiki/Adjugate_matrix implies that isn't the right terminology, but transjugate makes sense.
How about adjointLowerTriangle() ? This is slightly shorter.

bhourahine:
@aradi Depending on what you think, I can squash the last 2 commits in. The first renames the module as suggested, the second changes the routine name using adjoint.

aradi (approved):

fbonafe:
@bhourahine it seems the adjointLowerTriangle_BLACS that you added is not tested or used anywhere else in the code, can you give me an example of how to call it? It seems now we need it, maybe it's not necessary to have a wrapper but we definitely need to know how to call it.

PR #1339, opened 2023-10-27
Remove stdOut from global variables

jubich:
- [ ] **Solve "temporary fix"** (@aradi)
- [x] Fix tests (test/src/dftbp/unit/type/typegeometryhsd.F90 & test/src/dftbp/unit/dftb/periodic.F90)
- [x] Fix regression tests
- [x] Change error (dftbp_io_message) to accept output parameter (_will be changed when error propagation is reworked/implemented_) 
- [x] Find solution for mbdPrinter (src/dftbp/dftb/dispmbd.F90) (_temporary solution_)
- [x] Rebase to main
- [ ] Squash commits

### DIFF
--- a/app/dftb+/dftbplus.F90
+++ b/app/dftb+/dftbplus.F90
@@ -27,7 +27,7 @@ program dftbplus
   call TEnvironment_init(env)
   ! temporary fix
   env%stdOut = stdOut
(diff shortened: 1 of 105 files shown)


aradi:
Should be not closed yet, we should try to merge it in second half of August.

PR #1340, opened 2023-11-04
Change to broadcast in perturbation

bhourahine:
No reason to sum notionally identical numbers, better to broadcast the value. Also clean up module ordering for routines.


vanderhe (approved):
LGTM, besides, all changes seem to be covered by tests.

PR #1341, opened 2023-11-06
Change to broadcast of variable in main

bhourahine:
Instead of averaging the charges, just broadcast the lead rank's copy.
Same idea as #1340


vanderhe (approved):
LGTM, we should probably wait for #1342 to get merged and rebase to ensure that the Intel workflow passes.

PR #1345, opened 2023-11-09
Only set OMP env for MPI test on enabled build

bhourahine:
As this was triggering an internal error for the eigensolver for Vsi+O-plumed with some compilers.


vanderhe (approved):
dftb+_md/Vsi+O-plumed now passes nag-7.1_mpi1_debug1_p1_t1.

PR #1349, opened 2023-11-14
Capitalize FORD tag comments in main.F90

vanderhe:
Ran [capitalize_ford_comments](https://github.com/dftbplus/dftbplus/blob/main/utils/srcmanip/capitalize_ford_comments) on main.F90 with two exceptions:
- k-points instead of K-points
- libNEGF instead of LibNEGF


aradi:
@vanderhe  Let's rebase it and then merge it.

vanderhe:
@aradi done

PR #1355, opened 2023-11-24
Switch to ifx compiler in CI workflow

vanderhe:
In principle we could also just fix the workflow and keep `ifort` in place, but we might as well get some experience with `ifx` now, which seems to do just fine with the limited number of extensions enabled in the GitHub workflows.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -187,7 +187,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        fc: [ifort]
+        fc: [ifx]
         cc: [icx]
     env:
       FC: ${{ matrix.fc }}
@@ -195,7 +195,7 @@ jobs:
       WITH_MPI: false
       APT_PACKAGES: >-
         intel-oneapi-compiler-fortran
-        intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic
+        intel-oneapi-compiler-dpcpp-cpp
         intel-oneapi-mkl
         intel-oneapi-mkl-devel
       CMAKE_OPTIONS: >-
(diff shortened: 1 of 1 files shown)


vanderhe:
I guess we should merge PR #1214 to avoid `dftb+_helical/10-0CtubeC_10_sampled` failing.

.github/workflows/build.yml
```diff
@@ -187,17 +187,15 @@ jobs:
...
       FC: ${{ matrix.fc }}
       CC: ${{ matrix.cc }}
       WITH_MPI: false
       APT_PACKAGES: >-
-        intel-oneapi-compiler-fortran
-        intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic
-        intel-oneapi-mkl
-        intel-oneapi-mkl-devel
+        intel-basekit
+        intel-hpckit
```
aradi, line 198: That is a real overkill, the basekit contains tons of unnecessary packages.

Also, I've just installed `intel-oneapi-compiler-fortran` on my laptop, and it does install `ifort`. @vanderhe Could you first try just to update the repository keys? Probably that is already enough to fix the issue?

vanderhe: Will try, shall we move on to `ifx` then? I would be interested to see how it holds up, but sticking with `ifort` is the safe path, up to you.

vanderhe: @aradi Current iteration does not find `ifort` again, am I doing something wrong?

vanderhe:
@aradi Ready to go from my side, if we are willing to ignore the commensurate k-test fail for now.

aradi:
@vanderhe Thx, I am fine with it. Just to be on the safe side, could you rebase and check, whether the ifx test passes now? (I've merged #1214 to main)

bhourahine:
@vanderhe Can you rebase on main, now the helical fix is included, and we'll see if everything passes.

bhourahine:
@vanderhe Any reason that the coverage tests didn't run?

PR #1357, opened 2023-11-27
Fix build dependency probelm in api/mm/testers

aradi:
Seems to fix the problem of failing builds in the CI due to incorrect build order with Ninja backend.

Probably, this should be merged before other PRs; after rebasing the other PRs on this, the GitHub CI triggered in the pull requests should became meaningful again...


vanderhe (approved):
I would not consider myself a CMake expert, but looks fine to me.

aradi:
None of us really are, but the CI seems to be happy with the changes. :wink:

bhourahine (commented):
Seems to be a bit of a hack, but works. Do you think it's worth commenting this as a dependency work-around for CMake?

PR #1358, opened 2023-12-03
Removed helical case that macos finds problematic

bhourahine:
This removes the high-order C_n helical rotation test (which fails for the mac-OS CI pipeline for unknown reasons).


vanderhe:
@bhourahine Should we really remove a regtest just because the MacOS pipeline is broken? Or is this particular test redundant anyway?

bhourahine:
@vanderhe the problem is there is no way to debug the pipeline in the
absence of detailed output and stderr. This is a new test from the branch
merged a couple of days ago, and passes everywhere else.

bhourahine:
Probably superseded by #1359

PR #1359, opened 2023-12-05
Added missing parallel block in test input

bhourahine:
Possible fix for the helical test.

If it passes, this superceeds #1358


vanderhe:
> @vanderhe @aradi Tighter force tolerance seems to fix the C_1000 helical case, just waiting on buildbot tests to complete before ready to merge this.

@bhourahine Sounds good, fingers crossed.

PR #1360, opened 2023-12-10
Remove unused gbs eigensolver

bhourahine:
Drops the unused banded LAPACK eigensolvers.


vanderhe:
Was there any feature in the past that made use of the banded eigensolvers? Not relevant for merging this, I'm just curious...

bhourahine:
@vanderhe Not really. There was a really ancient branch that reorganized the atom numbering to make the H&S matrices more band-like, which reduced the memory use substantially for eigenvalues. The problem is that as soon as you need eigenvectors, the full n*n dense matrix is needed to store them and the memory costs explode.

bhourahine:
Just rebased to current main.

vanderhe (approved):
LGTM

PR #1364, opened 2023-12-16
Formatting for variable comments about k-points

bhourahine:
Changes comments starting "!> k-point" or "!> K-point" to "!> The k-point", avoiding caps for k, but also starting the comment in accordance with the convention.


vanderhe (approved):
Perfect, thanks for the follow-up PR.

vanderhe:
@bhourahine I guess at this point we should add this convention to the devel. style guide, I think it is not yet included there.

PR #1366, opened 2024-01-04
Facilitate vectorization of ewaldReciprocal

terminationshock:
These changes help vectorizing the loop in the function `ewaldReciprocal`. In my tests, its runtime is reduced to 33% of its previous runtime. Note that the dot products have to be written explicitly. Otherwise, the compiler will only vectorize the innermost loops (with 3 elements) instead of the outer large loop.

### DIFF
--- a/src/dftbp/dftb/coulomb.F90
+++ b/src/dftbp/dftb/coulomb.F90
@@ -2050,19 +2050,24 @@ contains
     !> contribution to the sum
     real(dp) :: recSum
 
-    real(dp) :: gg(3), g2
+    real(dp), allocatable :: recSumArray(:)
+
+    real(dp) :: g2Factor, g2, gr
     integer :: iG
 
     @:ASSERT(vol > 0.0_dp)
 
-    recSum = 0.0_dp
+    g2Factor = -1.0_dp / (4.0_dp*alpha**2)
+
+    allocate(recSumArray(size(gVec, dim=2)), source=0.0_dp)
+
     do iG = 1, size(gVec, dim=2)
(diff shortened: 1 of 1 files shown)


bhourahine:
It's probably not going to be a major memory requirement, checking over the test set in serial, the largest example of the new array is 9762 elements. This is for the hybrid/periodic/Acene-gamma-matrix-restart test case which is a fairly small unit cell.
Could you at some stage also have a look at also vectorising the equivalent sums in the derivEwaldReciprocal and derivStressEwaldRec routines? For XLBOMD or Ehrenfest calculations these could also show similar impacts on performance.

bhourahine (approved):
LGTM

terminationshock:
> Could you at some stage also have a look at also vectorising the equivalent sums in the derivEwaldReciprocal and derivStressEwaldRec routines?

Yes, sure, will do that.

terminationshock:
@bhourahine At the moment, I do not dare to touch `derivStressEwaldRec`, because there seems to be no test touching that routine. I deliberately put an error there and all tests were successful.
Therefore, I just changed `derivEwaldReciprocal`.

bhourahine:
> @bhourahine At the moment, I do not dare to touch `derivStressEwaldRec`, because there seems to be no test touching that routine. I deliberately put an error there and all tests were successful. Therefore, I just changed `derivEwaldReciprocal`.

That's strange, as there are tests producing stress tensors for cases that should be using Ewald summation. I'll merge your changes and then investigate. How large/what was the error?

PR #1367, opened 2024-01-13
Possible bug when deltaRhoOut generated

bhourahine:
As the work array is then not used, so nothing to copy. Would impact excited state use of resulting density matrix.

### DIFF
--- a/src/dftbp/dftbplus/main.F90
+++ b/src/dftbp/dftbplus/main.F90
@@ -3660,8 +3660,13 @@ contains
     #:endif
 
       if (allocated(rhoSqrReal)) then
-        rhoSqrReal(:,:, iSpin) = work
+        if (.not. allocated(deltaRhoOut)) then
+          rhoSqrReal(:,:, iSpin) = work
+        else
+          rhoSqrReal(:,:, iSpin) = deltaRhoOut(:,:,iSpin)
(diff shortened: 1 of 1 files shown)


vanderhe (approved):
Your fix seems fine, although I'm not sure if we should simultaneously store $\mathrm{P}(\mathbf{k})$ and $\Delta\mathrm{P}(\mathbf{k})$ at all. At this point they are of course still identical, but I don't see a reason why we could not store e.g. $\Delta\mathrm{P}(\mathbf{k})$ only and calculate $\mathrm{P}(\mathbf{k})$ on the fly in the linresp routines.

bhourahine:
Agreed, but for the moment lets just keep things working and then refactor later.

vanderhe:
@bhourahine As far as I can see this bug only shows up for colinear spin-pol. + excited gradients, but there is no such regression test, e.g. `Benzene-TD-LC-Spinpol` and `Benzene-TD-LC-Spinpol-Iter` do not test the gradients (not included in _autotest.tag)

bhourahine:
@vanderhe suggests we need a regression for this, either by extending those tests or adding a new case. The Casida routines only handle $\Gamma$-point and then only for excitation energies at the moment.

PR #1370, opened 2024-01-17
Spin-polarized, long-range corrected, LR gradients fix

vanderhe:
Fixes spin-polarized, long-range corrected, linear-response gradient evaluation and adds a regression test.


vanderhe:
@bhourahine ping

vanderhe:
@bhourahine NACV path not yet tested, I have to familiarize myself with that feature first. Could you run this branch on your local tests too, to double check?

aradi:
Wait a sec, I am slightly confused. Isn't this just reverting an earlier fix (which according to the comment had to be introduces, as `rhoSqr` did not contain the correct values?) Or am I missing something?

vanderhe:
> Wait a sec, I am slightly confused. Isn't this just reverting an earlier fix (which according to the comment had to be introduces, as `rhoSqr` did not contain the correct values?) Or am I missing something?

This is what the comment says, yes. But as a matter of fact the above version agrees with numeric forces to about 1e-09a.u. and the current version (which just ignores the down-channel) seems to be wrong.

bhourahine:
@aradi @vanderhe I'd read this a a comment requesting a change be made:
!! TO CHANGE: For tHybridXc density from call seems to be incorrect, have
and suspect it got squashed in from some suggestion in code review.

PR #1372, opened 2024-01-19
Enable spin-polarized general k-point hybrid path

vanderhe:
Enables spin-polarized hybrid DFTB ground state calculations within the neighbor-based general k-point algorithm.

In #1223 I implemented all equations with colinear spin-polarization in mind, however this path has been blocked until now due to the lack of testing and missing regression tests. The regtest added here is verified against the cluster case [test/app/dftb+/hybrid/cluster/Benzene-Spinpol-matrix](https://github.com/dftbplus/dftbplus/tree/main/test/app/dftb%2B/hybrid/cluster/Benzene-Spinpol-matrix).

Be aware that e.g. the total energy of charged systems (as present here) does not agree with the non-periodic case for medium sized boxes and slowly converges to the cluster case for huge box sizes. However, I first verified that the (shifted/aligned) eigenspectrum produced by the neighborlist-based Gamma-only algorithm agrees with the non-periodic matrix-based algorithm. As a second step I then verified the agreement of the neighbor-based Gamma-only  results with the general k-point neighborlist-based code path.

(The changes further include some more or less unrelated stuff, sorry for that.)

Still missing/blocked:
- spin-polarized gradient evaluation within the general k-point hybrid formalism
(in principle implemented but not yet tested)
- spin-polarized general k-point hybrid path for Coulomb matrix approaches other than 'Truncated'
(I still have to test e.g. the spin-polarized minimal image convention based routines)


bhourahine (approved):
LGTM. Looks like the vast majority, apart from the test cases, is comments and changes to error messages.

PR #1373, opened 2024-01-19
Update CHANGELOG

vanderhe:
Follow-up to #1367 and #1370.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -45,6 +45,9 @@ Fixed
 - Calls to setExternalCharges were failing if number of external charges
   changes.
 
+- Incorrect excited gradients for spin-polarized long-range corrected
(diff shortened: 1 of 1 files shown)


CHANGELOG.rst
```diff
@@ -45,6 +45,9 @@ Fixed
 - Calls to setExternalCharges were failing if number of external charges
   changes.
 
+- Incorrect excited gradients for spin-polarized, long-range corrected
```
bhourahine, line 48:
```suggestion
- Incorrect excited gradients for spin-polarized long-range corrected
```
As could be read that spin-polarised are generally affected, instead of specific to LR

vanderhe: @bhourahine applied

bhourahine (approved):
LGTM

PR #1374, opened 2024-01-19
Fixes problem in parser for bulk BC read

bhourahine:


bhourahine (commented):
Agreed as correct at today's devl. meeting.

PR #1375, opened 2024-01-23
Regression test for Ewald stress tensor

bhourahine:
Geometry is not relaxed, so tensor has finite values for all elements.


bhourahine:
@terminationshock hopefully this should be sensitive to the your test changes in the Ewald stress, so can be a valid regression test for refactoring that routine.
While looking at this, I was reminded that the reciprocal Ewald for 1/r and 1/r2 do not use the  distributeRangeInChunks() mechanism. Do you think that would also help with the timings for the Ewald use case that started this discussion of vectorising?

terminationshock:
Your new test is not sensitive to changes in `derivStressEwaldRec`. My diff is:

```
diff --git a/src/dftbp/dftb/coulomb.F90 b/src/dftbp/dftb/coulomb.F90
index d864717b..152292ba 100644
--- a/src/dftbp/dftb/coulomb.F90
+++ b/src/dftbp/dftb/coulomb.F90
@@ -2158,8 +2158,8 @@ contains
       sigma(:,:) = sigma + (-unity + sTmp * spread(gg, 1, 3)*spread(gg, 2, 3)) * cos(rg) * eTerm
     end do
     ! note factor of 2 as only summing over half of reciprocal space
-    recSum(:) = 2.0_dp * recSum * 4.0_dp * pi / vol
-    sigma(:,:) = 2.0_dp * sigma * 4.0_dp * pi / vol
+    recSum(:) = 1.0_dp * recSum * 4.0_dp * pi / vol
+    sigma(:,:) = 1.0_dp * sigma * 4.0_dp * pi / vol
 
   end subroutine derivStressEwaldRec
```

The test is still successful, though:
```
Test #14: dftb+_scc/GaAs_8 ....................   Passed
```

PR #1376, opened 2024-01-25
Clarify error message a bit for users

bhourahine:
Giving more detail in an error message, Just in case users don't know what causes complex hamiltonians.


vanderhe (approved):
LGTM

PR #1382, opened 2024-02-01
Remove imports without "only"

jubich:
I have noticed that the following import is not reached even if it is compiled with the "WITH_MPI" flag.

https://github.com/dftbplus/dftbplus/blob/13588bb08d666d3beadc579e345dec6683fa9145/app/misc/skderivs/skderivs.F90#L24-L26

### DIFF
--- a/app/misc/skderivs/skderivs.F90
+++ b/app/misc/skderivs/skderivs.F90
@@ -9,22 +9,23 @@
 
 !> Calculates the first and second derivatives of matrix elements
 program skderivs
-  use dftbp_common_accuracy
-  use dftbp_common_constants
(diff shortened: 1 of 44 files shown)


aradi (changes requested):
Please, make sure, that you align with our style guide everywhere. I see several continuation lines with incorrect indentation (should be 4 characters). Also, you can use line lengths up to 100 characters.

aradi (approved):

aradi:
Perfect, could you please re-base it to current main, and then we can merge it. (Sorry for taking so long...)

jubich:
.

PR #1384, opened 2024-02-03
Fix for junk in driver convergence block

bhourahine:
Non-parsed keywords inside block were silently ignored and pass through to dftb_pin.hsd even when IgnoreUnprocessedNodes = Yes

Example case:
Driver = GeometryOptimization {
  Convergence {Junk = Yes}
}

Note: not clear why the older construction using getChildValue was doing this.


awvwgk (approved):
Looks good to me

PR #1385, opened 2024-02-03
Yearupdate

jubich:
- [ ] Discuss new files (`utils/srcmanip/change_header_year, utils/srcmanip/set_new_source_headers`)
- [ ] Squash into one commit


bhourahine:
@jubich How does the change_header_year interact with set_new_source_headers? It seems like the former deals wit h a wider set of files, but does not process other changes to F90 headers.
The pylint score for the newer file isn't great:
pylint --rcfile=../srccheck/pylint/pylintrc-3.ini ./change_header_year

jubich:
The idea behind `change_header_year` is to change the year in any file whose header matches the template, while the second file `set_new_source_headers` is used to create a header in imported files (`.f90, .sh, .py, ...`) that do not yet have a header. 
I have decoupled the function of the original script (creating AND changing the year) into creating OR changing the year. 
This offers the advantage of finer control.

As for the pylint score, I have not checked it. :sweat_smile:

jubich:
Both files are now formatted with _black_. I have also added `encoding="utf-8"` to `with open()`. Now `change_header_year` has a pylint score of 9.52 and `set_new_source_headers` has a score of 9.30.

utils/srcmanip/change_header_year
```diff
@@ -0,0 +1,103 @@
...
+        with open(file, "r", encoding="utf-8") as fp:
+            newcontent = ""
+            for line in fp.readlines():
+                hit = PATTERN.search(line)
+                if hit:
+                    changed_date = True
+                    start, stop = hit.span("year")
+                    newcontent += line[:start] + YEAR + line[stop:]
+                else:
+                    newcontent += line
```
aradi, line 68: Is there any reason, why we have to process the file line by line (as instead of reading in one go and make a search and replace)?

jubich:
Should I squash these commits now or is there any further feedback?

jubich:
.

PR #1390, opened 2024-02-07
MPI implementation of timeprop 

mberdakin:
Time propagation using BLACS only for finite systems (REAL H ONLY)


bhourahine:
@mberdakin could you look at resolving the conflicts against the current main branch. Possibly it is best to do this by rebasing on main interactively and fixing conflicts.
There also seem to be some checked in temporary files (Testing/Temporary/*).
Eventually the commits should be squashed to a single commit.

src/dftbp/timedep/timeprop.F90
```diff
@@ -1872,16 +2047,28 @@ subroutine initializeTDVariables(this, rho, H1, Ssqr, Sinv, H0, ham0, Dsqr, Qsqr
...
         end if
       end do
-      write(stdOut,"(A)")'S inverted'
 
+    #:endif    
+
+    write(stdOut,"(A)")'S inverted'
+
+
+      !!!
```
bhourahine, line 2056: ?

src/dftbp/timedep/timeprop.F90
```diff
@@ -1872,16 +2047,28 @@ subroutine initializeTDVariables(this, rho, H1, Ssqr, Sinv, H0, ham0, Dsqr, Qsqr
...
+
+    write(stdOut,"(A)")'S inverted'
+
+
+      !!! 
       do iKS = 1, this%parallelKS%nLocalKS
         iK = this%parallelKS%localKS(1, iKS)
         iSpin = this%parallelKS%localKS(2, iKS)
         if (this%tRealHS) then
+          !!!
```
bhourahine, line 2061: ?

src/dftbp/timedep/timeprop.F90
```diff
@@ -1924,17 +2112,25 @@ subroutine initializeTDVariables(this, rho, H1, Ssqr, Sinv, H0, ham0, Dsqr, Qsqr
...
+
         else
           call makeDensityMatrix(rho(:,:,iKS), eigvecsCplx(:,:,iKS), filling(:,iK,iSpin))
         end if
-        do iOrb = 1, this%nOrbs-1
-          do iOrb2 = iOrb+1, this%nOrbs
-            rho(iOrb, iOrb2, iKS) = conjg(rho(iOrb2, iOrb, iKS))
-          end do
-        end do
+        ! do iOrb = 1, this%nOrbs-1
```
bhourahine, line 2129: ?

src/dftbp/timedep/timeprop.F90
```diff
@@ -4193,22 +4557,24 @@ subroutine finalizeDynamics(this)
...
     deallocate(this%RdotSprime)
     deallocate(this%qq)
     deallocate(this%deltaQ)
     deallocate(this%dipole)
     deallocate(this%chargePerShell)
     deallocate(this%occ)
-    deallocate(this%totalForce)
-    deallocate(this%trho)
-    deallocate(this%trhoOld)
+    deallocate(this%totalForce)
```
bhourahine, line 4575:
```suggestion
    deallocate(this%totalForce)
```

src/dftbp/timedep/timeprop.F90
```diff
@@ -4193,22 +4557,24 @@ subroutine finalizeDynamics(this)
 
     !> ElecDynamics instance
     type(TElecDynamics), intent(inout) :: this
+    !WARNING : Matias comment these lines:
```
bhourahine, line 4560: ?

sys/gnu.cmake
```diff
@@ -80,8 +80,8 @@ set(C_FLAGS_COVERAGE "-O0 -g --coverage")
 #set(ARPACK_LIBRARY_DIR "" CACHE STRING "Directories where Arpack library can be found")
 
 # ScaLAPACK -- only needed for MPI-parallel build
-#set(SCALAPACK_LIBRARY "scalapack-openmpi" CACHE STRING "Scalapack libraries to link")
-#set(SCALAPACK_LIBRARY_DIR "" CACHE STRING "Directories where Scalapack libraries can be found")
+set(SCALAPACK_LIBRARY "scalapack-openmpi" CACHE STRING "Scalapack libraries to link")
```
bhourahine, line 83: Any reason to change this? The old behaviour was to enable this to be set at the command line and is documented in INSTALL.rst

test/Testing/Temporary/CTestCostData.txt
bhourahine, line 1: Probably shouldn't be checked in

test/Testing/Temporary/LastTest.log
bhourahine, line 1: Probably shouldn't be checked in

test_fast_MPI/autotest.tag
bhourahine, line 1: Should this directory be checked in?

test_fast_TLS/OMP/autotest.tag
bhourahine, line 1: Should this directory be checked in?

test_fast_openMP/autotest.tag
bhourahine, line 1: Should this directory be checked in?

mberdakin:
Hi @bhourahine  and @baradi09 
This pull request is ready to review. 
Cheers !

fbonafe:
@baradi09 @bhourahine I could not do the rebase because I don't know how to handle the conflicts with the external libraries (it doesn't point me to a specific file, I guess there are some issues with the commits). Maybe it would be easier for you to squash the commits and then solve the few conflicts that could appear. 

Also, feel free to start reviewing, it would be nice to merge this in the upcoming weeks so we can move on to merging the Peierls branch.

Let us know!

vanderhe:
@aradi I created a squashed and rebased version of this branch in [here](https://github.com/vanderhe/dftbplus/tree/mpielectrondynamics). I decided to revert all (partly unintentional?) changes to the submodules, so scalapackfx still needs to be updated as the matrix inversion of [scalapack#37](https://github.com/dftbplus/scalapackfx/pull/37) is needed. Hope this is helpful.

vanderhe:
@aradi Separate commit added for (temporary) scalapackfx submodule update to current main.

fbonafe:
@vanderhe @mberdakin I enabled the hybrids with MPI, for which I had to write the propagation routine for complex matrices, and I enabled the tests with hybrids. Of the two butadiene tests we have, one passes (`timeprop/C4H6_rs`) and one fails (`timeprop/C4H6_rs_Singlet`). The second one uses spin polarisation and a kick instead of laser perturbation, but we have already tested these conditions without hybridxc (namely e.g. the test `timeprop/C60_triplet`). I would have to do more testing to see where the failure arises for the kick case.

src/dftbp/timedep/timeprop.F90
```diff
@@ -1545,12 +1598,23 @@ subroutine updateH(this, H1, ints, H0, speciesAll, qq, q0, coord, orb, potential
...
+      deltaRho = rho
+      if (this%nSpin > 2) then
+        @:RAISE_ERROR(errStatus, -1, "HybridXc: Not implemented for non-colinear spin.")
+      end if
+      call denseSubtractDensityOfAtomsCmplxNonperiodicBlacs(env, this%parallelKS, q0, this%denseDesc, deltaRho)
+
+      do iSpin = 1, this%nSpin
+        HSqrCplxCam(:,:) = (0.0_dp, 0.0_dp)
+        call hybridXc%addCamHamiltonianMatrix_cmplx_blacs(env, this%denseDesc, sSqr(:,:,iSpin),&
+            & deltaRho(:,:,iSpin), HSqrCplxCam)
```
fbonafe, line 1615: @vanderhe could you please check if these calls should be fine for the a colinear spin-polarised case? This is the only test that is failing with hybrids and MPI.

vanderhe: The error is buried somewhere in the `denseSubtractDensityOfAtomsCmplxNonperiodicBlacs()` routine. One of the reasons seems to be the `scale` factor, which should not be `0.5` in the spin-polarized case, as the time-propagation module seems to follow different conventions than the rest of the code base. But there is more to it, I'm still debugging...

vanderhe: See https://github.com/mberdakin/dftbplus/pull/1.

fbonafe: I don't remember if in the rest of the code it was like this, but in the real-time module the spin index of the density matrix means up-down, instead of q-m. That is why when we enter the Hamiltonian routine we run ud2qm() for the Mulliken populations.

fbonafe: > See [mberdakin#1](https://github.com/mberdakin/dftbplus/pull/1).

I hadn't seen you had the solution, thanks!

src/dftbp/timedep/timeprop.F90
```diff
@@ -1530,6 +1571,18 @@ subroutine updateH(this, H1, ints, H0, speciesAll, qq, q0, coord, orb, potential
...
     end if
 
+  #:if WITH_SCALAPACK
+    do iKS = 1, this%parallelKS%nLocalKS
+      iK = this%parallelKS%localKS(1, iKS)
+      iSpin = this%parallelKS%localKS(2, iKS)
+      if (this%tRealHS) then
+        call unpackHSRealBlacs(env%blacs, ints%hamiltonian(:,iSpin), neighbourList%iNeighbour,&
+            & nNeighbourSK, iSparseStart, img2CentCell, this%denseDesc, T2)
+        H1(:,:,iSpin) = cmplx(T2, 0.0_dp, dp)
```
aradi, line 1581:
```suggestion
        H1(:,:,iSpin) = cmplx(T2, kind=dp)
```

src/dftbp/timedep/timeprop.F90
```diff
@@ -1530,6 +1571,18 @@ subroutine updateH(this, H1, ints, H0, speciesAll, qq, q0, coord, orb, potential
...
+        ! TODO: add here the unpacking of H1 for kpoints
+      end if
+    end do
+  #:else
     do iKS = 1, this%parallelKS%nLocalKS
       iK = this%parallelKS%localKS(1, iKS)
       iSpin = this%parallelKS%localKS(2, iKS)
```
aradi, line 1588: We should leave away those lines as not needed.

src/dftbp/timedep/timeprop.F90
```diff
@@ -1638,17 +1763,21 @@ subroutine kickDM(this, rho, Ssqr, Sinv, iSquare, coord)
...
 
+#:endif
     write(stdout,"(A)")'Density kicked along ' // localDir(this%currPolDir) //'!'
 
   end subroutine kickDM
 
 
   !> Creates array for an external TD field
-  subroutine getTDFunction(this, startTime)
+  subroutine getTDFunction(this, env, startTime)
```
aradi, line 1773: This routine makes too many things IMO (it did already in the serial case, but in the MPI-one it get's really convoluted):
* Calculates or reads the TD-function
* Writes the TD-function
Note, that the outcome in the parallel case is different on different ranks:
* On lead this%tdfunction is filled up with the read value, but on the other ranks, the output is undeterministic, as far as I can see.

src/dftbp/timedep/timeprop.F90
```diff
@@ -1695,13 +1836,26 @@ subroutine getTDFunction(this, startTime)
...
 
       if (this%tEnvFromFile) then
+      #:if WITH_MPI
+        if (env%mpi%tGlobalLead) then
+          read(laserDat%unit, *)time, tdfun(1), tdfun(2), tdfun(3)
+        end if
+      #:else
         read(laserDat%unit, *)time, tdfun(1), tdfun(2), tdfun(3)
+      #:endif
         this%tdFunction(:, iStep) = tdfun * (Bohr__AA / Hartree__eV)
```
aradi, line 1846: Note: `tdfun(:)` has not been filled up on the follow ranks! Is that OK? (If yes, we should add at least a comment about this...)

fbonafe: Good catch Bálint, and thanks for the review. We'll work on the changes ASAP so we can move on.

src/dftbp/timedep/dynamicsrestart.F90
```diff
@@ -261,4 +270,178 @@ subroutine readRestartFile(rho, rhoOld, coord, veloc, time, dt, fileName, isAsci
 
   end subroutine readRestartFile
 
+
+  #:if WITH_SCALAPACK
+  !> Write to a restart file the DM in distributed format
```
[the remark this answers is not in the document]
fbonafe: There is already an error message in line 1142 of timeprop.F90 regarding this. Should we add an assert here to make sure?

https://github.com/dftbplus/dftbplus/blob/df6bf69c027b5e5a6517655860abd6d093988ee3/src/dftbp/timedep/timeprop.F90

src/dftbp/timedep/timeprop.F90
fbonafe: `iSpin` is always need and `iK` is also need in the complex case. Or there is a different way of accessing these indices?

src/dftbp/timedep/timeprop.F90
fbonafe: I see the third point you mention happening only in the the case that `tEnvFromFile` is true. In that lead knows tabout tdfun, and the others don't. But for all other conditions, tdfun is calculated on all ranks. But you are right, this function should be split at least in two cases, I'll do some refactoring now.

aradi (commented):
Thanks for fixing the issues so far. I have  a few more (rather stylistic) change requests. Additionally, please check code on dpdevel, as buildbot fails with several builders currently (e.g. `dftb_reg_devel_gnu-14`, `dftb_reg_prod_gnu-12` -- the failure of the nag-builder is due to an other bug in the code #1660)

fbonafe:
@aradi I did the changes you requested. I haven't yet seen the dp_devel, but I noticed already that the hybrid tests are failing in my local build with gcc and gfortran 10.2. I have to figure out if this is a real bug in the code or a matter of the compiler version.

fbonafe:
@aradi with @mberdakin we have disabled hybrids and tblite for the moment (will create issues to enable this in the future), now the tests are passing and all the review comments have been addressed. Feel free to merge if you consider so!

src/dftbp/dftb/populations.F90
```diff
@@ -1045,6 +1046,48 @@ subroutine denseSubtractDensityOfAtomsRealNonperiodicBlacs(env, parallelKS, q0,
...
+      do iAt = 1, nAtom
+        iOrbStart = denseDesc%iAtomStart(iAt)
+        nOrb = denseDesc%iAtomStart(iAt + 1) - iOrbStart
+        tmp(:,:) = (0.0_dp, 0.0_dp)
+        do iOrb = 1, nOrb
+          tmp(iOrb, iOrb) = -q0(iOrb, iAt, 1)
+        end do
+        call scalafx_addl2g(env%blacs%orbitalGrid, tmp(1:nOrb, 1:nOrb), denseDesc%blacsOrbSqr,&
+            & iOrbStart, iOrbStart, rho(:,:,iKS))
+      end do
```
aradi, line 1086: I have some doubts, whether this routine is correct for the spin polarized case. Did you test it for that?
I assume, the density matrix `rho(:,:,:)` is in the up-down representation when this routine is called, isn't it? On the other hand `q0` is usually in the qm-representation. So, in the spin polarized case, you would subtract the full charge in the q-channel, from both, the u- and the d- channel of `rho`. (Effectively removing twice as many electrons, as needed.)

src/dftbp/timedep/dynamicsrestart.F90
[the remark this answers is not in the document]
aradi: Yes, adding an ASSERT would be a good idea.

src/dftbp/timedep/dynamicsrestart.F90
```diff
@@ -261,4 +270,177 @@ subroutine readRestartFile(rho, rhoOld, coord, veloc, time, dt, fileName, isAsci
...
+    real(dp), intent(in) :: veloc(:,:)
+
+    !> Simulation time (in atomic units)
+    real(dp), intent(in) :: time
+
+    !> Time step being used (in atomic units)
+    real(dp), intent(in) :: dt
+
+    !> Name of the dump file
+    character(len=*), intent(in) :: fileName
```
aradi: if we take it literally, it should be called `fileNamePrefix` as the full file name is composed within the routine.
```suggestion
    !> Name prefix of the dump file (extension appended within the routine automatically)
    character(len=*), intent(in) :: fileNamePrefix
```
But then it should be edited in all read/write routines above...

src/dftbp/timedep/dynamicsrestart.F90
```diff
@@ -261,4 +270,177 @@ subroutine readRestartFile(rho, rhoOld, coord, veloc, time, dt, fileName, isAsci
...
+          call collector%getline_follow(env%blacs%orbitalGrid, icol, rho(:,:,iKS))
+        end if
+      end do
+    end do
+
+    do iKS = 1, parallelKS%nLocalKS
+      do icol = 1, nOrb
+        if (env%mpi%tGlobalLead) then
+          call collector%getline_lead(env%blacs%orbitalGrid, icol, rhoOld(:,:,iKS), localRhoCol)
+          write(fd%unit)localRhoCol(:)
```
aradi:
```suggestion
          write(fd%unit) localRhoCol
```

src/dftbp/timedep/dynamicsrestart.F90
```diff
@@ -261,4 +270,177 @@ subroutine readRestartFile(rho, rhoOld, coord, veloc, time, dt, fileName, isAsci
...
+      call openFile(fd, file=trim(fileName)//'.bin',&
+          & options=TOpenOptions(form='unformatted', access='stream', action='read',&
+          & position="rewind"))
+
+      read(fd%unit) version
+      read(fd%unit) nOrb, nSpin, nAtom, time, deltaT
+      read(fd%unit) coord, veloc
+    end if
+
+    nOrb = denseDesc%fullSize
```
aradi, line 418: I'd like to note, that this reading routine is very fragile.  If you feed it with a restart file from a system with bigger size, it will happily digest it and use the meaningless numbers as starting density. (Missing check on nOrb, nSpin, nAtom, etc. consistency...) At least we should add an appropriate note as comment.

src/dftbp/timedep/dynamicsrestart.F90
```diff
@@ -261,4 +270,177 @@ subroutine readRestartFile(rho, rhoOld, coord, veloc, time, dt, fileName, isAsci
...
+        else
+          call distributor%setline_follow(env%blacs%orbitalGrid, icol, rho(:,:,iKS))
+        end if
+      end do
+    end do
+
+    do iKS = 1, parallelKS%nLocalKS
+      do icol = 1, nOrb
+        if (env%mpi%tGlobalLead) then
+          read(fd%unit) localRhoCol(:)
```
aradi:
```suggestion
          read(fd%unit) localRhoCol
```

src/dftbp/timedep/dynamicsrestart.F90
```diff
@@ -261,4 +270,177 @@ subroutine readRestartFile(rho, rhoOld, coord, veloc, time, dt, fileName, isAsci
...
+    end if
+
+    nOrb = denseDesc%fullSize
+    allocate(localRhoCol(nOrb))
+
+    call distributor%init(env%blacs%orbitalGrid, denseDesc%blacsOrbSqr, "c")
+    do iKS = 1, parallelKS%nLocalKS
+      do icol = 1, nOrb
+        if (env%mpi%tGlobalLead) then
+          read(fd%unit) localRhoCol(:)
```
aradi:
```suggestion
          read(fd%unit) localRhoCol
```

src/dftbp/timedep/dynamicsrestart.F90
```diff
@@ -261,4 +270,177 @@ subroutine readRestartFile(rho, rhoOld, coord, veloc, time, dt, fileName, isAsci
...
+          call distributor%setline_follow(env%blacs%orbitalGrid, icol, rho(:,:,iKS))
+        end if
+      end do
+    end do
+
+    do iKS = 1, parallelKS%nLocalKS
+      do icol = 1, nOrb
+        if (env%mpi%tGlobalLead) then
+          read(fd%unit) localRhoCol(:)
+          call distributor%setline_lead(env%blacs%orbitalGrid, icol, localRhoCol(:), rhoOld(:,:,iKS))
```
aradi, line 437:
```suggestion
          call distributor%setline_lead(env%blacs%orbitalGrid, icol, localRhoCol, rhoOld(:,:,iKS))
```

src/dftbp/timedep/dynamicsrestart.F90
```diff
@@ -261,4 +270,177 @@ subroutine readRestartFile(rho, rhoOld, coord, veloc, time, dt, fileName, isAsci
...
+
+    nOrb = denseDesc%fullSize
+    allocate(localRhoCol(nOrb))
+
+    call distributor%init(env%blacs%orbitalGrid, denseDesc%blacsOrbSqr, "c")
+    do iKS = 1, parallelKS%nLocalKS
+      do icol = 1, nOrb
+        if (env%mpi%tGlobalLead) then
+          read(fd%unit) localRhoCol(:)
+          call distributor%setline_lead(env%blacs%orbitalGrid, icol, localRhoCol(:), rho(:,:,iKS))
```
aradi, line 426:
```suggestion
          call distributor%setline_lead(env%blacs%orbitalGrid, icol, localRhoCol, rho(:,:,iKS))
```

src/dftbp/timedep/timeprop.F90
```diff
@@ -807,6 +826,8 @@ subroutine TElecDynamics_init(this, inp, species, speciesName, tWriteAutotest, a
     if (allocated(tblite)) then
       this%tblite = tblite
     end if
+    ! we assume denseDesc is allocated
```
aradi:
```suggestion
```
Strip comment. If `denseDesc` was not allocated, already the line `type(TDenseDescr), intent(in) :: denseDesc` above would be invalid. (And `denseDescr` is allocated...)

src/dftbp/timedep/timeprop.F90
```diff
@@ -1470,8 +1508,19 @@ subroutine updateH(this, H1, ints, H0, speciesAll, qq, q0, coord, orb, potential
...
+  #:if WITH_SCALAPACK
+    nLocalRows = size(H1, dim=1)
+    nLocalCols = size(H1, dim=2)
+  #:else
+    nLocalRows = this%nOrbs
+    nLocalCols = this%nOrbs
+  #:endif
+
+    if (this%tRealHS) then
+      allocate(T2(nLocalRows,nLocalCols))
```
aradi:
```suggestion
      allocate(T2(nLocalRows, nLocalCols))
```

src/dftbp/timedep/timeprop.F90
```diff
@@ -1530,28 +1579,53 @@ subroutine updateH(this, H1, ints, H0, speciesAll, qq, q0, coord, orb, potential
...
             & neighbourList%iNeighbour, nNeighbourSK, this%iCellVec, this%cellVec, iSquare,&
             & iSparseStart, img2CentCell)
         call adjointLowerTriangle(H1(:,:,iKS))
       end if
     end do
+  #:endif
 
-    ! add hybrid xc-functional contribution
+  ! add hybrid xc-functional contribution
+  ! TODO: this is currently disabled because tests C4H6_rs and C4H6_rs_Singlet were failing.
```
aradi, line 1622: I think, those tests had been fixed on main now.

src/dftbp/timedep/timeprop.F90
```diff
@@ -1672,51 +1801,66 @@ subroutine getTDFunction(this, startTime)
...
+      if (env%mpi%tGlobalLead) then
+#:endif
       call openFile(laserDat, "laser.dat", mode="r")
+      do iStep = 0,this%nSteps
+        read(laserDat%unit, *)time, tdfun(1), tdfun(2), tdfun(3)
+        this%tdFunction(:, iStep) = tdfun * (Bohr__AA / Hartree__eV)
+      end do
+#:if WITH_MPI
+      end if
+      call mpifx_allreduceip(env%mpi%globalComm, this%tdFunction, MPI_SUM)
```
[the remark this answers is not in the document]
aradi: Once more, can you use a simple broadcast here? (As nothing is summed...)

src/dftbp/timedep/timeprop.F90
```diff
@@ -1672,51 +1801,66 @@ subroutine getTDFunction(this, startTime)
...
-      end if
+        if (this%envType == envTypes%constant) then
+          envelope = 1.0_dp
+        else if (this%envType == envTypes%gaussian) then
+          envelope = exp(-4.0_dp*pi*(time-midPulse)**2 / deltaT**2)
+        else if (this%envType == envTypes%sin2 .and. time >= this%time0 .and. time <= this%time1) then
+          envelope = sin(pi*(time-this%time0)/deltaT)**2
+        else
+          envelope = 0.0_dp
+        end if
```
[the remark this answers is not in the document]
aradi: Note, the indentation here is still inconsistent. See previous comment. :wink:

src/dftbp/timedep/timeprop.F90
```diff
@@ -1672,51 +1801,66 @@ subroutine getTDFunction(this, startTime)
...
+              & this%tdFunction(:, iStep) * (Hartree__eV / Bohr__AA)
+      #:if WITH_MPI
+        end if
+      #:endif
 
-    end do
+      end do
 
-    call closeFile(laserDat)
+    end if
```
aradi, line 1851: I'd prefer
```
#:if WITH_MPI
  writeLaserDat = env%mpi%tGlobalLead
#:else
  writeLaserDat = .true.
#:endif
```
at the beginning of the routine. and then two normal blocks without preprocessor options:
```
if (writeLaserDat) then
  ! Open or write
end if
```
so that we do not have to split block constructs with preprocessor.

src/dftbp/timedep/timeprop.F90
```diff
@@ -2210,15 +2447,30 @@ subroutine initializeTDVariables(this, densityMatrix, rho, H1, Ssqr, Sinv, H0, h
               & this%iCellVec, this%cellVec, iSquare, iSparseStart, img2CentCell)
           call adjointLowerTriangle(T4)
           Ssqr(:,:,iKS) = T4
-          Sinv(:,:,iKS) = cmplx(0,0,dp)
+          Sinv(:,:,iKS) = cmplx(0, 0, dp)
```
aradi, line 2455:
```suggestion
          Sinv(:,:,iKS) = cmplx(0, 0, kind=dp)
```

src/dftbp/timedep/timeprop.F90
```diff
@@ -2866,7 +3287,7 @@ end subroutine writeTDOutputs
 
   !> Initialize matrices for populations
   !! Note, this will need to get generalised for complex eigenvectors
-  subroutine tdPopulInit(this, Eiginv, EiginvAdj, eigvecsReal, eigvecsCplx)
+  subroutine tdPopulInit(this, Eiginv, EiginvAdj,errStatus, eigvecsReal, eigvecsCplx)
```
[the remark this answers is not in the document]
aradi: Remove `errStatus`!

src/dftbp/timedep/timeprop.F90
```diff
@@ -2964,22 +3428,25 @@ subroutine updateBasisMatrices(this, env, electronicSolver, Eiginv, EiginvAdj, H
       @:PROPAGATE_ERROR(errStatus)
       if (this%tRealHS) then
         T2(:,:) = real(T1, dp)
-        call tdPopulInit(this, Eiginv(:,:,iKS), EiginvAdj(:,:,iKS), T2)
+        call tdPopulInit(this, Eiginv(:,:,iKS), EiginvAdj(:,:,iKS), errStatus,T2 )
```
aradi:
```suggestion
        call tdPopulInit(this, Eiginv(:,:,iKS), EiginvAdj(:,:,iKS), errStatus, eigVecsReal=T2)
```

src/dftbp/timedep/timeprop.F90
```diff
@@ -3123,8 +3641,7 @@ subroutine initIonDynamics(this, coordNew, coord)
     ! Velocities should actually be v(t+0.5*dt), not v(t),
     ! like this: this%movedVelo(:,:) = this%movedVelo + 0.5_dp * movedAccel * this%dt
     coordNew(:,:) = coord
-    coordNew(:,this%indMovedAtom) = coordNew(:,this%indMovedAtom) &
-        & + this%movedVelo(:,:) * this%dt
+    coordNew(:,this%indMovedAtom) = coordNew(:,this%indMovedAtom) + this%movedVelo(:,:) * this%dt
```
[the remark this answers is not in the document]
aradi: Note prev. comment

src/dftbp/timedep/timeprop.F90
```diff
@@ -3811,7 +4372,13 @@ subroutine getBondPopulAndEnergy(this, bondWork, lastBondPopul, rhoPrim, ham0, i
         call addPairWiseBondInfo(bondWork, rhoPrim(:,1), ints%overlap, iSquare,&
             & iNeighbour, nNeighbourSK, img2CentCell, iSparseStart)
       end do
+    #:if WITH_SCALAPACK
+      if (env%mpi%tGlobalLead) then
+      write(fdBondPopul%unit) time * au__fs, sum(bondWork), bondWork
```
aradi: Indentation!

src/dftbp/timedep/timeprop.F90
```diff
@@ -4017,40 +4585,54 @@ subroutine initializeDynamics(this, boundaryCond, coord, orb, neighbourList, nNe
...
-    allocate(this%RdotSprime(this%nOrbs,this%nOrbs))
     allocate(this%totalForce(3, this%nAtom))
     this%RdotSprime(:,:) = 0.0_dp
     this%totalForce(:,:) = 0.0_dp
     this%occ(:) = 0.0_dp
 
     if (this%tReadRestart) then
+#:if WITH_SCALAPACK
+      call readRestartFileBlacs(this%trho, this%trhoOld, coord, this%movedVelo, this%time, this%dt,&
+      & restartFileName, env, this%denseDesc, this%parallelKS,  errStatus)
```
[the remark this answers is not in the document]
aradi: Deindent preproc by one level.

src/dftbp/timedep/timeprop.F90
```diff
@@ -4710,4 +5314,80 @@ subroutine finalizeDynamics(this)
...
+    !> K-Spin mixed index
+    integer, intent(in) :: iKS
+
+    !> Molecular orbital occupations
+    real(dp), intent(inout) :: occ(:)
+
+    !> Dense matrix description
+    type(TDenseDescr), intent(in) :: desc
+
+    real(dp), intent(in) :: T1_R(:,:)
```
aradi:
```suggestion
    real(dp), intent(in) :: T1R(:,:)
```

src/dftbp/timedep/timeprop.F90
```diff
@@ -4710,4 +5314,80 @@ subroutine finalizeDynamics(this)
...
+    do iAtom1 = 1, this%nAtom
+      arrayMO(:,:) = 0.0_dp
+      ii = desc%iAtomStart(iAtom1)
+      nOrb1 = desc%iAtomStart(iAtom1+1) - ii
+      iNeigh = 0
+      iOrig = iSparseStart(iNeigh, iAtom1) + 1
+      arrayMO(1:nOrb1, 1:nOrb1) = reshape(popSparse(iOrig:iOrig+nOrb1*nOrb1-1), [nOrb1, nOrb1])
+      do jj = 1, nOrb1
+        occ(ii+jj-1) = arrayMO(jj,jj)
+      end do
```
aradi:
```suggestion
      ! Diagonal =  every (nOrb + 1)th element in the flat form of an nOrb1 x nOrb1 matrix
      occ(ii : ii + nOrb1 - 1) = popSparse(iOrig : iOrig + nOrb1 * nOrb1 - 1 : nOrb + 1)
 ```
We do not need the reshape, the loop and the temporary matrix `arrayM0`

File: src/dftbp/dftb/populations.F90 (hunk 13 above)
Comment (aradi) [reply src/dftbp/dftb/populations.F90:1086]: OK, it might be correct, if q0 is in the ud-form within this routine. But make sure to check it explicitely.

File: src/dftbp/timedep/dynamicsrestart.F90 [hunk 35]
```diff
@@ -261,4 +272,180 @@ subroutine readRestartFile(rho, rhoOld, coord, veloc, time, dt, fileName, isAsci
...
+    end do
+
+    if (env%mpi%tGlobalLead) then
+      call closeFile(fd)
+    end if
+
+  end subroutine writeRestartFileBlacs
+
+
+  !> Write to a restart file the DM in Blacs format
```
aradi:
```suggestion
  !> Read from a restart file the DM in disributed Blacs format
```

src/dftbp/timedep/timeprop.F90
```diff
@@ -1671,52 +1805,53 @@ subroutine getTDFunction(this, startTime)
...
-        envelope = 1.0_dp
-      else if (this%envType == envTypes%gaussian) then
-        envelope = exp(-4.0_dp*pi*(time-midPulse)**2 / deltaT**2)
-      else if (this%envType == envTypes%sin2 .and. time >= this%time0 .and. time <= this%time1) then
-        envelope = sin(pi*(time-this%time0)/deltaT)**2
-      else
-        envelope = 0.0_dp
-      end if
-
+    if (writeLaserDat) then
```
aradi: I find the name misleading, as depending on `this%tEnvFromFile`, this block might not write at all.

src/dftbp/timedep/timeprop.F90
```diff
@@ -1671,52 +1805,53 @@ subroutine getTDFunction(this, startTime)
...
+          envelope = exp(-4.0_dp*pi*(time-midPulse)**2 / deltaT**2)
+        else if (this%envType == envTypes%sin2 .and. time >= this%time0 .and. time <= this%time1) then
+          envelope = sin(pi*(time-this%time0)/deltaT)**2
+        else
+          envelope = 0.0_dp
+        end if
+
+        this%tdFunction(:, iStep) = E0 * envelope * aimag(exp(imag*(time*angFreq + this%phase))&
+            & * this%fieldDir)
+        if (writeLaserDat) then
```
aradi: As we are here already inside an `if (writeLaserDat)` block, this if does not make really sense.

src/dftbp/timedep/timeprop.F90
```diff
@@ -2385,12 +2640,22 @@ subroutine initializePropagator(this, env, step, rho, rhoNew, H1, Sinv, coordAll
...
+      #:if WITH_SCALAPACK
+        call propagateRhoBlacs(this, rhoNew(:,:,iKS), rho(:,:,iKS), H1(:,:,iKS), Sinv(:,:,iKS),&
+            & step)
+      #:else
         call propagateRho(this, rhoNew(:,:,iKS), rho(:,:,iKS), H1(:,:,iKS), Sinv(:,:,iKS), step)
+      #:endif
       else
         ! The following line is commented to make the fast propagate work since it needs a real H
-        !H1(:,:,iKS) = imag * H1(:,:,iKS)
+      #:if WITH_SCALAPACK
```
aradi, line 2664: The comment in the prev. is not valid any more, should be removed.

src/dftbp/timedep/timeprop.F90
```diff
@@ -2883,13 +3293,51 @@ subroutine tdPopulInit(this, Eiginv, EiginvAdj, eigvecsReal, eigvecsCplx)
...
+    nLocalCols = this%denseDesc%fullSize
+  #:endif
+
+    allocate(T1(nLocalRows, nLocalCols))
+    allocate(T2(nLocalRows, nLocalCols))
+    allocate(T3(nLocalRows, nLocalCols))
 
-    allocate(T2(this%nOrbs, this%nOrbs))
-    allocate(T3(this%nOrbs, this%nOrbs))
+  #:if WITH_SCALAPACK
```
aradi, line 3233: Any reason, why the parallel implementation uses a different algorithm as the serial one?

aradi (commented):
I have only very minor comments and one question. Afterwards it can be merged IMO .

sys/gnu.cmake
```diff
@@ -71,9 +71,9 @@ set(C_FLAGS_COVERAGE "-O0 -g --coverage")
 # LAPACK and BLAS
 # (if the BLAS library contains the LAPACK functions, set LAPACK_LIBRARY to "NONE")
 #set(BLAS_LIBRARY "openblas" CACHE STRING "BLAS libraries to link")
-#set(BLAS_LIBRARY_DIR "" CACHE STRING "Directories where BLAS libraries can be found")
+set(BLAS_LIBRARY_DIR "" CACHE STRING "Directories where BLAS libraries can be found")
```
aradi: This change (and the two similar ones below) should be reverted. Settig a cache variable to the empty string is not necerrarily the same as not setting it at all. (Or we should discuss if change is needed...)

sys/gnu.cmake
```diff
@@ -82,7 +82,7 @@ set(C_FLAGS_COVERAGE "-O0 -g --coverage")
 #set(PARPACK_LIBRARY "parpack" CACHE STRING "PARPACK library (with path if necessary)")
 
 # ScaLAPACK -- only needed for MPI-parallel build
-#set(SCALAPACK_LIBRARY "scalapack-openmpi" CACHE STRING "Scalapack libraries to link")
+set(SCALAPACK_LIBRARY "scalapack-openmpi" CACHE STRING "Scalapack libraries to link")
```
aradi: This should remain commented out. The name for the scalapack library is in most cases `scalapack`. This is only a specific case for Debian/Ubuntu installation, so users should it set manually on those systems.

test/app/dftb+/timeprop/NO_pSIC/dftb_in.hsd
```diff
@@ -74,4 +74,7 @@ Parallel {
     # Allow OMP threads explicitely to test for hybrid parallelisation with
     # MPI-binary. (Check the manual before using this in production runs!)
     UseOmpThreads = Yes
+   Blacs = BlockSize {
+        1
```
bhourahine, line 79:
```suggestion
        # Do not use this setting in production runs
        1
```

test/app/dftb+/timeprop/CH3_dftbu_singlet/dftb_in.hsd
```diff
@@ -74,4 +74,7 @@ Parallel {
   # Allow OMP threads explicitely to test for hybrid parallelisation with
   # MPI-binary. (Check the manual before using this in production runs!)
   UseOmpThreads = Yes
+   Blacs = BlockSize {
+        1
```
bhourahine, line 79:
```suggestion
        # Do not use this setting in production runs
        1
```

test/app/dftb+/tests
```diff
@@ -22,32 +22,31 @@ timeprop/C4H6_rs                       #? OMP_THREADS <= 4 and not WITH_MPI
...
 timeprop/GNR_kpoint_restart            #? OMP_THREADS <= 4 and not WITH_MPI and LONG_TEST
-timeprop/H2_singlet                    #? OMP_THREADS <= 1 and not WITH_MPI
-timeprop/H2_triplet                    #? OMP_THREADS <= 1 and not WITH_MPI
-timeprop/H2_exc_atoms                  #? OMP_THREADS <= 1 and not WITH_MPI
-timeprop/C60_kick                      #? OMP_THREADS <= 4 and not WITH_MPI and LONG_TEST
-timeprop/C60_triplet                   #? OMP_THREADS <= 4 and not WITH_MPI and LONG_TEST
+timeprop/H2_singlet                    #? OMP_THREADS <= 4 and MPI_PROCS <=2
+timeprop/H2_triplet                    #? OMP_THREADS <= 4 and MPI_PROCS <=2
+timeprop/H2_exc_atoms                  #? OMP_THREADS <= 4 and MPI_PROCS <=2
+timeprop/C60_kick                      #? OMP_THREADS <= 4 and MPI_PROCS <=2 and LONG_TEST
```
bhourahine: Large enough that can probably run on up to 4 processors

src/dftbp/timedep/timeprop.F90
```diff
@@ -50,7 +50,7 @@ module dftbp_timedep_timeprop
   use dftbp_dftbplus_qdepextpotproxy, only : TQDepExtPotProxy
   use dftbp_elecsolvers_elecsolvers, only : TElectronicSolver
   use dftbp_extlibs_tblite, only : TTBLite
-  use dftbp_io_message, only : warning
+  use dftbp_io_message, only : error, warning
```
bhourahine: You are using the @:RAISE_ERROR macro, so is the error() routine needed?

src/dftbp/timedep/timeprop.F90
```diff
@@ -2265,6 +2508,18 @@ subroutine initializeTDVariables(this, densityMatrix, rho, H1, Ssqr, Sinv, H0, h
...
     if (.not.this%tReadRestart) then
       rho(:,:,:) = 0.0_dp
+    #:if WITH_SCALAPACK
+      do iKS = 1, this%parallelKS%nLocalKS
+        iK = this%parallelKS%localKS(1, iKS)
+        iSpin = this%parallelKS%localKS(2, iKS)
+        if (this%tRealHS) then
+          call makeDensityMtxRealBlacs(env%blacs%orbitalGrid, this%denseDesc%blacsOrbSqr,&
+              & filling(:,1,iSpin), eigvecsReal(:,:,iKS), T2)
+          rho(:,:,iKS) = cmplx(T2, 0, dp)
```
bhourahine: If matching Bálint's comment above:
```suggestion
          rho(:,:,iKS) = cmplx(T2, 0, kind=dp)
```

src/dftbp/timedep/timeprop.F90
```diff
@@ -2428,6 +2694,122 @@ subroutine propagateRho(this, rhoOld, rho, H1, Sinv, step)
...
+
+    complex(dp), allocatable :: T1(:,:)
+
+    integer :: nLocalCols, nLocalRows
+
+    nLocalRows = size(rho, dim=1)
+    nLocalCols = size(rho, dim=2)
+    allocate(T1(nLocalRows, nLocalCols))
+
+    T1(:,:) = 0.0_dp
```
bhourahine: Probably not needed, as pgemm beta=0.0 by default.

src/dftbp/timedep/timeprop.F90
```diff
@@ -2883,13 +3293,51 @@ subroutine tdPopulInit(this, Eiginv, EiginvAdj, eigvecsReal, eigvecsCplx)
...
 
-    allocate(T2(this%nOrbs, this%nOrbs))
-    allocate(T3(this%nOrbs, this%nOrbs))
+  #:if WITH_SCALAPACK
     if (this%tRealHS) then
-      T2(:,:) = cmplx(eigvecsReal, kind=dp)
+      T2(:,:) = cmplx(eigvecsReal, 0, dp)
+    end if
+
+    ! invert eigvecsReal with pgetrf and pgetri
```
bhourahine, line 3238: Something for another pull request later: We should probably eventually move this to a pmatinv routine in math/ somewhere, as there is already a lapack equivalent in math/matrixops.F90.

bhourahine (commented):
Just some comments/suggestions

src/dftbp/timedep/timeprop.F90
fbonafe: This is due to performance reasons, should we mention the reason in a comment?

sys/gnu.cmake
fbonafe: Sorry, this was an accidental commit!

sys/gnu.cmake
fbonafe: This was also accidental, sorry

PR #1391, opened 2024-02-11
Improve import manipulation of Fortran files

jubich:
This PR adds new scripts for editing/formatting imports in Fortran files (`.F90 and .f90`). 

`sort_modules` now sorts not only the modules alphabetically, but also the imported functions/methods. This script was also used for sorting the imports in #1382. You might consider switching from `.sort()` to .`sort(key=str.casefold)` to get a case-insensitive sort.

`check_import_usage` checks whether the imported functions/methods (with the exception of assignment() and operator()) are used or not. This is done by finding all "import lines". These are removed and then the remaining file is converted into a `set` of lowercase words. These are then compared with a `set` created from the imported functions/methods.

### DIFF
--- a/utils/srcmanip/check_import_usage
+++ b/utils/srcmanip/check_import_usage
@@ -1,0 +1,137 @@
+#!/usr/bin/env python3
+
+"""Simple script to check usage of imports in Fortran files"""
+
+import argparse
+import fnmatch
+import os
+import re
+
+
+_DESCRIPTION = """Checks the imports in DFTB+ Fortran file(s) for usage
+WARNING: This script uses a simple heuristics to check import usage. It only checks wether the
+names of imported functions/methods are present in the file or not."""
+
+_PAT_USE_MODULE = re.compile(
+    r"""^(?P<indent>[ ]*)use
+    (?P<attrib>(?:\s*,\s*intrinsic)?)
+    (?P<separator>[ ]*::[ ]*|[ ]*)
+    (?P<name>\w+)
+    (?P<rest>.*?(?:&[ ]*\n(?:[ ]*&)?.*?)*)\n
+    """,
+    re.VERBOSE | re.MULTILINE | re.IGNORECASE,
+)
+
+WORD_PATTERN = re.compile(r"""\w+""")
+
+COMMENT_SYMBOLS = ["!!", "!>", "! "]
+
+
+def main():
+    """Main script driver."""
+    args = parse_arguments()
+    filenames = []
+    if args.folders:
+        filenames = get_files(args.folders)
+    if args.files:
+        filenames += args.files
+    if args.file:
+        filenames += args.file
+
+    for fname in filenames:
+        with open(fname, "r", encoding="utf-8") as file:
+            file_txt = file.read()
+            matches = list(_PAT_USE_MODULE.finditer(file_txt))
+            matches.reverse()
+            rest_list = []
+
+            # collects all imported functions/methods and removes imports from file
+            for match in matches:
+                if match["rest"]:
+                    rest_list.append(match["rest"])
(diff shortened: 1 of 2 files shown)


aradi (commented):
Please adapt to line lenght 100. (You can reformat mit `black -l 100`)

aradi:
ping (unstale)

utils/srcmanip/check_import_usage
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env python3
+
+"""Simple script to ckeck usage of imports in Fortran files"""
```
aradi, line 3:
```suggestion
"""Simple script to check usage of imports in Fortran files"""
```

utils/srcmanip/check_import_usage
```diff
@@ -0,0 +1,110 @@
...
+_DESCRIPTION = "Checks the imports in DFTB+ Fortran file(s) for usage"
+
+_PAT_USE_MODULE = re.compile(
+    r"""^(?P<indent>[ ]*)use
+    (?P<attrib>(?:\s*,\s*intrinsic)?)
+    (?P<separator>[ ]*::[ ]*|[ ]*)
+    (?P<name>\w+)
+    (?P<rest>.*?(?:&[ ]*\n(?:[ ]*&)?.*?)*)\n
+    """,
+    re.VERBOSE | re.MULTILINE,
```
aradi, line 20: Maybe we should add also `re.IGNORECASE` (as we're dealing with Fortran :wink: )

utils/srcmanip/check_import_usage
```diff
@@ -0,0 +1,110 @@
...
+    for fname in filenames:
+        with open(fname, "r", encoding="utf-8") as file:
+            file_txt = file.read()
+            matches = list(_PAT_USE_MODULE.finditer(file_txt))
+            matches.reverse()
+            rest_list = []
+            for match in matches:
+                if match["rest"]:
+                    rest_list.append(match["rest"])
+                file_txt = file_txt[: match.start()] + file_txt[match.end() :]
```
aradi, line 43: What's the point to check the source *before* the import? It is enough to check only the text which comes after the match.

aradi (changes requested):
Thanks for the PR. Please change following:
* remove the `--files` and `--folders` options in both scripts, that is quite unusual. The commands should simply accept files as positional arguments. (We can easily generate file names in folders, if necessary).
* We should add a warning into the `check_import_usage` that it uses a very simple heuristics, so that people are aware, that it can fooled in many ways.

utils/srcmanip/check_import_usage
aradi: Could, you, please also try to rebase it to current main?

jubich:
I have now changed both files so that they now accept both **positional** and optional arguments. This allows the scripts to be called in the _usual_ way with positional arguments as well as _unusually_ with optional arguments. This also allows to call both functions e.g. without files, but only with folders.


I have also “improved” the import checker so that comments are ignored. Regarding your question, why lines before the import should be checked:
They obviously do not need to be checked. The current implementation first searches _all_ imports. Then it reverses the order, removes the matches from the read file and stores the matches in a `list`. The reversal is necessary because otherwise the `.start()` and `.end()` positions of the matches would be wrong due to the shift after removing the above import. I now additionally remove the lines before the first import (which mainly removes the header). 
Next, the comments are removed and the remaining file is converted to a `set`.
Now the `list` with the saved imports is also converted into a `set` and the difference between the two `set's` is calculated. The difference then contains the _unused_ imports. 

I cannot simply remove all lines before the last import, because sometimes there are imports that are not at the beginning of the file. So I wouldn't check the usage between such imports and imports at the beginning of the file. I could check the file for each import individually, but I doubt that would bring much benefit.

PR #1394, opened 2024-02-13
Optimize construction of real HFX Hamiltonian

vanderhe:
Exploits symmetric/hermitian property of $\mathbf{S}(\Gamma)$ and $\mathbf{P}(\Gamma)$ in serial HFX Hamiltonian construction routines.


bhourahine:
The symm operations aren't actually much different in performance than gemm, but what they do is to ensure correct symmetry.

vanderhe:
> The symm operations aren't actually much different in performance than gemm, but what they do is to ensure correct symmetry.

I see, so we really only spare a subsequent manual symmetrization to account for rounding effects.

vanderhe:
@bhourahine Rebased on main.

bhourahine (approved):
LGTM

PR #1396, opened 2024-02-14
Minor simplify/clean-up in parsers

bhourahine:
Simplify a few if then else in various code parsers, plus a line length fix.


vanderhe (approved):
LGTM. Do we have a general strategy for these cases?

vanderhe:
@bhourahine I couldn't reproduce the failing tests on dpdevel yet, any idea?

bhourahine:
@tammo  I think it's the parser, I also saw this when trying to inline if statements in the getChild() changes of #1397

vanderhe:
> @tammo I think it's the parser, I also saw this when trying to inline if statements in the getChild() changes of #1397

@bhourahine I see, then I guess the Fortuno development has to be put on hold in favor of the parser getting fixed ;)
Btw. you send your message to some random guy, instead of me. Let's see if we can convince my namesake to join the devel. group :D

bhourahine:
@vanderhe Ran fine this time, suggesting it was a problem on github's side.

PR #1398, opened 2024-02-19
Clean up SK-parser/-data module

vanderhe:
I'm planning to implement a simple check that iterates through all SK-files and ensures that the hybrid xc-functional parameters actually match, since we're currently opting for a "hope-for-the-best" solution. This is just some pre-processing for the actual PR.


bhourahine (approved):
LGTM

PR #1399, opened 2024-02-20
Consistent imaginary unit

vanderhe:
Consistently use the imaginary unit parameter from the `dftbp_common_constants` module.


bhourahine (approved):
LGTM

PR #1401, opened 2024-02-24
Fixes SCC info being 1 iteration behind

bhourahine:
SCC step 1 was printing 0 for electronic energy, step 2 printed step 1 electronic energy, etc.

Bug probably introduced by 5bbb8f9a5d


vanderhe (approved):
LGTM

PR #1402, opened 2024-02-27
Checks consistency of cmake and submodule repo

bhourahine:
Changes script to optionally check for 
* Mismatch between CMakeLists.txt and .gitmodules urls
* Whether the git urls have consistent .git extension (i.e. headless or not)
* Update the 2 urls that point to the wrong places

### DIFF
--- a/utils/test/check_submodule_commits
+++ b/utils/test/check_submodule_commits
@@ -2,10 +2,11 @@
 
 """Checks the submodule commit IDs in integration CMake files."""
 
-import os
-import subprocess
-import re
 import argparse
+import configparser
+import os
+import re
+import subprocess
 import sys
 
 _DESCRIPTION = """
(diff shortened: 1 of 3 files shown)


utils/test/check_submodule_commits
```diff
@@ -64,8 +74,10 @@ def main():
 def _parse_arguments():
     """Returns command line arguments"""
     parser = argparse.ArgumentParser(description=_DESCRIPTION)
-    msg = "Whether the commits should be updated in the CMakeLists.txt files"
-    parser.add_argument("-u", "--update", action='store_true', help=msg)
+    msgu = "Whether the commits should be updated in the CMakeLists.txt files"
+    parser.add_argument("-u", "--update", action='store_true', help=msgu)
+    msgc = "Check consistency between CMakeLists.txt and submodules"
```
aradi, line 79: Any reason for not just reusing the `msg` variable? (Once `parser.add_argument()` had been called, the content of `msg` had been copied to an internal variable of the argument parser.)

utils/test/check_submodule_commits
```diff
@@ -33,23 +34,32 @@ def main():
...
+        gitconfig.read('.git/config')
     allsync = True
     with open("CMakeLists.txt", "r") as fp:
         origcontent = fp.read()
     content = origcontent
     for match in _PAT_SUBMOD_COMMIT.finditer(output):
         commit, submodule = match.groups()
+        if args.check:
+            locale = ('submodule "external/%s/origin"' % submodule)
+            active=gitconfig.get(locale, 'active')
```
aradi, line 48:
```suggestion
            active = gitconfig.get(locale, 'active')
```

utils/test/check_submodule_commits
```diff
@@ -33,23 +34,32 @@ def main():
...
+            if (active == 'true'):
```
aradi, line 49:
```suggestion
            if active == 'true':
```

aradi (approved):
Thx, green light from my side, I have only minor stylistic suggestions.

utils/test/check_submodule_commits
bhourahine: msg now used for both

PR #1404, opened 2024-03-04
Remove redundant variable

bhourahine:


vanderhe (approved):
True, `coeffsDiag` is superfluous here.

PR #1405, opened 2024-03-05
Re-enable constraint tests

bhourahine:
As this feature will be in the next release, and may get broken if not tested during PR merges.


vanderhe (approved):
LGTM

PR #1406, opened 2024-03-05
Add test for missing Meth-Pax k-point filling

bhourahine:
As the density matrix build goes through a different path if non-Fermi filling.


test/app/dftb+/non-scc/10-10Ctube_MP/C-C.skf
vanderhe, line 1: Can we get rid of this symlink? I know that e.g. `10-10Ctube` uses the same method but would suggest to switch over to the new strategy everywhere (on occasion).

bhourahine: Shall we do this after #1407 is merged?

vanderhe (approved):
LGTM

PR #1413, opened 2024-03-12
Optional text string for debug macro

bhourahine:
Adds optional string support for @:DEBUG()

### DIFF
--- a/src/dftbp/dftbplus/main.F90
+++ b/src/dftbp/dftbplus/main.F90
@@ -1161,7 +1161,7 @@ contains
       call buildS(env, this%ints%overlap, this%skOverCont, this%coord, this%nNeighbourSk,&
           & this%neighbourList%iNeighbour, this%species, this%iSparseStart, this%orb)
     case(hamiltonianTypes%xtb)
-      @:ASSERT(allocated(this%tblite))
+      @:ASSERT(allocated(this%tblite), "Compiled without TBLITE included")
       call this%tblite%buildSH0(env, this%species, this%coord, this%nNeighbourSk, &
           & this%neighbourList%iNeighbour, this%img2CentCell, this%iSparseStart, &
           & this%orb, this%H0, this%ints%overlap, this%ints%dipoleBra, this%ints%dipoleKet, &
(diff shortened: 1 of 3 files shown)


src/dftbp/include/common.fypp
```diff
@@ -42,12 +42,17 @@
 #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
 #! Check a condition if WITH_ASSERT is True and call assertError if condition is False.
-#:def ASSERT(cond)
+#! If an optional text string is included, print this in addition as an error
+#:def ASSERT(cond, *msg)
```
aradi, line 46: I'd suggest to go with `msg=None`

src/dftbp/include/common.fypp
```diff
@@ -42,12 +42,17 @@
...
   #:if WITH_ASSERT
   if (.not. (${cond}$)) then
     block
       use dftbp_common_assert, only : assertError
+      #:if len(msg) > 0
+      call assertError("${_FILE_}$", ${_LINE_}$, "${' '.join(msg)}$")
```
aradi, line 52:
```suggestion
      #:if msg
        call assertError("${_FILE_}$", ${_LINE_}$, ${msg}$)
```

src/dftbp/dftbplus/main.F90
```diff
@@ -1172,7 +1172,7 @@ subroutine processGeometry(this, env, iGeoStep, iLatGeoStep, tWriteRestart, tSto
       call buildS(env, this%ints%overlap, this%skOverCont, this%coord, this%nNeighbourSk,&
           & this%neighbourList%iNeighbour, this%species, this%iSparseStart, this%orb)
     case(hamiltonianTypes%xtb)
-      @:ASSERT(allocated(this%tblite))
+      @:ASSERT(allocated(this%tblite), Compiled without TBLITE included)
```
aradi, line 1175:
```suggestion
      @:ASSERT(allocated(this%tblite), "Compiled without TBLITE included")
```
I think, it would be easier to read and more robust, if we quoted the string argument as in Fortran.

bhourahine: Done.

src/dftbp/include/common.fypp
bhourahine: Done.

PR #1415, opened 2024-03-15
Merge parallel into schedule module

vanderhe:


src/dftbp/dftb/hybridxc.F90
```diff
@@ -26,7 +26,7 @@ module dftbp_dftb_hybridxc
   use dftbp_dftb_nonscc, only : buildS
   use dftbp_math_matrixops, only : adjointLowerTriangle
   use dftbp_math_simplealgebra, only : determinant33
-  use dftbp_common_parallel, only : getStartAndEndIndex
+  use dftbp_common_schedule, only : getStartAndEndIndex
```
bhourahine, line 29: Can you move this into alphabetical order with the other modules?

bhourahine (approved):
Almost ready...

src/dftbp/dftb/hybridxc.F90
vanderhe: done

PR #1422, opened 2024-03-20
Limit numer of procs for sub-grid test

### LINKED ISSUE #1309: Incorrect configuration selection rules for regression test(s)
There is at least one (_spinorbit/As4S4_) regression tests executed for an invalid build configuration, e.g. `-DTEST_MPI_PROCS=7`:
https://github.com/dftbplus/dftbplus/blob/9dd77907d532fae709f9a8ca606c8109b9d6dc1b/test/app/dftb%2B/tests#L394

Additional known cases (`-DTEST_MPI_PROCS=3`):
- api_mpi_intcharges_c
- dftb+_derivatives/Au2_LSmetalPolfreq
- dftb+_dispersion/diamond_dftd4
- dftb+_dispersion/diamond_dftd4_sc

I did not yet perform a systematic, exhaustive search for such cases, so there might be many more cases.

bhourahine:
As this MPI test case won't run on too many processors

Closes #1309

### DIFF
--- a/test/app/dftb+/tests
+++ b/test/app/dftb+/tests
@@ -253,7 +253,7 @@ derivatives/C60_Polfreq                #? MPI_PROCS <= 4
 
 derivatives/CH4_freq                   #? not WITH_MPI or MPI_PROCS == 2
 derivatives/CH4_kpt_freq               #? not WITH_MPI or MPI_PROCS == 4
-derivatives/Au2_LSmetalPolfreq         #? MPI_PROCS <= 4
+derivatives/Au2_LSmetalPolfreq         #? MPI_PROCS <= 2
 
 legacy/Si2_oldSKinterp                 #? MPI_PROCS <= 1
 legacy/Si2_polyRep                     #? MPI_PROCS <= 1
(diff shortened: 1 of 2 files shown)


vanderhe:
Can we smuggle in a fix for #1309?

bhourahine:
> Can we smuggle in a fix for #1309?

Sure, but at some point we need to systematically test over processor number  (shouldn't be too hard, just a shell loop over procs and test for fails for a binary with everything enabled in the build).

vanderhe:
> > Can we smuggle in a fix for #1309?
> 
> Sure, but at some point we need to systematically test over processor number (shouldn't be too hard, just a shell loop over procs and test for fails for a binary with everything enabled in the build).

It is quite annoying to correctly specify the requirements for MPI processes as $\leq$ will not do the trick, e.g. some higher (non-continuous) numbers might work.

bhourahine:
> It is quite annoying to correctly specify the requirements for MPI processes as ≤ will not do the trick, e.g. some higher (non-continuous) numbers might work.

Some tests use modulo, but what I was thinking is something simple, just go over:

```
for iproc in $(seq 32)
do
cmake -DTEST_MPI_PROCS=$iproc ..
make && ctest &>> log.txt
done
```

and grep for failures

vanderhe (approved):

PR #1427, opened 2024-03-30
Simplify logic and reuse calls

bhourahine:


vanderhe:
@bhourahine I did not yet have the time to figure out why the Intel workflow breaks, but #1419 now exhibits the same behavior.

bhourahine:
@vanderhe Intel had a new release recently, possibly not unrelated...

vanderhe:
@bhourahine Do we merge PRs, even though Intel fails (e.g. #1419)?

bhourahine:
@vanderhe I've seen Intel fail after just changing comments in code, so probably a compiler bug. Hence, I think we discuss tomorrow, but don't think it should hold up merging branches that pass CI on everything else.

PR #1432, opened 2024-04-06
Correct real to int and make fn pure

bhourahine:


aradi (approved):
I am fine with it. But we might consider to use the opportunity to add an `abs()`˙in lines 1084, which seems to more robust to me. (And maybe we could change the hardcoded literal `1e-16_dp` to `epsilon(1.0_dp)` to be precision independent? :wink:

vanderhe:
> I am fine with it. But we might consider to use the opportunity to add an `abs()`˙in lines 1084, which seems to more robust to me. (And maybe we could change the hardcoded literal `1e-16_dp` to `epsilon(1.0_dp)` to be precision independent? 😉

Sounds good to me, both suggestions.

vanderhe (approved):

bhourahine:
@aradi @vanderhe Can there be negative values? If so the logic is broken, since array would be sorted as +ve, 0s, -ve and the test stops at the 1st 0.

vanderhe:
> @aradi @vanderhe Can there be negative values? If so the logic is broken, since array would be sorted as +ve, 0s, -ve and the test stops at the 1st 0.

Negative values should not occur, since phase factors haven't been involved yet, i.e. the arrays contain raw $\gamma(\mathbf{g})$-evaluations. So the `abs()` actually won't have any effect.

bhourahine:
@vanderhe So the `abs()` actually won't have any effect. But there does need to be a comment explaing non-negative. Should we have an assert for negative number (dropping the pure status)?

@aradi Mentioning pure, is there a sensible way to have a macro for pure only when not in debugging mode? Something like
@:PURE ?

PR #1441, opened 2024-04-17
Change counting function to single pass

bhourahine:
Original version has worst case O(N^2) performance when all elements are unique, but not clear if that matters. This one has O(N ln N) from the sort, but uses same memory as original.


vanderhe (approved):
LGTM

PR #1442, opened 2024-04-19
Remove extraneous call and amalgamate loops

bhourahine:
The number of unique species variable is unchanged by the loops so can be reused.

Also, any reason for the hard coded 1E-8 values why not, for example,  single precision epsilon?


vanderhe:
> Also, any reason for the hard coded 1E-8 values why not, for example, single precision epsilon?

I guess single precision epsilon is a good idea. At some point we should also discuss the arbitrary decision made [here](https://github.com/dftbplus/dftbplus/pull/1442/files#diff-8a799342d3672af4df7fcd69aef919c8a083aa37cb51243b935858a603881cbfR518).

vanderhe:
@bhourahine Do you want to change the finite difference delta here or in another PR?

bhourahine:
Suggest doing this in another PR (or even analytically...)

vanderhe (approved):
LGTM

PR #1443, opened 2024-04-20
Analytical gamma second derivative

bhourahine:
Replace finite difference for global hybrids.

### DIFF
--- a/src/dftbp/dftb/rshgamma.F90
+++ b/src/dftbp/dftb/rshgamma.F90
@@ -23,7 +23,7 @@ module dftbp_dftb_rshgamma
   public :: getCamAnalyticalGammaValue_workhorse
   public :: getHfAnalyticalGammaValue_workhorse, getdHfAnalyticalGammaValue_workhorse
   public :: getLrAnalyticalGammaValue_workhorse, getdLrAnalyticalGammaValue_workhorse
-  public :: getddLrNumericalGammaValue_workhorse
+  public :: getddLrNumericalGammaValue_workhorse, getddHfAnalyticalGammaValue_workhorse
 
 
 contains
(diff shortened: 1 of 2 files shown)


vanderhe:
@bhourahine Did you already check that the new routines reproduce the old ones? This is not necessarily caught by the regression tests.

src/dftbp/dftb/rshgamma.F90
```diff
@@ -351,6 +351,56 @@ function getdHfAnalyticalGammaValue_workhorse(hubbu1, hubbu2, dist) result(dGamm
...
+    if (dist < tolSameDist) then
+      ! on-site case
+      ddGamma = 0.0_dp
+    else
+      ! off-site case, Ua == Ub
+      if (abs(tauA - tauB) < MinHubDiff) then
+        tauA = 0.5_dp * (tauA + tauB)
+
+        ddTmp = &
+            & (2.0_dp*tauA**3/48.0_dp + 2.0_dp/dist**3 - tauA*(2.0_dp*dist*tauA**3/48.0_dp&
```
vanderhe, line 385: Style guide violations in most lines ;)
```suggestion
            & (2.0_dp * tauA**3 / 48.0_dp + 2.0_dp / dist**3 - tauA * (2.0_dp * dist * tauA**3 / 48.0_dp&
```
+induced line breaks, if applicable

vanderhe (changes requested):

bhourahine:
> @bhourahine Did you already check that the new routines reproduce the old ones? This is not necessarily caught by the regression tests.

getddYGammaSubPart was explicity tested, but not the full term. If there is no coverage, how was the original expression tested? And any reason why a test can't be added?

vanderhe:
> > @bhourahine Did you already check that the new routines reproduce the old ones? This is not necessarily caught by the regression tests.
> 
> getddYGammaSubPart was explicity tested, but not the full term. If there is no coverage, how was the original expression tested? And any reason why a test can't be added?

Back then I verified that the 5-th order polynomial smoothly connects and plotted the whole thing. Although there is at least one test with the damped TCI, I'm not sure if it really hits the damped region.

vanderhe:
@bhourahine I have now systematically checked the second derivatives, let's discuss next devel. meeting.

bhourahine:
@aradi as the 2nd derivative is used in the polynomial tail, this might have a slight effect, will try and test more carefully.

PR #1445, opened 2024-04-25
Fix wrong printing for SA-REKS energy

inseonglee:
This error occurs from commit 2551dc7, and now correct SA-REKS energy is printed after converging the SCC procedure.

### DIFF
--- a/src/dftbp/dftbplus/main.F90
+++ b/src/dftbp/dftbplus/main.F90
@@ -571,7 +571,7 @@ contains
           & this%tStress, this%totalStress, pDynMatrix, pDipDerivMatrix, this%tPeriodic,&
           & this%cellVol, this%tMulliken, this%qOutput, this%q0, this%taggedWriter, this%cm5Cont,&
           & this%polarisability, this%dEidE, this%dqOut, this%neFermi, this%dEfdE,&
-          & this%dipoleMoment, this%multipoleOut, this%eFieldScaling)
+          & this%dipoleMoment, this%multipoleOut, this%eFieldScaling, this%reks)
     end if
     if (this%tWriteCosmoFile .and. allocated(this%solvation)) then
       call writeCosmoFile(this%solvation, this%species0, this%speciesName, this%coord0, &
@@ -1305,7 +1305,7 @@ contains
 
         if (tConverged .or. tStopScc) then
 
-          call printReksSAInfo(this%reks, this%dftbEnergy(1)%Etotal)
+          call printReksSAInfo(this%reks, this%dftbEnergy(1)%Eavg)
 
           call getStateInteraction(env, this%denseDesc, this%neighbourList, this%nNeighbourSK,&
               & this%iSparseStart, this%img2CentCell, this%coord, this%iAtInCentralRegion,&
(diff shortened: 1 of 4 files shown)


bhourahine (commented):
Should the averaged energy also be printed for the tagged output?

inseonglee:
> Should the averaged energy also be printed for the tagged output?

The correct value for SSR energies indicates the correct SCC process for averaged energy. So, we do not need to print the averaged energy for autotest.tag file. It is enough to print correct averaged energy in the stdout file.

bhourahine:
> > Should the averaged energy also be printed for the tagged output?
> 
> The correct value for SSR energies indicates the correct SCC process for averaged energy. So, we do not need to print the averaged energy for autotest.tag file. It is enough to print correct averaged energy in the stdout file.

Does that also apply to results.tag? Since that file contains the machine readable version of stdout and detailed.out data.

inseonglee:
> > > Should the averaged energy also be printed for the tagged output?
> > 
> > 
> > The correct value for SSR energies indicates the correct SCC process for averaged energy. So, we do not need to print the averaged energy for autotest.tag file. It is enough to print correct averaged energy in the stdout file.
> 
> Does that also apply to results.tag? Since that file contains the machine readable version of stdout and detailed.out data.

According to your opinion, it's better to include the energy for averaged state in results.tag file, but the CI fails.

bhourahine:
> According to your opinion, it's better to include the energy for averaged state in results.tag file, but the CI fails.

Do you mean the intel and macos? Those have other bugs due to recent compiler updates on github (see #1451). Or is this a change in the value of the tags (in which case the affected tests will need to be regenerated).

The question I have is essentially, which of the two energies is the physically correct one that users of the code should received?

inseonglee:
> > According to your opinion, it's better to include the energy for averaged state in results.tag file, but the CI fails.
> 
> Do you mean the intel and macos? Those have other bugs due to recent compiler updates on github (see #1451). Or is this a change in the value of the tags (in which case the affected tests will need to be regenerated).
> 
> The question I have is essentially, which of the two energies is the physically correct one that users of the code should received?

In my response, CI means the five tests in the PR (gcc/intel), the failed case is gcc-build w/ macos, nompi, debug mode.

From the commit 694072e, the results.tag file contains the energy of the averaged state, the example for new results.tag file for REKS case is below.
averaged_energy     :real:0:
 -0.165023470719967E+002
total_energy        :real:0:
 -0.165741684637794E+002
fermi_level         :real:1:1
  0.000000000000000E+000
...

In the your question, what the two energies mean?
If the two energies mean energy%Etotal and energy%Eavg, then both are important in REKS case, but the mainly used quantity would be energy%Etotal, which corrsponds to the energy of target state (may S0 or S1).

bhourahine:
> In the your question, what the two energies mean? If the two energies mean energy%Etotal and energy%Eavg, then both are important in REKS case, but the mainly used quantity would be energy%Etotal, which corrsponds to the energy of target state (may S0 or S1).

Yes. So it appears that for autotest.tag the energy%Etotal should be sufficient, but for the results.tag file both quantities should be available.

inseonglee:
> > In the your question, what the two energies mean? If the two energies mean energy%Etotal and energy%Eavg, then both are important in REKS case, but the mainly used quantity would be energy%Etotal, which corrsponds to the energy of target state (may S0 or S1).
> 
> Yes. So it appears that for autotest.tag the energy%Etotal should be sufficient, but for the results.tag file both quantities should be available.

Yes, as you said, autotest.tag does not contain the energy%Eavg, only results.tag contains energy%Eavg.

PR #1449, opened 2024-04-29
Draft for plugin loading mechanism

terminationshock:
This is a first draft of a plugin loading mechanism. An external C library implements certain functions (see `dftbplus_plugin.h`) and will be loaded in DFTB+ at startup. This example here has been created for `getSKIntegrals`.
Both the implementation of the plugin as well as the C binding in DFTB+ are quite lightweight. Further features are needed, like a geometry update notification.
At the moment, I am happy to hear comments and critizism.


bhourahine:
Thoughts so far:
* Eventually this will need some form of CI for plugins from outside of the test directory cmake system. So probably a cmake finder mechanism that can be used by others
* Multiple plugins eventually...
* in the parser, logically does the plugin belong in options {} ? Possibly a Plugins {} block?
* There is some overlap with #1420, so we should unify. Features from there that are probably relevant
  * plugins returning an announcement string and/or version number for what they are/do/which one
  * Consistency checking/announcement for what a plugin provides, and hence what code functionality can be used with it
* Features more generally: modify parser path depending on plugins loaded
* For current draft, once it gets more finalised: documentation and a changelog entry.

We should document all of this up and try to get buy in from other similar codes.

awvwgk:
Would it be possible to make a case study using an existing external library to implement it via the plugin mechanism? In this case many use cases would be covered which need to be supported for having a proper plugin working with DFTB+. An example could be tblite which is almost completely integrated via two main API routes (computing of integrals for Hamiltonian and exchange of charges and potentials), however still needs a significant change in the DFTB+ main code to work compared with SK based DFTB. Another case which is a bit simpler could be dispersion corrections like DFT-D3 or MBD (note that the latter requires quite some custom work).

terminationshock:
Further work will be done somewhat soon

PR #1451, opened 2024-05-02
Fix GitHub CI-related issues

aradi:
Quick fixes to make 
* Switch to cached Intel compiler with fixed version (fixes Intel hangup)
* Drop macOS tester (unreliably, failing recently more often than not, no possibility to debug it locally currently)
* Uncomment codecov related section, as codecov does not accept old style access any more and hangs.
*

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,11 +33,11 @@ jobs:
         mpi: [nompi, openmpi]
         config: [Debug]
         version: [13]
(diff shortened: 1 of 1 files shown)


bhourahine (approved):
Seems OK, but any chance of bringing the coverage back?

bhourahine:
@aradi OK, let's merge this, but we do need to get coverage working again soon.

PR #1453, opened 2024-05-04
Remove duplicate results tag

bhourahine:
The block after line 2136 of mainio.F90 is already printing the correct energy quantity (either internal or Mermin depending on the ensemble), so either the wrong energy is printed or two copies of the internal energy (total) are printed by the removed line.


bhourahine:
@aradi now with missing change returned to the tag writing.

PR #1455, opened 2024-05-05
Remove unused old energy eval routine

bhourahine:
Routine is not used anywhere and, as per its comments, is slower than the actual routines (plus is only for clusters).


vanderhe (approved):
This is just a naive implementation to verify the results of the more optimized versions. For debugging purposes I didn't remove it until now, but I guess one could always restore it from the history. LGTM.

PR #1456, opened 2024-05-06
Simplify real hybrid energy evaluation

bhourahine:
All three real code paths can use the same simple evaluation for energy (if the threshold one has the square H matrix).

Also fixes a comment for complex with k-points.


vanderhe (approved):
With the additional symmetrization in the thresholded case this should be equivalent to the previous behavior. LGTM.

PR #1457, opened 2024-05-08
Clean up HFX energy routines

vanderhe:
Clean up as discussed in the devel. meeting.

### DIFF
--- a/src/dftbp/timedep/timeprop.F90
+++ b/src/dftbp/timedep/timeprop.F90
@@ -2030,7 +2030,7 @@ contains
     ! Calculate nuclear kinetic energy
     energyKin = 0.0_dp
     if (this%tIons) then
-      energyKin = 0.5_dp * sum(this%movedMass(:,:) * this%movedVelo(:,:)**2)
+      energyKin = 0.5_dp * sum(this%movedMass * this%movedVelo**2)
       energy%Etotal = energy%Etotal + energyKin
     end if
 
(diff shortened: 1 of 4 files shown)


vanderhe:
@bhourahine I don't see any reason why the HFX energy should be queried by `getPositionDependentEnergy()`, since `calcEnergies()` inside of `getTDEnergy()` already calculates it. Any objections?

bhourahine:
> @bhourahine I don't see any reason why the HFX energy should be queried by `getPositionDependentEnergy()`, since `calcEnergies()` inside of `getTDEnergy()` already calculates it. Any objections?

I think I agree that it should be removed.
@fbonafe, any comments?

src/dftbp/timedep/timeprop.F90
```diff
@@ -2030,7 +2030,7 @@ subroutine getTDEnergy(this, env, energy, rhoPrim, rho, neighbourList, nNeighbou
     ! Calculate nuclear kinetic energy
     energyKin = 0.0_dp
     if (this%tIons) then
-      energyKin = 0.5_dp * sum(this%movedMass(:,:) * this%movedVelo(:,:)**2)
+      energyKin = 0.5_dp * sum(this%movedMass * this%movedVelo**2)
```
bhourahine, line 2033: Slightly off topic for this PR, but harmless and needed.

src/dftbp/timedep/timeprop.F90
```diff
@@ -3733,7 +3733,7 @@ end subroutine initLatticeVectors
 
   !> Calculates repulsive and dispersion energies
```
bhourahine, line 3734: This comment certainly suggests hybrid energy evaluation doesn't belong here.

bhourahine (approved):
LGTM

PR #1460, opened 2024-05-13
Add consistent finders arpack and parpack

aradi:


bhourahine:
@aradi what was the issue with the intel install? Otherwise, seems ready to merge.

PR #1464, opened 2024-05-16
Refactor reading routine with potential buffer overflow

aradi:
Code produced by the NAG compiler was crashing at this point

### DIFF
--- a/src/dftbp/solvation/gbsafile.F90
+++ b/src/dftbp/solvation/gbsafile.F90
@@ -199,7 +199,7 @@ contains
 
 
   !> Read a whole line from a formatted IO unit
-  subroutine nextLine(lineReader, line, lineno, file, iostat, node)
+  subroutine nextLine(lineReader, line, lineno, file, node)
 
     !> Line reader able to provide the next line from an open file
     type(TLineReader), intent(inout) :: lineReader
@@ -211,13 +211,10 @@ contains
     integer, intent(inout) :: lineno
 
     !> Name of the parametrisation file
-    character(len=*), intent(in), optional :: file
+    character(len=*), intent(in) :: file
 
-    !> Node for error handling
-    type(fnode), pointer, optional :: node
-
-    !> Error code
-    integer, intent(out), optional :: iostat
+    !> Node (needed for generating error messages)
+    type(fnode), pointer, intent(in) :: node
 
     integer :: iErr
     character(len=lc) :: errorStr, iomsg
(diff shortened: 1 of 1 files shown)


src/dftbp/solvation/gbsafile.F90
```diff
@@ -216,62 +219,26 @@ subroutine nextLine(unit, line, lineno, file, iostat, node)
...
 
-    read(unit, '(a)', advance='no', iostat=iErr, iomsg=iomsg, size=length) line
-    if (length >= len(line)) then
+    if (present(iostat)) then
+      iostat = iErr
+    else if (iErr /= 0) then
       if (present(file)) then
-        write(errorStr, '(a, "(", i0, "):", 1x, a)') &
-            & trim(file), lineno, "too many characters, line truncated"
+        write(errorStr, '(a, "(", i0, "):", 1x, a)') trim(file), lineno,&
```
bhourahine, line 232: How about a more informative message, something like:
While reading file XXX, an error was encountered on line YYY
?

src/dftbp/solvation/gbsafile.F90
```diff
@@ -216,62 +219,26 @@ subroutine nextLine(unit, line, lineno, file, iostat, node)
...
+            & "encountered error while reading line"
       else
         write(errorStr, '(a, 1x, i0, ":", 1x, a)') &
-            & "line", lineno, "too many characters, line truncated"
+           & "line", lineno, "encountered error while reading line"
```
bhourahine, line 236: Likewise:
Encountered an error when reading line YYY

src/dftbp/solvation/gbsafile.F90
```diff
@@ -216,62 +219,26 @@ subroutine nextLine(unit, line, lineno, file, iostat, node)
...
 
     lineno = lineno + 1
+    call lineReader%readLine(line, iErr)
 
-    read(unit, '(a)', advance='no', iostat=iErr, iomsg=iomsg, size=length) line
-    if (length >= len(line)) then
+    if (present(iostat)) then
+      iostat = iErr
+    else if (iErr /= 0) then
       if (present(file)) then
```
bhourahine, line 231: Am I missing something, or is file always present in calls to this routine?

bhourahine (commented):
Seems correct, but not sure if there is anything weird about the parameter files.

src/dftbp/solvation/gbsafile.F90
aradi: Fair enough. I've reduced the routine now to the bare minimum according to the actual use cases.

aradi:
Should be fine now. Changed error message and eliminated all superfluous code path in `nextLine()`

bhourahine (approved):
Other than updating and possibly squashing, could you add a comment in the commit log to the effect that the nexline routine now only reads from a named file, as was never called with a unit.

PR #1478, opened 2024-06-03
Fix restart of general k-point hybrid calculations

vanderhe:
While bandstructure calculations (k-point sampling changes as part of the restart) worked properly, simply restarting runs without any changes to the k-point sampling have erroneously been flagged as being converged already in the first iteration, due to a bug in the logic of calculating the SCC error.

The restart is implemented correctly, but the regression tests so far did not check this case as there is only a single iteration performed after the restart (that works as intended). I added another testcase that (while not being 100% safe on its own) supplements the existing tests and covers this code path.


test/app/dftb+/hybrid/periodic/Acene-kpts-matrix-restart-I/dftb_in.hsd
```diff
@@ -0,0 +1,63 @@
+Geometry = GenFormat {
```
bhourahine, line 1: Am I misunderstanding, or is this file not removed in testrun.sh before any calculation starts?

test/app/dftb+/tests
```diff
@@ -128,7 +128,8 @@ hybrid/periodic/Acene-gamma-neigh-force-cam           #? not WITH_MPI
 hybrid/periodic/Acene-gamma-neigh-force-hyb           #? not WITH_MPI
 hybrid/periodic/Acene-gamma-neigh-force-lc            #? not WITH_MPI
 hybrid/periodic/Acene-kpts-matrix-lc                  #? MPI_PROCS <= 1
-hybrid/periodic/Acene-kpts-matrix-restart             #? MPI_PROCS <= 1
+hybrid/periodic/Acene-kpts-matrix-restart-I           #? MPI_PROCS <= 1
+hybrid/periodic/Acene-kpts-matrix-restart-II          #? MPI_PROCS <= 1
```
bhourahine, line 132: Might be useful to have a comment somewhere why there are two separate versions of this test (either here, in the git log or ???) or rename by mixer type instead of I and II.

bhourahine (approved):
Generally fine, but see comments. Flags up again that the band-structure mechanism isn't very robust.

PR #1496, opened 2024-06-26
Removed direct LAPACK calls

thomas-niehaus:
Echanged direct calls to LAPACK by the provided wrappers.

### DIFF
--- a/src/dftbp/timedep/linrespcommon.F90
+++ b/src/dftbp/timedep/linrespcommon.F90
@@ -2846,7 +2846,7 @@ contains
 
 
   !> Calculate square root and inverse of sqrt of a real, symmetric positive definite matrix.
-  subroutine calcMatrixSqrt(matIn, spaceDim, memDim, workArray, workDim, matOut, matInvOut)
+  subroutine calcMatrixSqrt(matIn, spaceDim, memDim, matOut, matInvOut)
 
     !> Matrix to operate on
     real(dp), intent(in) :: matIn(:,:)
@@ -2862,12 +2862,6 @@ contains
 
     !> Inverse of matrix square root
     real(dp), intent(out) :: matInvOut(:,:)
-
-    !> Workspace array
-    real(dp), intent(out) :: workArray(:)
-
-    !> Size of work array
-    integer :: workDim
 
     real(dp) :: dummyEV(spaceDim)
     real(dp) :: dummyM(spaceDim, spaceDim), dummyM2(spaceDim, spaceDim)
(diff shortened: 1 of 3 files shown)


thomas-niehaus:
Final version.

thomas-niehaus:
Compiled and tested on dpdevel with gcc without errors. How can I get the config and compiler settings of the failing builds?

src/dftbp/timedep/linrespcommon.F90
```diff
@@ -2862,37 +2863,27 @@ subroutine calcMatrixSqrt(matIn, spaceDim, memDim, workArray, workDim, matOut, m
...
+    call heev(dummyM, dummyEV, 'U', 'V') 
 
     ! Calc. sqrt
     do ii = 1, spaceDim
       dummyM2(:,ii) = sqrt(dummyEV(ii)) * dummyM(:,ii)
     end do
 
-    call dgemm('N', 'T', spaceDim, spaceDim, spaceDim, 1.0_dp, dummyM2, spaceDim, dummyM,&
-        & spaceDim, 0.0_dp, matOut, memDim)
+    call gemm(matOut(:spaceDim,:spaceDim), dummyM2, dummyM, transB='T')
```
bhourahine, line 2879: Probably better to extend the gemm interface to include setting LDA for the matrix, as the change as is creates a temporary copy.

bhourahine:
@thomas-niehaus Look in /home/buildbot/testers/dftbplus/environments/ for specific instances with env.sh and arch.cmake files.

thomas-niehaus:
Hi Ben, on dpdevel I do not see any predefined gfortran environements. The  (ubuntu-latest, nompi, Debug, 13) is for example not available.

src/dftbp/timedep/linrespcommon.F90
thomas-niehaus: You are right. I completely redesigned the weird memory setup in Stratmann. Now temp arrays are avoided and arrays are allocated to their required size. This simplifies a lot of gemm calls. Code is 100 % faster in the limited cases I studied and does not need more memory.

thomas-niehaus:
I modified heev/dsyev to leave exit to the calling routine with an optional info parameter. For TD-DFTB info contains important information. If this is agreed on, I will also modify the other ssyev... routines.

bhourahine:
@thomas-niehaus What order of merging is this with respect to [#1515](https://github.com/dftbplus/dftbplus/pull/1515)?  This PR already partially adds info, so is there any conflicts?

thomas-niehaus:
The order is chronological. This PR is based on main, #1507 is based on this one, #1511 is based on #1507. To minimize conflicts they should be merged in this order. This one first.

aradi:
Supseeded by #1522

PR #1500, opened 2024-07-04
Clean up hybrid tests

vanderhe:
I guess a lot of users actually consult the regression tests when creating their input files. To this end, I cleaned up the hybrid test files and established a more consistent formatting.

Trivial PR, cosmetic changes only.


bhourahine (approved):
Seems OK.

PR #1513, opened 2024-08-12
Clean up short-gamma related modules

vanderhe:


src/dftbp/dftb/shortgammafuncs.F90
```diff
@@ -174,47 +169,44 @@ function expGammaPrime(rab,Ua,Ub)
     if (rab < 0.0_dp) then
       write(errorString, "('Failure in short-range gamma, r_ab negative :', f12.6)") rab
       call error(errorString)
-    else if (Ua < MinHubTol) then
+    elseif (Ua < MinHubTol) then
```
aradi, line 172: Do not change that (also not further doen), the offical (standard) way of writing it is `else if` (See Modern Fortran explained: The `else if` statement :wink)

vanderhe: I see :+1:. Shall we eventually add this to our style guide then?

vanderhe: done

aradi (approved):

PR #1515, opened 2024-08-12
Optional info parameter

thomas-niehaus:
All routines in eigensolver.F90 now have an optional parameter info. If this is present and info\=0 a quick return is done. Error handling is left to the caller. 

There are two issues: Sometimes several LAPACK subroutines are called, but info gives no information which of these routines caused the error. This could be solved by having another optional parameter or hard coding the routine information into info, like info +100 for error in the first, info + 200 for error in the second LAPACK call, etc. Not sure if this can be done in a unique way. 

Second issue: The magma routines already contain a similar parameter (err), which I don't understand (its left at zero if present).

Let me know...

### DIFF
--- a/src/dftbp/math/eigensolver.F90
+++ b/src/dftbp/math/eigensolver.F90
@@ -93,7 +93,7 @@ module dftbp_math_eigensolver
 contains
 
   !> Real eigensolver for a symmetric matrix
-  subroutine real_ssyev(a,w,uplo,jobz)
+  subroutine real_ssyev(a,w,uplo,jobz,info)
 
     !> contains the matrix for the solver, returns as eigenvectors if requested (matrix always
     !> overwritten on return anyway)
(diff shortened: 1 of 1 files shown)


bhourahine:
@thomas-niehaus The err flag is the same as info for the underlying magma library calls. It was done before the error return macros were set up, so the DFTB+ side should be updated to use the @:RAISE_ERROR macro mechanism.

thomas-niehaus:
So should I remove the 'err' parameter then? Doesn't this break existing code?

src/dftbp/math/eigensolver.F90
```diff
@@ -1813,7 +2062,7 @@ end subroutine dblecmplx_zhegvr
 #:for DTYPE, VPREC, VTYPE, NAME in [('real', 's', 'real', 'sgeev'), ('dble', 'd', 'real', 'dgeev')]
 
   !> Simple general matrix eigensolver
-  subroutine ${DTYPE}$_${NAME}$(a, wr, wi, vl, vr, err)
+  subroutine ${DTYPE}$_${NAME}$(a, wr, wi, vl, vr, err, info)
```
aradi, line 2065: The variables `err` and `info`  are redundant. We should eliminate `err`, so that the routine interface is similar to all other eigensolver wrapper interfaces.

thomas-niehaus: Simply replacing err by info will not work. 
Can you have a look at the routine? In case err is not present the value of err is undefined. What is 
 @:FORMATTED_ERROR_HANDLING(err, info_,
exactly doing? What is the function of the first parameter in this call?

aradi:
This is eventually superseded by #1662, so do not merge it yet.

aradi:
Closing as #1662 had been merged.

PR #1511, opened 2024-08-06
Td derived type

thomas-niehaus:
Massive refactoring of Casida. Introduced a new derived type that holds all array limits, number of excitations, index arrays and so on. In contrast to the existing derived type TLinResp only info that becomes available at runtime is included. The subroutine calls are now no longer cluttered. 

Running a few tests, the derived type leads to a performance loss of 10% (Intel 2021.3.0), but I think code readability is more important. 

Closes #1507
Closes #1496

PR #1520, opened 2024-08-14
Atq trans charges

thomas-niehaus:
Reintroduced ATQ.DAT which writes the transition charges for a given excited state. Now available also for non-hybrid and spin-polarized calculations. A test job has been added and the documentation modified.


aradi:
LGTM, but should be merged only once  #1511 had been merged to main.

bhourahine:
@thomas-niehaus does this need rebasing on main and squashing?

thomas-niehaus:
No, don't think so.

thomas-niehaus:
See newer PR #1532

PR #1521, opened 2024-08-16
Move modified Gram-Schmidt into math directory

bhourahine:
As this routine is useful outside of linear response routines.

Also fixed some line lengths and comment marks and added a comment to the MPI routines that block column distribution, not block cyclic, is used in the linrespgrad/linrespcommon routines for supervectors.

### DIFF
--- a/src/dftbp/math/matrixops.F90
+++ b/src/dftbp/math/matrixops.F90
@@ -10,7 +10,10 @@
 !> Simple matrix operations for which LAPACK does not have a direct call
 module dftbp_math_matrixops
   use dftbp_common_accuracy, only : dp
+  use dftbp_common_environment, only : TEnvironment
+  use dftbp_common_schedule, only : assembleChunks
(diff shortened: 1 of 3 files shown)


thomas-niehaus:
There are a couple of PR waiting and this one is based on main. The refactoring changed all calls to action* so some of your changes are obsolete. 

I fully agree that Gram-Schmidt should be general but maybe we should wait for this PR up to the merging of all other PR.

bhourahine:
@thomas-niehaus @aradi now rebased on main.

bhourahine:
@thomas-niehaus are there any outstanding PRs to merge first?

thomas-niehaus:
Let me work on ATQ first....

bhourahine:
> Let me work on ATQ first....

Sure. One thing I also noticed is that the Stratmann solver doesn't have sutable tests for the GS being broken

PR #1523, opened 2024-08-23
Buildbot fixes

aradi:
Fixes two issues experienced during buildbot testing:

* Timeout of the NAG compiler (due to a complete overload of the server)

* Failing Vsi+O-plumed test with the NAG compiler

### DIFF
--- a/config.cmake
+++ b/config.cmake
@@ -95,12 +95,8 @@ option(TEST_WITH_VALGRIND "Whether valgrind should be invoked when testing binar
 # Command line used to launch the test code.
 # The escaped variables (\${VARIABLE}) will be substituted by the corresponding CMake variables.
 if(WITH_MPI)
-  if(WITH_OMP)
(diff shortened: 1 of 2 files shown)


bhourahine (approved):
Seems fine. But neither commit log message mentions nag, so might be worth adding. Also, why is nag looking for the OMP shell variable as we build without support for that compiler, or did something change?

aradi:
> Seems fine. But neither commit log message mentions nag, so might be worth adding. Also, why is nag looking for the OMP shell variable as we build without support for that compiler, or did something change?

Well, the issue is not really NAG specific. The general problem is: building with `-DWITH_OMP=0` does not warranty that the external libraries linked are also without OMP-support. This is exactly, what happened in the NAG-case on dpdevel: the linked OpenBLAS started as many threads as cores available. So, the safest bet is to always set `OMP_NUM_THREADS` independent of the build configuration.

PR #1525, opened 2024-08-26
Fix uninitialized variable in LBFGS

aradi:
Hopefully fixes indeterministic behavior of the old LBFGS which made `GaAs_8_latopt_lbfgs` test fail in certain cases.

### DIFF
--- a/src/dftbp/include/common.fypp
+++ b/src/dftbp/include/common.fypp
@@ -136,5 +136,14 @@ block
 end block
 #:enddef
 
+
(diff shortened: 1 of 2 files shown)


src/dftbp/include/common.fypp
```diff
@@ -136,5 +136,14 @@ block
 end block
 #:enddef
 
+
```
bhourahine, line 139:
```suggestion

#! Simple macro to print (with flushing) to a unit
```

PR #1534, opened 2024-09-11
Fix concurrent file writes in MPI parallel version

aradi:
It also adds a check to detect such file I/O. Unfortunately, the check is rather unspecific, and forbids any `openFile()` calls in write mode on any ranks except the lead of `MPI_COMM_WORLD`. We should refine it later, but for current purposes it seems to be OK.

With those fixes, all buildbot tests pass on dpdevel.


test/src/dftbp/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-if(WITH_UNIT_TESTS)
+if(WITH_UNIT_TESTS AND NOT WITH_MPI)
```
bhourahine, line 1: Obviously will need to be undone in #1516

PR #1540, opened 2024-09-22
Fix for EXC.DAT not being written

bhourahine:
Also remove unused and miscommented variable


thomas-niehaus:
Ben, in which cases wasn't EXC.DAT written?

thomas-niehaus:
I see, the bug was introduced by changes in main.

PR #1541, opened 2024-09-22
Fix for common phase of subspace eigvecs

bhourahine:
MPI case for the Stratmann solver is currently diagonalising the same matrix on each rank, but then taking column slices from the resulting vectors. It may then get incompatible phases between ranks (depending on the serial eigensolver used). Work around this by broadcasting eigenvectors from the lead processor.

This bug manifested non-deterministically for gcc-13 with MKL 20.4 and shows up for hybrid/cluster/Benzene-TD-LC-Spinpol


thomas-niehaus (approved):
Great that you found it!

PR #1546, opened 2024-10-08
Hybrid gamma checks

bhourahine:
Some cleaning up of the hybrid functional initialization calls with error returns and the checks.

### DIFF
--- a/src/dftbp/dftb/sccinit.F90
+++ b/src/dftbp/dftb/sccinit.F90
@@ -14,6 +14,7 @@ module dftbp_dftb_sccinit
   use dftbp_common_globalenv, only : stdOut
   use dftbp_common_status, only : TStatus
   use dftbp_io_message, only : error
+  use dftbp_math_simplealgebra, only : diagonal
   use dftbp_type_commontypes, only : TOrbitals
   use dftbp_type_multipole, only : TMultipole
   use dftbp_dftb_densitymatrix, onLy : TDensityMatrix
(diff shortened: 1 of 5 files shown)


bhourahine:
@vanderhe Could you have a quick look at this? The missing gSummationCutoff in the initialization call seems to be a minor bug.

src/dftbp/dftb/sccinit.F90
```diff
@@ -555,8 +556,8 @@ subroutine initQFromFile(qq, fileName, tReadAscii, orb, qBlock, qiBlock, density
           else
             read(file%unit, iostat=iErr) coeffsAndShifts
           end if
-          call checkSupercellFoldingMatrix(coeffsAndShifts, errStatus,&
-              & supercellFoldingDiagOut=supercellFoldingDiag)
+          call checkSupercellFoldingMatrix(coeffsAndShifts, errStatus)
```
vanderhe, line 559: Don't we have to handle the error here in some way?

vanderhe (approved):
The missing `gSummationCutoff` in the initialization is actually the safer choice in order to obtain converged results, however, strictly spoken is a bug if the user intentionally reduced it to a non-converged value in the run that produced the restart file. So I guess your fix is correct.

src/dftbp/dftb/sccinit.F90
bhourahine: Yes (wasn't handled previous either)

vanderhe: I know :see_no_evil:, thanks for fixing it.

PR #1549, opened 2024-10-13
Symmetrises the charges in contact calculations

bhourahine:
As the two principal layers are identical, their charges should be symmetric. In the absence of code changes to just calculate one PL and duplicate it, this instead symmetrises the charge between equivalent atoms in the two layers.

Note, this will need revisiting for hybrids, but covers atomic, block and multipolar populations.

### DIFF
--- a/src/dftbp/dftbplus/initprogram.F90
+++ b/src/dftbp/dftbplus/initprogram.F90
@@ -93,6 +93,7 @@ module dftbp_dftbplus_initprogram
   use dftbp_io_commonformats, only : format2Ue
   use dftbp_io_message, only : error, warning
   use dftbp_io_taggedoutput, only : TTaggedWriter, TTaggedWriter_init
+  use dftbp_math_contactsymm, only : TEquivContactAtoms_init, TEquivContactAtoms
   use dftbp_math_duplicate, only : isRepeated
   use dftbp_math_randomgenpool, only : TRandomGenPool, init
   use dftbp_math_ranlux, only : TRanlux, getRandom
@@ -1053,11 +1054,14 @@ module dftbp_dftbplus_initprogram
     type(TNegfInt) :: negfInt
 
     !> Whether contact Hamiltonians are uploaded
-    !> Synonym for G.F. calculation of density
+    !! Synonym for G.F. calculation of density
     logical :: tUpload
 
     !> Whether a contact Hamiltonian is being computed and stored
     logical :: isAContactCalc
+
+    !> Equivalent atoms in the structure
+    type(TEquivContactAtoms), allocatable :: equivContactAtoms
 
     !> Whether Poisson solver is invoked
     logical :: tPoisson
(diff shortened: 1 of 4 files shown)


bhourahine:
@aradi Now done as a module to process the mapings

src/dftbp/dftbplus/initprogram.F90
```diff
@@ -1696,6 +1700,11 @@ subroutine initProgramVariables(this, input, env)
     this%isAContactCalc = .false.
   #:endif
 
+    if (this%isAContactCalc) then
+      allocate(this%equivalentAtoms)
+      call initTEquivalentAtoms(this%equivalentAtoms, this%nAtom)
```
aradi, line 1705:
```suggestion
      call TEquivalentAtoms_init(this%equivalentAtoms, this%nAtom)
```

src/dftbp/math/symmetrise.F90
```diff
@@ -0,0 +1,131 @@
...
+
+  private
+  public :: initTEquivalentAtoms, TEquivalentAtoms
+
+contains
+
+
+  !> Processing for transport contact calculations only at the moment, mapping atoms in two
+  !! principle layers to be identical
+  subroutine initTEquivalentAtoms(this, nAtom)
```
aradi, line 40:
```suggestion
  subroutine TEquivalentAtoms_init(this, nAtom)
```

src/dftbp/math/symmetrise.F90
```diff
@@ -0,0 +1,131 @@
...
+
+    !> Instance
+    class(TEquivalentAtoms), intent(out) :: this
```
aradi, line 43:
```suggestion
    type(TEquivalentAtoms), intent(out) :: this
```

src/dftbp/math/symmetrise.F90
```diff
@@ -0,0 +1,131 @@
...
+    do iAt = 1, nAtom/2
+      this%iMappedAtoms(1, iAt) = iAt
+    end do
+    this%iMappedAtoms(2, :nAtom/2) = this%iMappedAtoms(1, :nAtom/2) + nAtom/2
+
+  end subroutine initTEquivalentAtoms
+
+
+  !> Map (spin channel resolved) atom charges
+  subroutine map0D(this, q)
```
aradi, line 61:
```suggestion
  subroutine map0D(this, qq)
```

src/dftbp/math/symmetrise.F90
```diff
@@ -0,0 +1,131 @@
+!--------------------------------------------------------------------------------------------------!
+!  DFTB+: general package for performing fast atomistic simulations                                !
+!  Copyright (C) 2006 - 2025 DFTB+ developers group                                                !
+!                                                                                                  !
+!  See the LICENSE file for terms of usage and distribution.                                       !
+!--------------------------------------------------------------------------------------------------!
+
+!> Contains routines to symmetrize parts of a calculation
+module dftbp_math_symmetrize
```
aradi, line 9: I am not sure about the naming. Those routines are really transport/contact specific (symmetrizing the first half of the atoms with the second half). Maybe we should call this `dftbp_*_contactsymm` and `TEquivContactAtoms`?

bhourahine: In principle this could be generalized to other atom charge symmetries, but will rename for the moment.

PR #1571, opened 2024-11-23
Enforce SCC for hybrid hamiltonians

bhourahine:
Removes the option to turn off SCC when using hybrids.

Does Scc = No (i.e. the default) even make sense for hybrid calculations?

- [x] Conditionally parse Scc keyword (first commit)
- [x] Old compatibility for parser 14 and update regression tests for v14 parser
- [x] Documentation update

### DIFF
--- a/doc/dftb+/manual/dftbp.tex
+++ b/doc/dftb+/manual/dftbp.tex
@@ -1411,8 +1411,9 @@ To select the DFTB Hamiltonian you must set \is{Hamiltonian =
 
 \begin{description}
 \item[\is{SCC}] If set to \is{Yes}, a self consistent charge (SCC)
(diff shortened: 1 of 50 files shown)


vanderhe (approved):
LGTM. Are there any tests left that cover the input conversion? If yes, ready to go from my side.

bhourahine:
The older paser version tests for the non-periodic code get updated (you
did all of the periodic changes within the unreleased v14 parser, but the
older cluster ones will be updated).
One thought before merging, is it worth adding a check for <= v13 input
actually being SCC, or shall I put this into a separate PR?

bhourahine:
@vanderhe I'll do the old SCC check as a separate PR.

PR #1574, opened 2024-12-03
Update documentation of hybrids module

vanderhe:
References equations associated with the various routines.


vanderhe:
@bhourahine The DOI is not yet active, but will be in a few days.

bhourahine (approved):
LGTM, do you want to wait until the DOI goes live?

vanderhe:
> LGTM, do you want to wait until the DOI goes live?

I think we can merge it right away.

PR #1576, opened 2024-12-09
Update GitHub workflows

vanderhe:
Besides upgrading/cleaning up, this PR also proposes to run the MPI enabled case with `-DTEST_MPI_PROCS=2`, as this might catch regressions that would otherwise go unnoticed (at least until buildbot reveals them).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,6 +9,7 @@ env:
   HOMEBREW_NO_GITHUB_API: "ON"
   HOMEBREW_NO_INSTALL_CLEANUP: "ON"
   BUILD_DIR: _build
+  INSTALL_DIR: _install
   CMAKE_OPTIONS: >-
     -DWITH_API=true
     -DWITH_SDFTD3=true
@@ -26,6 +27,11 @@ jobs:
 
     runs-on: ${{ matrix.os }}
 
+    env:
+      FC: gfortran
+      CC: gcc
+      CXX: g++
+
     strategy:
       fail-fast: false
       matrix:
@@ -33,18 +39,13 @@ jobs:
         mpi: [nompi, openmpi]
         config: [Debug]
         version: [13]
-        # include:
-        #   - os: ubuntu-latest
-        #     mpi: nompi
-        #     config: Coverage
-        #     version: 13
 
     steps:
     - name: Checkout code
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
 
     - name: Set up python
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v5
       with:
         python-version: '3.x'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -107,48 +108,48 @@ jobs:
...
-        -DCMAKE_INSTALL_PREFIX=${PWD}/_install
+        cmake -B ${BUILD_DIR} -G Ninja
+        -DCMAKE_INSTALL_PREFIX=${PWD}/${INSTALL_DIR}
         -DCMAKE_BUILD_TYPE=${{ matrix.config }}
         ${CMAKE_OPTIONS}
         -DWITH_MPI=${WITH_MPI}
 
     - name: Build project
-      run: cmake --build ${BUILD_DIR}
+      run: cmake --build ${BUILD_DIR} -j 2
```
bhourahine, line 124: I'm wondering if just "-j" without number would also be OK, as the VM should be limited to 2 processes (and if it's ever changed to run on something larger, will also take advantage).

.github/workflows/build.yml
```diff
@@ -175,54 +176,49 @@ jobs:
...
-        cmake -B _build -G Ninja
-        -DCMAKE_INSTALL_PREFIX=${PWD}/_install
+        cmake -B ${BUILD_DIR} -G Ninja
+        -DCMAKE_INSTALL_PREFIX=${PWD}/${INSTALL_DIR}
         ${CMAKE_OPTIONS}
         -DWITH_MPI=${WITH_MPI}
 
     - name: Build project
-      run: cmake --build ${BUILD_DIR}
+      run: cmake --build ${BUILD_DIR} -j 2
```
bhourahine, line 218: Same comment

bhourahine (approved):
Seems OK, interested what the intel 25 will do. See minor comment.

.github/workflows/build.yml
vanderhe: I guess that depends on whether they've enabled hyperthreading. In this case the hard-coded -j 2 will (?) restrict the compilation to the two physical cores.

vanderhe:
> Seems OK, interested what the intel 25 will do. See minor comment.

While working on [Fortnet](https://github.com/vanderhe/fortnet), I've recently found a compiler bug in oneAPI 24, that forced me to move on to 25, but DFTB+ is not affected as far as I can see. Since we are not yet testing for 25 anywhere I thought it might be a good idea to test things on GitHub.

.github/workflows/build.yml
vanderhe: Seems that I'm wrong in any case, see:
https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories
They seem to have upgraded to 4 cores for public repos.

vanderhe: I have changed things accordingly...

PR #1539, opened 2024-09-20
Migrate pythonapi's setup.cfg to pyproject.toml

vanderhe:
I also reorganized the flat-/src-layout hybrid as a standard src-layout project and renamed the project from 'pythonapi' to 'dftbplus', which should not have any side-effects e.g. for the [dftbplus-python-feedstock](https://github.com/conda-forge/dftbplus-python-feedstock).

In any case, the recipes are outdated (still referring to `setup.py`), so a supplementary PR in [dftbplus/recipes](https://github.com/dftbplus/recipes) is needed.

PR #1579, opened 2024-12-12
Fix installation path of Python components

vanderhe:
My 'fix' in #1539 actually only fixed the cases where `CMAKE_INSTALL_PREFIX` is an absolute path, however broke the installation directory in the case of relative prefixes.


bhourahine (approved):
LGTM

PR #1580, opened 2024-12-13
Implement uploading output of failed tests

aradi:
Currently only for the GNU compiler.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -131,20 +131,20 @@ jobs:
           echo "TEST_OMP_PROCS=2" >> $GITHUB_ENV
         fi
         echo "TEST_UUID=$(uuidgen)" >> ${GITHUB_ENV}
-        ctest -j 2 --output-on-failure --test-dir ${BUILD_DIR}
+        ctest -j ${TEST_OMP_PROCS} --output-on-failure --test-dir ${BUILD_DIR}
 
-    - name: Archive failed test output
+    - name: Archive any failed test output
       if: failure()
       run: |
         ./utils/test/collect_failed_output.sh ${BUILD_DIR} ${BUILD_DIR}/failed-tests.${TEST_UUID}.tar
 
-    - name: Upload archive of failed test output
+    - name: Upload the archive of failed test output
       if: failure()
       uses: actions/upload-artifact@v4
       with:
         name: failed-tests.${{ env.TEST_UUID }}.tar
         path: ${{ env.BUILD_DIR }}/failed-tests.${{ env.TEST_UUID }}.tar
-        retention-days: 3
+        retention-days: 2
 
     - name: Install project
       run: cmake --install ${BUILD_DIR}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -130,9 +130,21 @@ jobs:
         else
           echo "TEST_OMP_PROCS=2" >> $GITHUB_ENV
         fi
-        pushd ${BUILD_DIR}
-        ctest -j ${TEST_OMP_PROCS} --output-on-failure
-        popd
+        echo "TEST_UUID=$(uuidgen)" >> ${GITHUB_ENV}
+        ctest -j 2 --output-on-failure --test-dir ${BUILD_DIR}
+
+    - name: Archive failed test output
+      if: failure()
+      run: |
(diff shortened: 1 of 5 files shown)


aradi:
Note, documentation of the new script is not correct, it must be still updated to the latest changes I've made before merging...

aradi:
Now, it should be OK.

.github/workflows/build.yml
```diff
@@ -130,9 +130,21 @@ jobs:
         else
           echo "TEST_OMP_PROCS=2" >> $GITHUB_ENV
         fi
-        pushd ${BUILD_DIR}
-        ctest -j ${TEST_OMP_PROCS} --output-on-failure
-        popd
+        echo "TEST_UUID=$(uuidgen)" >> ${GITHUB_ENV}
+        ctest -j 2 --output-on-failure --test-dir ${BUILD_DIR}
+
+    - name: Archive failed test output
```
bhourahine, line 136:
```suggestion
    - name: Archive any failed test output
```

.github/workflows/build.yml
```diff
@@ -130,9 +130,21 @@ jobs:
...
+      if: failure()
+      run: |
+        ./utils/test/collect_failed_output.sh ${BUILD_DIR} ${BUILD_DIR}/failed-tests.${TEST_UUID}.tar
+
+    - name: Upload archive of failed test output
```
bhourahine, line 141:
```suggestion
    - name: Upload the archive of failed test output
```

.github/workflows/build.yml
```diff
@@ -130,9 +130,21 @@ jobs:
...
+      if: failure()
+      uses: actions/upload-artifact@v4
+      with:
+        name: failed-tests.${{ env.TEST_UUID }}.tar
+        path: ${{ env.BUILD_DIR }}/failed-tests.${{ env.TEST_UUID }}.tar
+        retention-days: 3
```
bhourahine, line 147: Would 2 be OK ? (thinking about iterative attempts to fix something that failed, which could end up archiving a lot).

utils/test/collect_failed_output.sh
```diff
@@ -0,0 +1,54 @@
...
+#
+# Collect the output of failed tests into a tarball
+#
+# Usage: collect_failed_output.sh <build_dir> <archive_name>
+#
+# build_dir: The build directory where code was built and tested with CMake
+# archive_name: The name of the tarball to create
+#
+# The program assumes that the test names in the failure log are of the form *:<test_name>
+# <test_name> is the name of the test directory below <build_dir>/test.
```
bhourahine, line 11:
```suggestion
# <test_name> is the name of the test directory that is below <build_dir>/test.
```

bhourahine (commented):
Looks generally good (see some minor comments), but is the artifact-url exposed somewhere visible on return from upload-artifact ?

bhourahine:
> Looks generally good (see some minor comments), but is the artifact-url exposed somewhere visible on return from upload-artifact ?

Just tried it and now get that the url appears on stdout.

.github/workflows/build.yml
```diff
@@ -130,9 +130,21 @@ jobs:
         else
           echo "TEST_OMP_PROCS=2" >> $GITHUB_ENV
         fi
-        pushd ${BUILD_DIR}
-        ctest -j ${TEST_OMP_PROCS} --output-on-failure
-        popd
+        echo "TEST_UUID=$(uuidgen)" >> ${GITHUB_ENV}
+        ctest -j 2 --output-on-failure --test-dir ${BUILD_DIR}
```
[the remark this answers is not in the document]
aradi: Thanks for spotting it. It is a copy & paste mistake I've made when resolving the conflicts with main.

.github/workflows/build.yml
aradi: Sure, I've changed it.

aradi:
> Looks generally good (see some minor comments), but is the artifact-url exposed somewhere visible on return from upload-artifact ?

Yes, as you've found out, it it only accessible via the link shown in the stdout. Probably github does not create the menu for artifacts, if the job failed (just guessing). But we need stdout anyway to find out the uuid I use to prevent name conflicts...

bhourahine (approved):
LGTM

PR #1584, opened 2024-12-21
Fix CI for private repos

vanderhe:
Sets the number of processes based on the actual resources available.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,6 +44,17 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v4
 
+    - name: Get the number of available processes
+      run: |
+        NPROC=$(nproc)
+        echo "NPROC=$NPROC" >> $GITHUB_ENV
+        if [ ${NPROC} == 1 ]; then
+          NPROC_HALF=1
+        else
+          NPROC_HALF=$((NPROC / 2))
+        fi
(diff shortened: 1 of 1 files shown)


vanderhe:
@bhourahine Does this fix the issue you have encountered?

.github/workflows/build.yml
```diff
@@ -44,6 +44,13 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v4
 
+    - name: Get the number of available processes
+      run: |
+        NPROC=$(nproc)
+        echo "NPROC=$NPROC" >> $GITHUB_ENV
+        NPROC_HALF=$((NPROC / 2))
```
bhourahine, line 51: What happens if NPROC happens to be 1? Probably needs a ternary to make sure NPROC_HALF is not set to 0.

bhourahine (approved):
Looks OK, but one edge case to trap for.

.github/workflows/build.yml
bhourahine: @vanderhe oh, and it ran OK for the private repo.

vanderhe: I assumed that GitHub never downgrades its runners to a single process, but let me deal with this edge case.

vanderhe: @bhourahine The updated version should now correctly handle the `NPROC=1` case.

bhourahine (approved):
LGTM

PR #1586, opened 2024-12-22
Reciprocal Ewald optimization

vanderhe:
@terminationshock Or does this prevent a vectorization of the outer loop with some compilers?

I did a quick and dirty benchmark with `maxGEwald = maxGEwald + 200_dp` on my machine and found that the proposed version actually performs better...

### DIFF
--- a/src/dftbp/dftb/coulomb.F90
+++ b/src/dftbp/dftb/coulomb.F90
@@ -2034,41 +2034,39 @@ contains
   !> Returns the reciprocal part of the Ewald sum.
   function ewaldReciprocal(rr, gVec, alpha, vol) result(recSum)
 
-    !> Vector where to calculate the Ewald sum.
+    !> Vector where to calculate the Ewald sum
     real(dp), intent(in) :: rr(:)
 
-    !> Reciprocal space vectors to sum over (Should not contain either origin nor inversion related
-    !> points).
+    !> Reciprocal space vectors to sum over
+    !! (Should not contain either origin nor inversion related points).
     real(dp), intent(in) :: gVec(:,:)
 
-    !> Parameter for the Ewald summation.
+    !> Parameter for the Ewald summation
     real(dp), intent(in) :: alpha
 
-    !> Volume of the real space unit cell.
+    !> Volume of the real space unit cell
     real(dp), intent(in) :: vol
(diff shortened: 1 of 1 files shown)


terminationshock:
> Or does this prevent a vectorization of the outer loop with some compilers?

I am pretty sure that the `iG` loop will not vectorize anymore with this change, because you are computing the sum in that loop directly. Maybe it is faster because the array allocation is not needed in your version.

vanderhe:
> > Or does this prevent a vectorization of the outer loop with some compilers?
> 
> I am pretty sure that the `iG` loop will not vectorize anymore with this change, because you are computing the sum in that loop directly. Maybe it is faster because the array allocation is not needed in your version.

I see, thanks Tobias. Is there any reason to keep the current version, if the proposed one is (possibly) faster and has a smaller memory footprint? Could you maybe verify that the non-vectorized version is actually faster, also on your machine?

terminationshock:
> Could you maybe verify that the non-vectorized version is actually faster, also on your machine?
On the Raven system @ MPCDF (Intel Xeon CPUs) with 4 MPI tasks computing the SiC super cell that Bálint has sent me some time ago, I see that your code is 1.5% **slower**. That is probably not very significant, but it is at least not faster in my setup.

vanderhe:
> > Could you maybe verify that the non-vectorized version is actually faster, also on your machine?
> 
> On the Raven system @ MPCDF (Intel Xeon CPUs) with 4 MPI tasks computing the SiC super cell that Bálint has sent me some time ago, I see that your code is 1.5% **slower**. That is probably not very significant, but it is at least not faster in my setup.

@terminationshock Thanks for testing. Interesting that the outcomes are so different for different benchmarks/machines. I guess my artificial test with `maxGEwald = maxGEwald + 200_dp` is also not a good representation of real systems after all.

bhourahine:
Instead of explicitly allocating recSumArray, would it be more likely to end up in the stack memory with 
real(dp) :: recSumArray(size(gVec, dim=2)))
?
If we can work out the CPU level of parallelism, we could also size the recsumArray appropriately and pre-process to unroll the loop explicitly with Fypp.
Or of course change to a different algorithm than Ewald...

PR #1587, opened 2024-12-25
MPI parallel Waveplot

vanderhe:
To do:
- [x] parallelize over (spin, state, $k$-point) composite index
- [x] CHANGELOG entry
- [x] manual testing (as there are no regression tests)

**Caveat:**
Depending on the ordering of the composite index, backwards compatibility of the file 'wp-spinpol.cube' is only ensured up to a sign change.

**To discuss:** file I/O


app/waveplot/waveplot.F90
```diff
@@ -237,135 +236,142 @@ program waveplot
...
-        write (comments(2), 9992) wp%input%identity, iSpin, iKPoint, iLevel
+        buffer(:,:,:) = aimag(gridValCmpl)
+        write(comments(2), 9992) wp%input%identity, iSpin, iKPoint, iLevel
 9992    format('Calc-Id:',I11,', Spin:',I2,', K-Point:',I6,', State:',I6, ', imag')
         fileName = "wp-" // i2c(iSpin) // "-" // i2c(iKPoint) // "-" // i2c(iLevel) // "-imag.cube"
         call writeCubeFile(wp%input%geo, wp%aNr%atomicNumbers, wp%loc%gridVec, wp%opt%gridOrigin,&
-            & buffer, fileName, comments, wp%opt%repeatBox)
-        write(stdout, "(A)") "File '" // trim(fileName) // "' written"
+            & buffer, fileName, comments=comments, repeatBox=wp%opt%repeatBox)
+        write(*, "(A)") "File '" // trim(fileName) // "' written"
```
bhourahine, line 319: Why not stdout? Also, should the lead proc be writing this only?

app/waveplot/gridcache.F90
```diff
@@ -41,7 +45,7 @@ module waveplot_gridcache
     complex(dp), allocatable :: eigenvecCmpl(:,:)
 
     !> Parameters of the levels
-    integer, allocatable :: levelIndex(:,:)
+    integer, allocatable :: lvlIdx(:,:)
```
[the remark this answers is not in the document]
vanderhe: Was a bit less annoying to type (even with autocomplete), just less verbose, no other reason.

app/waveplot/waveplot.F90
vanderhe: Because each MPI rank owns different entries of the composite index and there should be a message printed for each index processed. This also means that all processes write their own stuff (individual grids are not distributed at the moment), which collides with the logic in `file.F90` (see the changes in `writeCubeFile()`, that we have to discuss).

vanderhe: https://github.com/dftbplus/dftbplus/blob/6002d2ec8ebcfc0ce2cba2db8b8e09208d10dd49/src/dftbp/common/file.F90#L285-L299

This code currently blocks I/O with ranks other than the lead process, which collides with what I'm trying to achieve in Waveplot. What I'm not aware of is if my current strategy is actually dangerous. To me it is the logical thing to do, should the filesystem be able to handle this in a save manner and should we be fine with not distributing individual grids. My opinion on this is that parallelizing over the composite index only is elegant for both small (many k-points) and large (many states) systems. Should we ever reach the point where this strategy (in combination with `NrOfCachedGrids=1`) is too memory consuming, we are probably screwed anyway.

bhourahine: Could you collect the messages and either the error state (MPI_LAND ?) or the proc numbers in error state on the global lead processor and print there before a global abort and stop?

app/waveplot/gridcache.F90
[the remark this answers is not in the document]
bhourahine: Could you make a search and replace at the end to something more informative?

bhourahine:
@vanderhe Stray thought, is there anything specific to hybrids that also needs to be done, reference populations?

app/waveplot/gridcache.F90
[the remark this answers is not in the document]
vanderhe: Sure can do. Do you have any suggestion on that?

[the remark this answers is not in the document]
bhourahine: Back to the original levelIndex ?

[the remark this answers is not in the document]
vanderhe: done

vanderhe:
> @vanderhe Stray thought, is there anything specific to hybrids that also needs to be done, reference populations?

@bhourahine Nothing that I'm aware of. What do you mean by reference populations?

app/waveplot/waveplot.F90
vanderhe: But collecting the messages on the lead only delays printing them, the order will be just as non-deterministic. Is this just about restricting any stdout printing to the lead process or am I missing the point here?

Regarding the error state, you suggest that the `writeCubeFile()` routine passes back the `iostat` integer and at the end of the `lpStates` loop we collect all such error states via an `MPI_LAND` operation and possibly abort? I'm just trying to understand. What do you mean by 'the proc numbers in error state'?

bhourahine: Doing this distributed means the text from difference ranks can collide on stdout (it may still be legible, depending on compiler and MPI implementation). We really should sort out MPIIO use in these cases for the actual files, but for the failed ranks this is probably better for them to wait and then the leader process announce what went wrong and start everything shutting down.

vanderhe: I'm now collecting the `iostat` and abort the program should there be an error. The messages printed to stdout should be fine, since they fit within one record, respectively. At least I didn't see any scrambled cases. Might just be a coincidence, let's discuss this afternoon.

vanderhe:
@bhourahine Rebased on main. Can you have another look at it?

vanderhe:
@bhourahine Rebased on main and removed merge commit.

bhourahine (approved):
LGTM

PR #1589, opened 2024-12-26
Clean up the Coulomb module

vanderhe:
...as I was working in this module anyway.


bhourahine (approved):
Ready to go. Not sure it's a full clean-up per-se as there is a lot of premature encapsulation in this module that eventually needs to be undone.

PR #1591, opened 2024-12-28
Remove unused shift routine

bhourahine:
As this case is not used (the block version is the commonly used one, the atom one is used for the dipole moment Hellman-Feynman test).

### DIFF
--- a/src/dftbp/dftb/shift.F90
+++ b/src/dftbp/dftb/shift.F90
@@ -24,7 +24,6 @@ module dftbp_dftb_shift
   !> Add shifts to a given Hamiltonian
   interface addShift
     module procedure addShift_atom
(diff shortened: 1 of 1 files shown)


awvwgk (approved):
Looks good to me.

One thought about the usefulness of this routine here: Having those all in one place provides a good example on how different shift dependencies can be done in DFTB+. I guess it would be better to have this in the developer documentation than in the code to reference easily.

bhourahine:
> One thought about the usefulness of this routine here: Having those all in one place provides a good example on how different shift dependencies can be done in DFTB+. I guess it would be better to have this in the developer documentation than in the code to reference easily.
I'd certainly prefer removing unused code, so in documentation would be a better location.

PR #1593, opened 2024-12-29
Waveplot superposition of atomic densities

### LINKED ISSUE #1592: Waveplot superposition of atomic densities
The file 'wp-atomdens.cube' written by Waveplot, which should contain the superposition of atomic densities, always contains garbage only.

https://github.com/dftbplus/dftbplus/blob/98e9fe0fe559b4befde94191e61203898253b530/app/waveplot/waveplot.F90#L170-L171

While `buffer` is allocated, it is not initialized with any meaningful values.

vanderhe:
Fixes #1592.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -38,6 +38,8 @@ Added
 Fixed
 -----
 
+- Incorrect superposition of atomic densities written by Waveplot
+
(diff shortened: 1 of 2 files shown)


bhourahine (approved):
Good catch.

PR #1594, opened 2024-12-29
Fix Readme download links

elv3rs:
Replaces the currently advertised stable release download page link
[https://www.dftbplus.org/download/dftb-stable/](https://www.dftbplus.org/download/dftb-stable/) (which returns 404)
with
[https://www.dftbplus.org/download/stable.html](https://www.dftbplus.org/download/stable.html)

Removes a stray backslash that was causing a fypp warning.


bhourahine (approved):
Thanks

PR #1597, opened 2024-12-30
Fix comment in file module

vanderhe:


src/dftbp/common/file.F90
```diff
@@ -424,7 +425,7 @@ subroutine openFile(fileDescr, file, options, mode, ioStat, ioMsg)
     !! * "w+": readwrite (file created or truncated if it already exists)
     !! * "a": append-write (file opened if exists, otherwise created, positioned at the end)
     !! * "a+": append-read/write (file opened if exists, otherwise created, positioned at the end)
-    !! The values above can be followed by "t" for text/unformatted mode (default) or "b" for
+    !! The values above can be followed by "t" for text/formatted mode (default) or "b" for
```
vanderhe, line 428: @bhourahine This actually is the primary reason for this PR.

PR #1600, opened 2025-01-02
Increase max. length of message string

### LINKED ISSUE #1598: Problematic fixed length char in file module
https://github.com/dftbplus/dftbplus/blob/6a03f349070f931460828fd10018bca1182f0d50/src/dftbp/common/file.F90#L285-L299

@aradi Is there any reason to restrict the `msg` char to a length of 100 characters? This recently triggered a cryptic error message for me, when working on #1587 with a path name exceeding this limit.

vanderhe:
Closes #1598.


bhourahine (approved):
We should make this properly dynamic in future.

PR #1601, opened 2025-01-02
Enable file writing for non-lead processes

vanderhe:
...as required by #1587.

### DIFF
--- a/src/dftbp/common/file.F90
+++ b/src/dftbp/common/file.F90
@@ -200,7 +200,7 @@ contains
 
 
   !> Opens a file and returns the descriptor.
-  subroutine TFileDescr_connectToFile(this, file, options, mode, ioStat, ioMsg)
+  subroutine TFileDescr_connectToFile(this, file, options, mode, parallelWriting, ioStat, ioMsg)
 
     !> File descriptor on exit (invalid if the file could not be opened and ioStat /= 0)
     class(TFileDescr), intent(out) :: this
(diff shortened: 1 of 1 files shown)


src/dftbp/common/file.F90
```diff
@@ -218,13 +218,16 @@ subroutine TFileDescr_connectToFile(this, file, options, mode, ioStat, ioMsg)
...
     !! * "a+": append-read/write (file opened if exists, otherwise created, positioned at the end)
-    !! The values above can be followed by "t" for text/unformatted mode (default) or "b" for
+    !! The values above can be followed by "t" for text/formatted mode (default) or "b" for
     !! binary/unformatted mode.
     !!
     !! When arguments options and mode are both specified, the resulting options are determined by
     !! applying options first and then set the fields which mode manipulates.
     character(*), optional, intent(in) :: mode
 
+    !> If false, blocks writing to files for all processes other than the lead (default: false).
```
bhourahine, line 228:
```suggestion
    !> If false or absent, block writing to files by all processes other than the lead (default: false).
```
Also, possibly need to make this clear that this is the global lead for MPI_COMM_WORLD, as per discussion about library use on a sub-group.

src/dftbp/common/file.F90
```diff
@@ -251,6 +255,12 @@ subroutine TFileDescr_connectToFile(this, file, options, mode, ioStat, ioMsg)
       end if
     end if
 
+    if (present(parallelWriting)) then
```
bhourahine, line 258: Would

parallelWriting_ = .false.
if (present(parallelWriting)) parallelWriting_ = parallelWriting_

Also be OK, but more concise?

bhourahine (approved):
Looks OK, but see comments.

PR #1604, opened 2025-01-07
Move nLocal* sizing out of hybrids

bhourahine:
As this is useful for non-hybrid calculations as well.


vanderhe (approved):
Makes sense :+1:.

PR #1605, opened 2025-01-09
Transition dipole moments from DeltaDFTB

timkowalczyk:
Implements evaluation of the transition dipole moment between DFTB ground state and DeltaDFTB excited state as described in Deshaye et al. JCP 158, 134104. Adds a regression test for the feature and updates the manual.


src/dftbp/dftb/determinants.F90
```diff
@@ -471,5 +496,228 @@ pure subroutine applyZieglerAlloc(eM, eT, eF)
 
   end subroutine applyZieglerAlloc
 
+  subroutine tiTDM(g, e, pt, gfilling, mfilling)
```
bhourahine, line 499:
```suggestion
  !> Evaluate transition density matrix for time independent excitation
  subroutine tiTDM(g, e, pt, gfilling, mfilling)
```

src/dftbp/dftb/determinants.F90
```diff
@@ -289,6 +310,9 @@ subroutine TDftbDeterminants_init(this, isNonAufbau, isSpinPurify, isGroundGuess
     !> Should there be a ground state initial guess before Non-Aufbau calc?
     logical, intent(in) :: isGroundGuess
 
+    !> Should the transition dipole moment with the ground state be computed?
```
bhourahine, line 313:
```suggestion
    !> Should the transition dipole moment between the excited and ground state be computed?
```

bhourahine (commented):
Just some comments, but also, what happens for periodic systems and MPI enabled calculations? We might need to block those cases for the moment.

PR #1606, opened 2025-01-09
Fix for case where a slice is passed

bhourahine:
If the lattice vector array did not contain shifts (as happens in call from initprogram) the logic was incorrect, but largely masked by the aliasing.


vanderhe (approved):
Thanks for fixing :+1:.

PR #1608, opened 2025-01-09
Fix logic for OMP target in exported dftbplus-config.cmake

aradi:
Fixes a logic bug in our cmake export file. Additionally cures the problem with missing BLAS in the GitHUB workflow.


bhourahine (approved):
LGTM

PR #1611, opened 2025-01-19
Serial real matrix adjugate

bhourahine:
* Unit tests
* Simple and also stable algorithms

Application for this is when calculating overlap/other properties between different Slater determinants. For example Eq. 15 of https://pubs.acs.org/doi/10.1021/acs.jpca.4c04139

### DIFF
--- a/src/dftbp/math/matrixops.F90
+++ b/src/dftbp/math/matrixops.F90
@@ -6,12 +6,16 @@
 !--------------------------------------------------------------------------------------------------!
 
 #:include 'common.fypp'
+#:include 'error.fypp'
 
 !> Simple matrix operations for which LAPACK does not have a direct call
 module dftbp_math_matrixops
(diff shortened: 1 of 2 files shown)


src/dftbp/math/matrixops.F90
```diff
@@ -194,4 +205,103 @@ subroutine orthonormalizeVectors(env, iStart, iEnd, vecs)
...
+    end do
+
+    U(:,:) = matmul(U, Vt)
+    eta = det(U)
+    A(:,:) = eta * A
+
+  end subroutine adjugate_stable
+
+
+  !> Evaluate adjugate of matrix from det(A) A^-1, note can be unstable even when resilting quantity
```
vanderhe, line 247:
```suggestion
  !> Evaluate adjugate of matrix from det(A) A^-1, note can be unstable even when resulting quantity
```

PR #1614, opened 2025-01-24
Skip invalid test configuration

vanderhe:
Skip invalid test configuration for `mpi_intcharges_c`, i.e. `MPI_PROCS = 3`.


bhourahine (approved):
Is this the 3 processor problem (i.e. grid shape not being square-ish)? Otherwise OK (quite a long construct for what it is doing, I wonder if we can simplify later in some way).

vanderhe:
> Is this the 3 processor problem (i.e. grid shape not being square-ish)? Otherwise OK (quite a long construct for what it is doing, I wonder if we can simplify later in some way).

Yes, it is. We could probably define a new variable such as `MPI_PROCS_ODD` and use it in combination with e.g. `MPI_PROCS <= 4`.

bhourahine:
> Yes, it is. We could probably define a new variable such as `MPI_PROCS_ODD` and use it in combination with e.g. `MPI_PROCS <= 4`.
That could be combined with (and WITH_MPI), so ONLY_MPI_PROCS_ODD /  ONLY_MPI_PROCS_EVEN ?

PR #1621, opened 2025-01-31
Abbreviate the initialization to zero in hybridxc

vanderhe:


src/dftbp/dftb/hybridxc.F90
```diff
@@ -1761,9 +1756,8 @@ subroutine checkAndInitScreening(this, matrixSize, tmpDRho)
       real(dp), intent(in), allocatable :: tmpDRho(:,:)
 
       if (.not. this%tScreeningInited) then
-        allocate(this%hprev(matrixSize, matrixSize))
+        allocate(this%hprev(matrixSize, matrixSize), source=0.0_dp)
         allocate(this%dRhoPrev(matrixSize, matrixSize))
```
bhourahine, line 1760:
```suggestion
        this%dRhoPrev = tmpDRho
```
Any reason not to use implicit allocation?

bhourahine (approved):
One comment, but otherwise LGTM

src/dftbp/dftb/hybridxc.F90
vanderhe: I'm not a huge fan of it, but otherwise there is no reason. Shall I change it in all places?

bhourahine: We can talk about it, so separate PR later for that if neccessary...

PR #1622, opened 2025-02-01
Remove some older compiler workarounds

bhourahine:


bhourahine:
Probably should wait until buildbot is back in action.

src/dftbp/dftbplus/parser.F90
```diff
@@ -6208,26 +6204,18 @@ subroutine reduceGeometry(contactVec, contactRange, lateralContactSeparation, ge
     if (geom%tPeriodic) then
       contUnitVec = contactVec / sqrt(sum(contactVec**2, dim=1))
       dots = abs(matmul(contUnitVec, geom%latVecs))
-      mask = (abs(dots - sqrt(sum(geom%latVecs, dim=1)**2)) < 1e-8_dp)
+      mask(:) = (abs(dots - sqrt(sum(geom%latVecs, dim=1)**2)) < 1e-8_dp)
       if (count(mask) /= 1) then
         call error("Too many lattice vectors parallel to the contact")
       end if
-      ! Workaround for bug in Intel compiler (can not use index function)
```
vanderhe, line 6215: Which Intel versions are actually affected?

bhourahine: 5+ years old, the original instance being in commit 3226cf2031 from 2017-10-26 and probably a mistake to call it a workaround, since that isn't what the index() function does. It's for strings, but might have been a compiler extension somewhere to work for logicals at some point.

vanderhe (approved):
LGTM, but let's wait for buildbot.

src/dftbp/dftbplus/initprogram.F90
```diff
@@ -1434,11 +1434,7 @@ subroutine initProgramVariables(this, input, env)
       call this%tblite%getOrbitalInfo(this%species0, input%slako%orb)
       allocate(input%slako%skOcc(input%slako%orb%mShell, input%geom%nSpecies))
       call this%tblite%getReferenceN0(this%species0, input%slako%skOcc)
-      ! Workaround: ifort 2021.7
```
vanderhe, line 1437: Intel 2021 is still listed under [Tested build environments](https://github.com/dftbplus/dftbplus/blob/main/INSTALL.rst#L114C1-L114C26). Is that supposed to be dropped for release 25.1?

bhourahine:
> LGTM, but let's wait for buildbot.

Everything now passes on the new buildbot setup after rebasing.

src/dftbp/dftbplus/initprogram.F90
bhourahine: I guess we need to update that, probably job for a separate PR to match currently tested environments.

PR #1624, opened 2025-02-05
Fix init of q constraints for 2 component

bhourahine:


vanderhe (commented):
Is this in anticipation of an upcoming two-component compatibility with the constraints or is the feature fully functional already? In the latter case I would suggest to add corresponding test cases for this bug fix.

bhourahine:
@vanderhe The main branch already has functioning constraints for Pauli hamiltonians (charge only), so we definitely need a test case there. This PR instead fixes a logical bug for spin constraints (but does not implement them), where only the s_x channel is initialised.

vanderhe (approved):
LGTM.

PR #1627, opened 2025-02-08
Fix the printed Meth.-Pax. order

bhourahine:
Corrects the printed Methfessel-Paxton distribution function order (was off by 2).


vanderhe (approved):
Broken indeed.
Can we also add a useful error message for `Order < 2` in the HSD input?

bhourahine:
> Can we also add a useful error message for `Order < 2` in the HSD input?

I'll try and make the parser error clearer, or are you suggesting also in initprogram?

vanderhe:
> > Can we also add a useful error message for `Order < 2` in the HSD input?
> 
> I'll try and make the parser error clearer, or are you suggesting also in initprogram?

In debug mode I'm just getting an out of bounds error, when requesting e.g. `Order = 1`. I wasn't aware that there is any useful parser error already.

bhourahine:
> In debug mode I'm just getting an out of bounds error, when requesting e.g. `Order = 1`. I wasn't aware that there is any useful parser error already.
I think you've found another bug in the routines (order 1 should be equivalent to gaussian smeared filling).

vanderhe:
> > In debug mode I'm just getting an out of bounds error, when requesting e.g. `Order = 1`. I wasn't aware that there is any useful parser error already.
> > I think you've found another bug in the routines (order 1 should be equivalent to gaussian smeared filling).

Makes sense. The check seems to be off by 1.

bhourahine:
Now fixed (MP = 0 == Gauss smearing, default for MP = 1, which is default for codes like ATK, but due to bug fix here gives same results as old buggy MP=2 request).

PR #1629, opened 2025-02-10
Avoid division by zero in screened y

### LINKED ISSUE #1628: Potentially dangerous division by zero
If `tau = 0.0`, the first line causes a (removable) singularity, which can easily be avoided when looking at the underlying expression.

https://github.com/dftbplus/dftbplus/blob/7eaa773878a536a3937d39818c859b9f9d5af8f0/src/dftbp/dftb/rshgamma.F90#L270-L271

I suggest changing the above two lines into:
```
tmp = tmp * 0.0625_dp - omega * tau**5
tmp = tmp * tau**3 / (tau**2 - omega**2)**4
```

vanderhe:
Closes #1628.


bhourahine (approved):
LGTM

PR #1630, opened 2025-02-11
Compiler fixes

aradi:
Fixes several compiler issues causing buildbot tester to fail.

This branch seems to work with gnu-12, gnu-13, gnu-14, intel-2025 and nag-72 (also with MPI).

### DIFF
--- a/src/dftbp/extlibs/magmac.c
+++ b/src/dftbp/extlibs/magmac.c
@@ -6,10 +6,17 @@
 /*------------------------------------------------------------------------------------------------*/
 
 /**
+ * Note, MAGMA header file (as of version 2.9.0) has no declaration for type bool, causing newer
+ * (more pedantic) C compilers to stop with error. Including <stdbool.h> before MAGMA seems provides
+ * the missing declaration.
+*/
+#include <stdbool.h>
+
+/**
  *  Functions to count or request GPUs
  */
 
-#include <magma.h>
+#include <magma_v2.h>
 
 /**
  *  Large enough to fit all device IDs
--- a/src/dftbp/extlibs/magmac.c
+++ b/src/dftbp/extlibs/magmac.c
@@ -7,7 +7,7 @@
 
 /**
  * Note, MAGMA header file (as of version 2.9.0) has no declaration for type bool, causing newer
- * (more pedantic) C compilers to stop with error. Including <stdbool.h> before MAGMA seems provides
+ * (more pedantic) C compilers to stop with error. Including <stdbool.h> before MAGMA provides
  * the missing declaration.
 */
 #include <stdbool.h>
--- a/src/dftbp/timedep/timeprop.F90
+++ b/src/dftbp/timedep/timeprop.F90
@@ -2474,10 +2474,13 @@ contains
     call gemm(T4R,T2R,T1R)
 
     ! build the commutator combining the real and imaginary parts of the previous result
-    !$OMP WORKSHARE
+    ! Note: parallelizing this construct via OMP WORKSHARE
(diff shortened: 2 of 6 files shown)


aradi:
OK, I've just rebased to current main, waiting on buildbot to check it again...

aradi:
Now with latest commit (f8ada91) it finally passes. :tada:

bhourahine:
> Now with latest commit ([f8ada91](https://github.com/dftbplus/dftbplus/commit/f8ada91e5783583b7af7b029d201a3da758d05b9)) it finally passes. 🎉

Not great that the solver has to be switched to get around a compiler bug (wondering if this will show up in production use with that compiler and how much impact the default will have on results). Is this a marginal tolerance fail or something significant?

src/dftbp/extlibs/magmac.c
```diff
@@ -5,11 +5,18 @@
 /*  See the LICENSE file for terms of usage and distribution.                                     */
 /*------------------------------------------------------------------------------------------------*/
 
+/**
+ * Note, MAGMA header file (as of version 2.9.0) has no declaration for type bool, causing newer
+ * (more pedantic) C compilers to stop with error. Including <stdbool.h> before MAGMA seems provides
```
vanderhe, line 10:
```suggestion
 * (more pedantic) C compilers to stop with error. Including <stdbool.h> before MAGMA provides
```
?

aradi:
Rebased (and added one more fix...)

src/dftbp/timedep/timeprop.F90
```diff
@@ -2883,36 +2886,32 @@ subroutine tdPopulInit(this, Eiginv, EiginvAdj, eigvecsReal, eigvecsCplx)
     complex(dp), allocatable :: T2(:,:), T3(:,:)
     integer :: iOrb
 
-    allocate(T2(this%nOrbs, this%nOrbs), T3(this%nOrbs, this%nOrbs))
-
+    allocate(T2(this%nOrbs, this%nOrbs))
+    allocate(T3(this%nOrbs, this%nOrbs))
     if (this%tRealHS) then
-      T2 = cmplx(eigvecsReal, 0, dp)
+      T2(:,:) = cmplx(eigvecsReal, kind=dp)
```
bhourahine, line 2892: Is this a compiler work around or just cleaning up source code?

test/app/dftb+/hybrid/periodic/GaAs-kpts-matrix-ct-np2/dftb_in.hsd
```diff
@@ -37,6 +37,8 @@ Hamiltonian = DFTB {
     0.0 0.0 0.0
   }
 
+  Solver = QR {}
```
bhourahine, line 40: Probably should comment this line that the change is due to eigensolver instability with Intel xx

bhourahine (approved):
Might as well get this merged and then come back to work out what the instability in the eigensolver actually was.
See minor comments.

bhourahine:
@aradi any reason not to merge this now? There are a couple of minor comments from me, but probably we should just do it and then come back afterwards about the ScaLAPACK behaviour.

src/dftbp/timedep/timeprop.F90
aradi: It prohibits reallocation, as `T2` is allocatable. Not sure, whether the reallocation was part of the problem, but probably it was, so I'd keep it.

aradi:
OK, I've applied the comment changes. Yes, we definitely should merge it.

PR #1631, opened 2025-02-12
Minor optimization of the HFX gamma

vanderhe:
A minor optimization of the (range-separated) HFX related $\gamma$ expressions, by storing frequently used terms.

### DIFF
--- a/src/dftbp/dftb/rshgamma.F90
+++ b/src/dftbp/dftb/rshgamma.F90
@@ -206,8 +206,7 @@ contains
     !> Resulting gamma
     real(dp) :: gamma
 
-    real(dp) :: tauA, tauB
-    real(dp) :: tmp, tau
+    real(dp) :: tmp, tau, tauA, tauB, distTauA, invDist
 
     tauA = 3.2_dp * hubbu1
     tauB = 3.2_dp * hubbu2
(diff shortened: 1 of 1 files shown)


src/dftbp/dftb/rshgamma.F90
```diff
@@ -206,8 +206,10 @@ function getHfAnalyticalGammaValue_workhorse(hubbu1, hubbu2, dist) result(gamma)
     !> Resulting gamma
     real(dp) :: gamma
 
-    real(dp) :: tauA, tauB
-    real(dp) :: tmp, tau
+    real(dp) :: tmp, tau, tauA, tauB, distTauA, invDist
+
+    distTauA = 0.0_dp
+    invDist = 0.0_dp
```
aradi, line 212: Is there any reason to initialize those variables explicitly? (Or just a kind of "good practice" in order not to leave variables uninitialized?)

src/dftbp/dftb/rshgamma.F90
```diff
@@ -456,31 +483,31 @@ pure function getYGammaSubPart(tauA, tauB, R, omega) result(yGamma)
...
+    real(dp), intent(in) :: dist
 
     !> Range-separation parameter
     real(dp), intent(in) :: omega
 
     !> Resulting off-site Y-Gamma-integral
     real(dp) :: yGamma
 
-    !!
+    !! Auxiliary variables
```
aradi, line 494: Let's simply remove this comment, as it does not state anything useful...

aradi (approved):

src/dftbp/dftb/rshgamma.F90
vanderhe: Yes, "good practice" was my intention. I decided against an initialization at the declaration, because this implies the `save` attribute (at least as far as I understand), which might lead to some unexpected behavior.

src/dftbp/dftb/rshgamma.F90
vanderhe: done

src/dftbp/dftb/rshgamma.F90
vanderhe: I have removed the added explicit initializations.

PR #1633, opened 2025-02-14
Numerically stable replacement for poly5zero

vanderhe:
My original implementation turned out to be numerically unstable, failing tests with several compiler combinations. This is a more stable drop-in replacement of `poly5zero`.

I have re-calculated four regression tests, with their respective `autotest.tag` files differing by about 1e-9-1e-7 au.

This instability might have influenced our conclusions in #1443.


bhourahine (changes requested):
Noticeably much simpler. So what to do about #1443?

bhourahine (approved):
LGTM

PR #1636, opened 2025-02-16
Match trim behaviour from 96b56a3289

bhourahine:
Warning strings now also trimmed to match handing of error strings.


vanderhe (approved):
`detailedWarning()` now consistent with `detailedError()`.

PR #1637, opened 2025-02-16
Calculate CAM gamma only if non-zero

vanderhe:
Calculate the full-range and long-range parts of the CAM $\gamma$ only if non-zero.

### DIFF
--- a/src/dftbp/dftb/rshgamma.F90
+++ b/src/dftbp/dftb/rshgamma.F90
@@ -187,9 +187,15 @@ contains
     !> Resulting CAM gamma
     real(dp) :: gamma
 
-    gamma =&
-        & camAlpha * getHfAnalyticalGammaValue_workhorse(hubbu1, hubbu2, dist)&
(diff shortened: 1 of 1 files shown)


src/dftbp/dftb/rshgamma.F90
```diff
@@ -187,9 +187,15 @@ function getCamAnalyticalGammaValue_workhorse(hubbu1, hubbu2, omega, camAlpha, c
     !> Resulting CAM gamma
     real(dp) :: gamma
 
-    gamma =&
-        & camAlpha * getHfAnalyticalGammaValue_workhorse(hubbu1, hubbu2, dist)&
-        & + camBeta * getLrAnalyticalGammaValue_workhorse(hubbu1, hubbu2, omega, dist)
+    gamma = 0.0_dp
+
+    if (abs(camAlpha) > 100_dp * epsilon(0.0_dp)) then
```
bhourahine, line 192: Any particular reason for this value?

bhourahine (approved):
Seems OK

PR #1640, opened 2025-02-21
Revise the TCI damping radius

vanderhe:
This PR changes the beginning of the TCI damping region $R_\mathrm{damp}$ from 95% of the truncation cutoff $R_\mathrm{cut}$ to a fixed value of $R_\mathrm{damp} = R_\mathrm{cut}-1.0a_0$.


src/dftbp/dftb/hybridxc.F90
```diff
@@ -525,12 +525,12 @@ subroutine THybridXcFunc_init(this, nAtom0, species0, hubbu, screeningThreshold,
...
-      ! Start beginning of the damping region and 95% of the gamma cutoff.
-      this%gammaDamping = 0.95_dp * this%gammaCutoff
+      ! Start of the damping region of the truncated Coulomb kernel
+      this%gammaDamping = this%gammaCutoff - distFudge
 
       if (this%gammaDamping <= 0.0_dp) then
-        @:RAISE_ERROR(errStatus, -1, "Beginning of damped region of electrostatics must be&
-            & positive.")
+        @:RAISE_ERROR(errStatus, -1, "The beginning of the damping region of the truncated Coulomb&
+            & kernel must be at positive radii.")
```
bhourahine, line 533:
```suggestion
            & kernel must be at a positive radius.")
```
Or is this not for a single value?

bhourahine (approved):
See comment, but otherwise OK.

src/dftbp/dftb/hybridxc.F90
vanderhe: In my mind I was evaluating the TCI kernel for several different radii and went with the plural, but your suggestion might be more intuitive.

Applied.

PR #1642, opened 2025-02-24
Switch the default HFX algo

vanderhe:
Switch the default HFX algorithm from `Thresholded` to `MatrixBased`.

@bhourahine Seems to work as intended, but I have to admit I was never quite up to speed on the whole backwards compatibility input transformation. Shall we print so kind of warning message? Is the meta-loop even necessary, since the `CAM` block was only introduced in parser version 14?


vanderhe:
@bhourahine I'm wondering how to test this. We could add a case for e.g. parser version 8 and omit the `Screening` block, which should run through the `Thresholded` path and might yield slightly different results as e.g. the MatrixBased algo (due to the default integral pre-screening setting). But this is not guaranteed, right?

src/dftbp/dftbplus/oldcompat.F90
```diff
@@ -904,7 +904,7 @@ subroutine convert_13_14(root)
         call getChildValue(ch2, "", iOrder, child=ch3)
         if (iOrder > 1) then
           write(strtmp,"(A,I0,A,I0,A)")"Older Methfessel-Paxton requested order of ", iOrder,&
-              & " is now equvalent to ", (iOrder -1), " from parser version 14."
+              & " is now equivalent to ", (iOrder -1), " from parser version 14."
```
bhourahine, line 907: Should really be a separate PR, but thanks for fixing.

bhourahine:
> @bhourahine I'm wondering how to test this. We could add a case for e.g. parser version 8 and omit the `Screening` block, which should run through the `Thresholded` path and might yield slightly different results as e.g. the MatrixBased algo (due to the default integral pre-screening setting). But this is not guaranteed, right?

The results don't have to be different, just consistent with the code before the change was made. I'd suggest version 13 parser input if there isn't already input relying on default settings.

bhourahine:
> Is the meta-loop even necessary, since the `CAM` block was only introduced in parser version 14?

If CAM only appeared in the (unrelease) v14, it doesn't need to be processed. All user calculations with it (once the feature is introduced to the userbase) will be MatrixBased.

src/dftbp/dftbplus/oldcompat.F90
vanderhe: I know... :sweat_smile:

vanderhe:
> > Is the meta-loop even necessary, since the `CAM` block was only introduced in parser version 14?
> 
> If CAM only appeared in the (unrelease) v14, it doesn't need to be processed. All user calculations with it (once the feature is introduced to the userbase) will be MatrixBased.

Yes, `CAM` only appeared in the (unreleased) v14. I have removed the meta-loop.

vanderhe:
> > @bhourahine I'm wondering how to test this. We could add a case for e.g. parser version 8 and omit the `Screening` block, which should run through the `Thresholded` path and might yield slightly different results as e.g. the MatrixBased algo (due to the default integral pre-screening setting). But this is not guaranteed, right?
> 
> The results don't have to be different, just consistent with the code before the change was made. I'd suggest version 13 parser input if there isn't already input relying on default settings.

But let's say I add a test case for v13 that does not set the `Screening` block. If in the future someone messes up the input compatibility transformation, this test might still pass. So this test case would require manual checking of `stdout` to see if the correct algorithm is used. Or am I missing something here?

test/app/dftb+/hybrid/cluster/CH3-default-compat/dftb_in.hsd
```diff
@@ -0,0 +1,48 @@
...
+    Separator = "-"
+    Suffix = ".skf"
+  }
+
+  Hybrid = LC {
+    # The new default (since parser version 14) is:
+    # Screening = MatrixBased {}
+
+    # Test the backwards compatibility of the default screening algorithm.
+    # The results obtained should be equivalent to the old default:
```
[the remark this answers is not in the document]
vanderhe: done

src/dftbp/dftbplus/oldcompat.F90
```diff
@@ -930,6 +930,17 @@ subroutine convert_13_14(root)
 
     end if
 
+    call getDescendant(root, "Hamiltonian/DFTB/Hybrid/LC", ch1)
+    if (associated(ch1)) then
+      call getDescendant(ch1, "Screening", ch2)
+      if (.not. associated(ch2)) then
+        call setChild(ch1, "Screening", ch2)
+        call setChild(ch2, "Thresholded", ch3)
```
bhourahine, line 938: Shouldn't this be MatrixBased (as the new default from v14)?

bhourahine (commented):
Did this get reverted when the preprocessor macros were removed?

bhourahine:
> ~~Did this get reverted when the preprocessor macros were removed?~~

Sorry, talking rubbish, the oldcompat is correct.

bhourahine (approved):
Ready to merge

vnquanvuong:
@vanderhe 
I found the test app/dftb+/hybrid/cluster/CH3-default-compat Failed
with the error message: MPI parallelized hybrid calculations of non-periodic systems only available for the matrix-based algorithm.

bhourahine:
> @vanderhe I found the test app/dftb+/hybrid/cluster/CH3-default-compat Failed with the error message: MPI parallelized hybrid calculations of non-periodic systems only available for the matrix-based algorithm.

The check is correct, but the test is being incorrectly run with MPI enabled on 1 processors (instead the test only being enabled for non-MPI). See #1645 for a fix.

PR #1646, opened 2025-03-03
Warning promotion of invalid escape

bhourahine:
Python 3.12 raises DeprecationWarning to SyntaxWarning for '\ ', so avoids generating these in the RELEASE.

### DIFF
--- a/test/app/dftb+/CMakeLists.txt
+++ b/test/app/dftb+/CMakeLists.txt
@@ -2,17 +2,13 @@ set(builddir ${CMAKE_CURRENT_BINARY_DIR})
 set(srcdir ${CMAKE_CURRENT_SOURCE_DIR})
 set(projectdir ${PROJECT_SOURCE_DIR})
 
-# execute_process is verbatim, therefore ";" must be replaced with " "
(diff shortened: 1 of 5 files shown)


vanderhe (commented):
The line you've changed appears in several places. Any reason to only change it in one place?

bhourahine:
> The line you've changed appears in several places. Any reason to only change it in one place?

This is the only place where the escaped space makes it into the pattern substitution (not sure why).

vanderhe:
> > The line you've changed appears in several places. Any reason to only change it in one place?
> 
> This is the only place where the escaped space makes it into the pattern substitution (not sure why).

Interesting :thinking:

bhourahine:
@vanderhe @aradi Think I've worked it out, the CMake in the api testers was adding escapes to the string as `-DRELEASE=\"'${releasename}'\"` and seems to be the only place escaping like this (the other places just use a string literal without escapes).
It's looking like there aren't any actual pattern substitutions happening inside these testers in any case, so it can be removed.

bhourahine:
The new commit removes unused CMake definitions for places where there is no fypp pattern substitution for RELEASE.

aradi (approved):

PR #1648, opened 2025-03-04
Debug failed test upload

bhourahine:
Attempt to debug why only 1 tests is uploaded on failure.


bhourahine:
@aradi fix was fairly minimal in the end. Which version of bash was expanding the array out?

bhourahine:
I suspect the original testing might have been done on a variant shell (not
actual bash), as the syntax is apparently standardized. It can't be dash
though (the one many systems set as /bin/sh) as that doesn't have arrays.

PR #1670, opened 2025-04-18
Lapack gesv routine status return

bhourahine:
Also use preprocessing, correct some comments and add missing complex case as a result.


aradi (approved):
LGTM, but let's use preprocessor variables according `TYPE(KIND)` nomenclature, similar to #1671

PR #1671, opened 2025-04-18
Serial XLBOMD solution least squares

bhourahine:
Switch from inversion of overlap to solving A X = B instead, using posv call, which assumes A is positive definite. But as this is for Choleskii factoring, hence as the generalised eigen-problem use this already to get to this point, it should be stable.

### DIFF
--- a/src/dftbp/math/lapackroutines.F90
+++ b/src/dftbp/math/lapackroutines.F90
@@ -20,7 +20,7 @@ module dftbp_math_lapackroutines
   implicit none
 
   private
-  public :: gesv, getri, getrf, sytri, sytrf, larnv, hetri, hetrf, gesvd, potrf, trsm, getrs
+  public :: gesv, getri, getrf, sytri, sytrf, larnv, hetri, hetrf, gesvd, potrf, trsm, getrs, posv
 
 
   !> Computes the solution to a real system of linear equations A * X = B, where A is an N-by-N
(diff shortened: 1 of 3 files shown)


src/dftbp/math/lapackroutines.F90
```diff
@@ -320,6 +330,89 @@ subroutine gesv_dcomplex(aa, bb, nEquation, nSolution, iError)
   end subroutine gesv_dcomplex
 
 
+#:for KIND, NAME, PRC, LABEL, PRF in [('real', 'symmetric', 'rsp', 'real', 'sposv'),&
```
aradi, line 333:
```suggestion
#:for TYPE, NAME, KIND, LABEL, PRF in [('real', 'symmetric', 'rsp', 'real', 'sposv'),&
```
In the ususal Fortran nomenclature would be `TYPE(KIND)`, so we should probably name the preprocessor variables also accordingly

aradi (approved):
Ready to go, apart of a renaming request of preprocessor-variables.

src/dftbp/math/lapackroutines.F90
bhourahine: The *tri and *trf routines in the same file do not follow that convention either, so should be I change those and document here or in another PR after this?

bhourahine: > The *tri and *trf routines in the same file do not follow that convention either, so should be I change those and document here or in another PR after this?

Looking at your comment on #1670 I think a new PR to fix those two...

PR #1694, opened 2025-05-25
Error propagation

aradi:
Lays the foundations for exception like error handling with manual propagation.

* Defines exception type containing exception event and propagation stack
* Exception event is polymorphic, so that different events can be distinguished if needed (and additional data might be passed).
* Early program termination (e.g. after H/S writing) can be propagated up using appropriate exception.

This PR implements up-propagation for the internal errors during H/S writing and the early exit after successful H/S writing as a proof of concept. This is needed in order to extend #1339 and remove the global variable `mpiGlobalComm` from the `globalenv` module.

* [ ] Exit codes might need some discussions, although only slightly relevant for the usual usage scenario of the code.

### DIFF
--- a/app/dftb+/dftbplus.F90
+++ b/app/dftb+/dftbplus.F90
@@ -9,7 +9,9 @@
 
 program dftbplus
   use dftbp_common_environment, only : TEnvironment, TEnvironment_init
-  use dftbp_common_globalenv, only : destructGlobalEnv, initGlobalEnv
+  use dftbp_common_exception, only : TEarlyExit, TException, TGeneralError, TInputError,&
+      & TInternalError, TModelError, TSystemError
+  use dftbp_common_globalenv, only : destructGlobalEnv, initGlobalEnv, stdOut0
   use dftbp_common_release, only : releaseName, releaseYear
   use dftbp_dftbplus_hsdhelpers, only : parseHsdInput
   use dftbp_dftbplus_initprogram, only : TDftbPlusMain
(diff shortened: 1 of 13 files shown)


src/dftbp/common/environment.F90
```diff
@@ -186,12 +186,20 @@ end subroutine TEnvironment_init
 
 
   !> Finalizes the environment.
-  subroutine TEnvironment_destruct(this)
+  subroutine TEnvironment_destruct(this, writeTimings)
```
bhourahine, line 189: How about isTimeWritten (1 more character, but folows the convention of indicating data type for logical variable names)?

app/dftb+/dftbplus.F90
```diff
@@ -19,24 +21,84 @@ program dftbplus
...
-  deallocate(input)
-  call runDftbPlus(main, env)
-  call main%destructProgramVariables()
-  deallocate(main)
-#:if WITH_MAGMA
-  call magmaf_finalize()
-#:endif
-  call env%destruct()
+
+  try: block
```
bhourahine, line 32: If this becomes a pattern with a try label, can you document, either in the dev guide or the doc/dftb+/code directory?

app/dftb+/dftbplus.F90
```diff
@@ -19,24 +21,84 @@ program dftbplus
...
+  end block try
+
+  call handleException(exc, exitCode)
+  call env%destruct(writeTimings=(exitCode == 0))
   call destructGlobalEnv()
+  if (exitCode /= 0) error stop exitCode, quiet=.true.
+
+contains
+
+  !> Handle eventual exception
```
bhourahine, line 59: Not sure in what sense 'eventual' means here
```suggestion
  !> Handle exception(s) that bubble up to the top of the code
```

app/dftb+/dftbplus.F90
```diff
@@ -19,24 +21,84 @@ program dftbplus
...
+  subroutine handleException(exc, exitCode)
+
+    !> Exception, might be unallocated, if no exception occured.
```
bhourahine, line 62: Seems uncertain, is it only allocated if exception occurs or might it be allocated even without exception (also typo in occurred).
```suggestion
    !> Exception, allocated if an exception occurred.
```

app/dftb+/dftbplus.F90
```diff
@@ -19,24 +21,84 @@ program dftbplus
...
+
+  call handleException(exc, exitCode)
+  call env%destruct(writeTimings=(exitCode == 0))
   call destructGlobalEnv()
+  if (exitCode /= 0) error stop exitCode, quiet=.true.
+
+contains
+
+  !> Handle eventual exception
+  subroutine handleException(exc, exitCode)
```
bhourahine, line 60: Any reason not to spell it out?
```suggestion
  subroutine handleException(exception, exitCode)
```

app/dftb+/dftbplus.F90
```diff
@@ -19,24 +21,84 @@ program dftbplus
...
+
+    withPropagationPath = .false.
+    exitCode = 0
+    if (allocated(exc)) then
+      select type (cause => exc%cause)
+      class is (TInputError)
+        write(stdOut0, "(/, a, /)") "*** Input error occured! ***"
+        exitCode = 2
+      class is (TModelError)
+        write(stdOut0, "(/, a, /)") "*** Model error occured! ***"
```
bhourahine, line 78: Meaning what? As the internal error specifically asks for a bug report, is this not to be reported (suggesting it is user error, but there is a flag for that already).

app/dftb+/dftbplus.F90
```diff
@@ -19,24 +21,84 @@ program dftbplus
...
+        exitCode = 3
+      class is (TSystemError)
+        write(stdOut0, "(/, a, /)") "*** System error occured! ***"
+        exitCode = 4
+      class is (TInternalError)
+        write(stdOut0, "(/, a, /)") "*** Internal error occured (please file a bug report)! ***"
+        exitCode = 100
+        withPropagationPath = .true.
+      ! TGeneralError is superclass of all other errors, must be checked after specific ones
```
bhourahine, line 87:
```suggestion
      ! TGeneralError is a superclass of all other errors, must be checked after the specific ones
```

app/dftb+/dftbplus.F90
```diff
@@ -19,24 +21,84 @@ program dftbplus
   implicit none
 
   type(TEnvironment) :: env
+  type(TException), allocatable :: exc
   type(TInputData), allocatable :: input
   type(TDftbPlusMain), allocatable, target :: main
+  integer :: exitCode
```
bhourahine, line 27: Need document the codes (possibly an enumerated type of them as well?). Also, what are the best conventions for the codes? The very old UNIX choices are 0, 1, 2 (see for example 2nd answer https://stackoverflow.com/questions/1101957/are-there-any-standard-exit-status-codes-in-linux ).

src/dftbp/common/exception.F90
```diff
@@ -0,0 +1,273 @@
...
+  !> Finalizer of the exception, might stop the code, if the exception is still active
+  subroutine TException_final(this)
+
+    !> Instance
+    type(TException), intent(inout) :: this
+
+    if (stopOnActiveException .and. this%active) then
+      write(stdErr, "(a)") "*** Unhandled exception went out of scope ***"
+      call this%writeTo(stdErr, withPropagationPath=.true.)
+      error stop 255
```
bhourahine, line 149: Error codes should be centralised and documented, as with dftbplus.F90 about reporting cases.

src/dftbp/dftbplus/main.F90
```diff
@@ -150,11 +151,14 @@ module dftbp_dftbplus_main
...
 
   !> The main DFTB program itself
-  subroutine runDftbPlus(this, env)
+  subroutine runDftbPlus(this, exc, env)
 
     !> Global variables
     type(TDftbPlusMain), intent(inout) :: this
 
+    !> Exception
+    type(TException), allocatable, intent(out) :: exc
```
bhourahine, line 160:
```suggestion
    type(TException), allocatable, intent(out) :: exception
```

app/dftb+/dftbplus.F90
```diff
@@ -9,7 +9,9 @@
 
 program dftbplus
   use dftbp_common_environment, only : TEnvironment, TEnvironment_init
-  use dftbp_common_globalenv, only : destructGlobalEnv, initGlobalEnv
+  use dftbp_common_exception, only : TEarlyExit, TException, TGeneralError, TInputError,&
```
bhourahine, line 12: Quite a few new ADTs, is there any argument for aggregating them into variables in a containing type? Or have a state variable inside a single type for the difference cases?

bhourahine (commented):
The type extension structuring seems to give quite a lot of types to import and the error codes need to be centrally defined and documented.

PR #1712, opened 2025-06-16
Improvements on the mdftb changes

bhourahine:
* Enable MPI runs for the various test cases
* Block some incompatible/untested combinations
* Start of external field + dipole compatibility via enabling the Helmann-Feynaman dipole tests (work in progress)


src/dftbp/dftbplus/initprogram.F90
```diff
@@ -6987,6 +6990,9 @@ subroutine ensureMdftbCompatibility(this, input)
     if (allocated(this%reks)) then
       call error("DFTB multipole expansion currently incompatible with REKS calculations")
     end if
+    if (this%isHybridXc) then
```
vnquanvuong, line 6993: Has this check already been done?

    if (allocated(input%ctrl%hybridXcInp)) then
      call error("DFTB multipole expansion currently incompatible with hybrid functionals")
    end if

vnquanvuong, line 6994: Has this check already been done?

    if (allocated(input%ctrl%hybridXcInp)) then
      call error("DFTB multipole expansion currently incompatible with hybrid functionals")
    end if

vnquanvuong: I meant

    if (allocated(input%ctrl%hybridXcInp)) then
      call error("DFTB multipole expansion currently incompatible with hybrid functionals")
    end if
and
    if (this%isHybridXc) then
      call error("DFTB multipole expansion currently incompatible with hybrid calculations")
    end if

bhourahine: Do you mean that one of these tests is redundant?

vnquanvuong: Yes, I think one of them is redundant.

bhourahine: Resolved by commit 53cc97466e131b1337e0f69617be6710d2b16c15

PR #1715, opened 2025-06-20
Fix input parsing bug in Override*BC tags

aradi:
`OverrideBulkBC = None {}` was recjected after 8e68bcb0 leading to test failure.


bhourahine (approved):
The github regressions aren't catching this, so perhaps we need a short case for that.

PR #1725, opened 2025-08-18
Remove atomic dipole double inclusion

bhourahine:
Also add (missing) dipole moments to the autotests for mdftb


vnquanvuong (approved):
Everything looks good to me.

PR #1729, opened 2025-09-23
Rewrite waveplot kernel

elv3rs:
### Main changes
- Separated molorb kernel from waveplot to allow for reuse
- Rewrote molorb kernel to employ OpenMP CPU parallelisation
- Added optional CUDA GPU offloading
- Added regression tests
- Added support for specifying arbitrary radial functions using LUTs

### What does the kernel do?
Evaluate STOs across a 3d grid.

**Regular calculation:**
Calculate the wavefunction summed over all orbitals at each grid point, for each state.
```
Psi(x,y,z,iEig) = \sum_{atoms} \sum_{species orbitals} c_{iOrbital, iEig} * sto_iOrbital(dx,dy,dz)
                = sum_{iOrbital} c_{iOrbital, iEig} * b_{iOrbital}(dx,dy,dz)
```

**Total charge calculation:**
Calculate, provided the occupation vector f_i, the total charge density at each grid point, summing over all states.
`Psi(x,y,z) = \sum_{iEig} f_i |Psi(x,y,z,iEig)|^2`

**Atomic densities calculation:**
Square the basis contributions before summation:
`Psi(x,y,z,iEig) = sum_{iOrbital} c_{iOrbital, iEig} * [b_{iOrbital}(dx,dy,dz)]^2`

### Background
Several other approaches for accelerating waveplot were investigated:
1. **Precomputed 3D LUTs with interpolation**

   - Found slower than original implementation, likely due to still requiring multiple calculations per sto evaluation and encountering frequent cache misses.

2. **Commensurable 3D LUTs (stencil approach)**

   - Memory laid out for contiguous 1D-strided access by grid distance.
   - Found significantly degraded precision at low LUT replication.
   - Higher-resolution LUTs gave acceptable accuracy but ended up slower than the exact original implementation, due to LUT count growing cubically (`n_shifted^3`), quickly exceeding cache capacity (by several GBs...)

3. **GCC OpenMP GPU offloading (nvptx)**

   - Found not production-ready -- unstable implementation, frequent crashing with cryptic error messages.

### Final approach
* Pointwise CPU calculation with OpenMP parallelisation.
* Tesserals are fast. Do not use a 3d LUT (grows cubically)
* Keep optional 1D radial-part LUT.
* Refactored tesseral harmonics to favor multiplications over divisions.
* In-place accumulation for total charge reduces memory usage/churn.
* Optional CUDA kernel for GPU acceleration.

###  Some numbers
System: 1000xH2O, grid 100^3, 0.6Bohr spacing 
`-DCMAKE_BUILD_TYPE=Release`, `/usr/bin/time` on dpdevel (1xIntel Xeon Gold 5218 96GB + 2xNVIDIA V100 16GB)

| Method     | Total charge \[s] | Single state \[s]
| ---------- | ----------------- | --------------------- |
| Original   | 364.3             | 69.2                  |
| w/o OMP    | 205.4             | 16.2                  |
| CPU        |  30.1             |  2.3                  | 
| GPU        |   7.7             |  2.2                  |
| CPU direct |  30.6             |  2.3                  |
| GPU direct |   8.3             |  2.2                  |


config.cmake
```diff
@@ -17,6 +17,9 @@ option(WITH_GPU "Whether DFTB+ should support GPU-acceleration" FALSE)
 # For serial builds, the GPU support requires the MAGMA library. For MPI parallel builds it
 # requires the ELSI library built with GPU support.
 
+option(WITH_CUDA "Whether CUDA support should be enabled" FALSE)
+# Enable offloading support for Wavegrids MO calculations
```
bhourahine:
```suggestion
# Enable GPU offloading support for waveplot via the wavegrids routines for MO calculations
```
Probably a good idea to make it very clear which codes in the build system benefit from this (otherwise we get questions about using cuda with dftb+)

config.cmake
```diff
@@ -163,6 +166,10 @@ set(HYBRID_CONFIG_METHODS "Submodule;Find;Fetch" CACHE STRING
 #     process (works also in cases where the source tree is not a Git repository)
 
 
+set(CMAKE_CUDA_ARCHITECTURES "native" CACHE STRING
+    "CUDA architectures for Wavegrid GPU acceleration (e.g. native/all/70;75;80).")
```
bhourahine:
```suggestion
    "Which CUDA architecture wavegrid GPU acceleration will be compiled for (e.g. native/all/70;75;80;...).")
```

test/src/dftbp/unit/wavegrid/basis/generate_ylm_tests.py
```diff
@@ -0,0 +1,105 @@
+""" Calculate reference values using scipy to check the Fortran spharmonics implementation """
```
bhourahine, line 1: Does this file belong in the unit test part of the source tree, or is the utils/ part of the repoitory more appropriate?

elv3rs: Im not sure.
I left it next to its output since it only functions as a fancy macro.
Since it depends on scipy, we presumably want to refrain from running it during the cmake build.

I dont think it could be applied anywhere else and thus might not have much value as a standalone tool.
Aside from someone coming along needing l=5 for some obscure reason, its mainly useful for reproducibility/trust, making it easy for anyone to verify that the ylm implementation is sound. 
Thus, the utils folder might see more traffic than the ylm tests and it could end up only being distracting clutter in there.

app/waveplot/waveplot.F90
```diff
@@ -22,9 +19,11 @@ program waveplot
   use dftbp_math_simplealgebra, only : invert33
   use dftbp_type_linkedlist, only : append, asArray, init, len, TListInt, TListRealR1
   use dftbp_type_typegeometry, only : TGeometry
+  use dftbp_wavegrid, only : getAtomicDensities, getTotalChrg
 #:if WITH_MPI
   use dftbp_extlibs_mpifx, only : MPI_LOR, MPI_SUM, mpifx_allreduceip, mpifx_bcast
 #:endif
+  use waveplot_initwaveplot, only : TProgramVariables, TProgramVariables_init
```
aradi:
```suggestion
  use waveplot_initwaveplot, only : TProgramVariables, TProgramVariables_init
#:if WITH_MPI
  use dftbp_extlibs_mpifx, only : MPI_LOR, MPI_SUM, mpifx_allreduceip, mpifx_bcast
#:endif
```
Conditional imports are always after unconditional imports.

src/dftbp/wavegrid/basis/basis.F90
```diff
@@ -0,0 +1,28 @@
+!--------------------------------------------------------------------------------------------------!
```
aradi: this file is at the wrong location: Module `dftbp_wavegrid_basis` corresponds in our convention to `src/dftbp/wavegrid/basis.F90`.

src/dftbp/wavegrid/basis/lut.F90
```diff
@@ -0,0 +1,139 @@
...
+  public :: TRadialTableOrbital, TRadialTableOrbital_initFromArray, TRadialTableOrbital_initFromOrbital
+  
+  !> Concrete class for an orbital represented by a linearly
+  !! interpolated lookup table.
+  type, extends(TOrbital) :: TRadialTableOrbital
+    !> Grid spacing (resolution)
+    real(dp) :: gridDist
+
+    !> Inverse of the grid spacing
+    real(dp) :: invLutStep
```
aradi: if it is the inverse of `gridDist` it should be probably called `invGridDist` for consistency.

src/dftbp/wavegrid/basis/lut.F90
```diff
@@ -0,0 +1,139 @@
...
+    this%gridDist = gridDist
+    this%invLutStep = 1.0_dp / gridDist
+
+    cutoff = this%gridDist * real(size(gridValue) - 1, dp)
+    this%cutoffSq = cutoff**2
+
+    allocate(this%gridValue, source=gridValue)
+
+  end subroutine TRadialTableOrbital_initFromArray
+
```
aradi: Please 2 newlines between subroutines (similar to black-formatting in Python :wink: )

src/dftbp/wavegrid/basis/spharmonics.F90
```diff
@@ -0,0 +1,135 @@
...
+  ! Normalisation Constants labeled C_l_m
+  ! l = 0
+  real(dp), parameter :: C0_0 = 0.28209479177387814_dp ! 1/2 * sqrt(1/pi)
+
+  ! l = 1
+  real(dp), parameter :: C1 = 0.4886025119029199_dp ! m =-1,0,1  1/2 * sqrt(3/pi)
+
+  ! l = 2
+  real(dp), parameter :: C2_0         = 0.31539156525252005_dp  ! m=0        1/4 * sqrt(5/pi)
+  real(dp), parameter :: C2_abs1_neg2 = 1.0925484305920792_dp   ! m=-2,-1,1  1/2 * sqrt(15/pi)
```
aradi: Suggestion, let's just list the m's it is valid for, e.g.  `C2_m2_m1_1`

src/dftbp/wavegrid/basis/spharmonics.F90
```diff
@@ -0,0 +1,135 @@
...
+  use dftbp_common_accuracy, only : dp
+  implicit none
+
+  private
+
+  public :: realTessY
+
+  ! Normalisation Constants labeled C_l_m
+  ! l = 0
+  real(dp), parameter :: C0_0 = 0.28209479177387814_dp ! 1/2 * sqrt(1/pi)
```
aradi: Maybe, we should let the compiler to calculate it in the default precision, e.g.
```suggestion
  real(dp), parameter :: C0_0 = 0.5_dp * sqrt(1.0_dp / pi)
```
etc.

src/dftbp/wavegrid/molorb/cuda/device_params.cuh
```diff
@@ -0,0 +1,347 @@
...
+            deallocate();
+            _devicePtr       = other._devicePtr;
+            _count           = other._count;
+            other._devicePtr = nullptr;
+            other._count     = 0;
+        }
+        return *this;
+    }
+
+    T*       get() { return _devicePtr; }
```
aradi: Actually, this could be also a `const` method, couldn't it?
```suggestion
    T*       get() const { return _devicePtr; }
```

src/dftbp/wavegrid/molorb/cuda/device_params.cuh
```diff
@@ -0,0 +1,347 @@
...
+    }
+    DeviceBuffer& operator=(DeviceBuffer&& other) noexcept {
+        if (this != &other) {
+            deallocate();
+            _devicePtr       = other._devicePtr;
+            _count           = other._count;
+            other._devicePtr = nullptr;
+            other._count     = 0;
+        }
+        return *this;
```
aradi: Do we ever need / use those contructors? If not, we should probably simplify the code and leave them away.

src/dftbp/wavegrid/molorb/cuda/host_logic.cu
```diff
@@ -0,0 +1,103 @@
...
+ * @return An elapsedTime_ms struct containing the time spent in kernel execution and D2H copy.
+ */
+elapsedTime_ms runBatchOnDevice(const GpuLaunchConfig& config, const GridParams* grid,
+    const SystemParams* system, const PeriodicParams* periodic, const StoBasisParams* basis,
+    const CalculationParams* calc) {
+    if (!config.z_count) return {0.0f, 0.0f, 0.0f};
+
+    CHECK_CUDA(cudaSetDevice(config.deviceId));
+
+    GpuTimer everything_timer(true), kernel_timer, d2h_timer;
```
aradi:
```suggestion
    GpuTimer total_timer(true), kernel_timer, d2h_timer;
```

PR #1736, opened 2025-10-13
MPI implementation for the excited states forces and non-adiabatic coupling vectors.

gonzalodm:
This is the first implementation of the excited-state forces and non-adiabatic coupling vector calculations performed in parallel using MPI. The code significantly improves the computational speed of both types of calculations. However, there is still a memory-related issue, as some eigenvectors are not yet distributed across the different MPI processes.

### TODO:
- [ ] Distribute the eigenvectors across the available MPI processes.
- [ ] Modify the gradient and NACV algorithms to operate on the distributed ground-state eigenvectors.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -89,3 +89,5 @@ contributed to DFTB+ :
 * Bo Thomsen (Japan Atomic Energy Agency, Japan)
 
 * Victor Yu (Duke University, USA)
+
+* Gonzalo Díaz Mirón (International Centre for Theoretical Physics, Italy)
--- a/src/dftbp/dftbplus/initprogram.F90
+++ b/src/dftbp/dftbplus/initprogram.F90
@@ -5494,11 +5494,6 @@ contains
     end if
 
     if (this%isLinResp) then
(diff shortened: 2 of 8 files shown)


bhourahine:
@gonzalodm Shall I push the [rebased version](https://github.com/gonzalodm/dftbplus/pull/2) to here? Also, are there any more changes planned from your TODO list on this PR?

bhourahine:
@gonzalodm This is now updated for the current main. Do you have any plans about the listed TODO items for distributed instead of replicated vectors, or is this for another pull request later?

bhourahine:
@gonzalodm I've just enabled the DFTB2 model NACV test case for your parallel changes, and it should pass the regressions for for up to >4 processors. The hybrid functional version of the test does not run, so was not changed.

bhourahine (approved):
I'm going to merge this as it is (passes regression testing for the furan test case for a variety of compilers). The performance changes can be done as anther pull request.

gonzalodm:
Hello Ben, thanks a lot. I will do the distribution of the eigenvectors in another PR.

PR #1757, opened 2025-11-24
SVD clean-up

bhourahine:
Added regression test for the real solver in a couple of modes.

### DIFF
--- a/src/dftbp/math/lapackroutines.F90
+++ b/src/dftbp/math/lapackroutines.F90
@@ -837,206 +837,114 @@ contains
   end subroutine larnv_dblecplx
 
 
-  !> real svd decomposition of matrix A into left and right vectors and singular values
-  subroutine sgesvd_real(A,u,sigma,vt)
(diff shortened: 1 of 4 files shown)


src/dftbp/math/lapackroutines.F90
```diff
@@ -837,206 +837,114 @@ subroutine larnv_dblecplx(iDist,iSeed,x)
...
+#!    case ('S')
+#!      @:ASSERT(all(shape(u) >= [m0,mn]))
+#!    end select
+#!    select case(jobvt)
+#!    case ('A')
+#!      @:ASSERT(all(shape(u) >= [n0,n0]))
+#!    case ('S')
+#!      @:ASSERT(all(shape(u) >= [n0,mn]))
+#!    end select
+#!  #:endblock DEBUG_CODE
```
aradi: Note: the two `select case` statements use formal arguments with the `optional` attribute. If the actual argument is not defined, it leads to segfault. Using `jobu0` and `jobvt0` should fix the segfault issue when uncommenting this block.

PR #1760, opened 2025-11-30
Partial fix of GeometryOptimisation

bhourahine:
Now follows documented behaviour according to the manual for maxsteps of -1, and prints that it is doing something (not that it is a static calculation).

Needs further future PRs to make it fully workable, starting with printing for the user what is it actually doing (algorithm choice, tolerances etc.)

### DIFF
--- a/src/dftbp/dftbplus/initprogram.F90
+++ b/src/dftbp/dftbplus/initprogram.F90
@@ -2093,6 +2093,9 @@ contains
       this%displ(:) = 0.0_dp
       this%elast = 0.0_dp
       this%nGeoSteps = input%ctrl%geoOpt%nGeoSteps
+      if (this%nGeoSteps == -1) then
+        this%nGeoSteps = huge(1)
+      end if
       this%geoOutFile = input%ctrl%geoOpt%outFile
     end if
 
@@ -3318,7 +3321,9 @@ contains
       case default
         call error("Unknown thermostat mode")
       end select
-    elseif (this%isGeoOpt) then
+
+    elseif (this%isGeoOpt .or. allocated(this%geoOpt)) then
+
       if (allocated(this%conAtom)) then
         strTmp = "with constraints"
       else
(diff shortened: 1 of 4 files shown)


src/dftbp/dftbplus/initprogram.F90
```diff
@@ -3318,7 +3321,9 @@ subroutine initProgramVariables(this, input, env)
       case default
         call error("Unknown thermostat mode")
       end select
-    elseif (this%isGeoOpt) then
+
+    elseif (this%isGeoOpt .or. allocated(this%geoOpt)) then
```
aradi, line 3325: Is there any situation, where `this%isGeoOpt` is `.false.`, but `this%geoOpt` is allocated?

src/dftbp/dftbplus/initprogram.F90
```diff
@@ -3338,6 +3343,8 @@ subroutine initProgramVariables(this, input, env)
       case (geoOptTypes%fire)
         write(stdout, "('Mode:',T30,A)") 'FIRE relaxation' // trim(strTmp)
         tGeoOptRequiresEgy = .false.
+      case (geoOptTypes%geometryoptimisation)
```
aradi, line 3346:
```suggestion
      case (geoOptTypes%geometryOptimisation)
```
However, I don't understand the logic. Why is "geometryOptimization" a specific type of geometry optimization (at the same level with CG, Fire, etc.)?

src/dftbp/dftbplus/initprogram.F90
bhourahine: Yes, unfortunately. The newer optimizers were layered on top of the code separately without any adjustment to the older structures.

src/dftbp/dftbplus/initprogram.F90
bhourahine: It was implemented totally separately from them, with its ADT having internal settings for the optimimizers it is using.

The whole thing needs to be refactored.

PR #1764, opened 2025-12-04
Upgrade tblite to version 0.5.0

aradi:
Invokes recalculation of tests with d-orbitals, as order of d-orbitals had been changed (fixed) in tblite.


bhourahine:
Release 0.6.0 is now out

PR #1771, opened 2025-12-14
Fix default behaviour for setNodeName

bhourahine:
Bug from 07976ccf43 for default behaviour in setNodeName, which led to written dftb_pin files not being usable, as the old hsd keywords were left in place after oldcompat substitutions should have happened.

Need to add regression testing example(s) to check updated dftb_pin files match behaviour of of updated dftb_in.


aradi (approved):
This will slightly affect, how American English to British English transformation will show up in the dftb_pin.hsd file (namely the British English name will appear, as previously the American English one remained). It could be changed by passing the `updateHsdNames=.false.` option in all `renameChildren()` calls, but I think, that is not really worth the effort and won't break backwards compatibility.

bhourahine:
I'll make a PR in a couple of minutes to address your comment.

PR #1774, opened 2025-12-15
Fix localization/localisation of dftb_pin

bhourahine:
Restore use of localised spellings in dftb_pin file affected by change in 6098677d7

### DIFF
--- a/src/dftbp/io/hsdutils2.F90
+++ b/src/dftbp/io/hsdutils2.F90
@@ -32,7 +32,7 @@ module dftbp_io_hsdutils2
   public :: getNodeName2, setNodeName, removeModifier, splitModifier
   public :: setUnprocessed, getDescendant, setProcessed
   public :: convertUnitHsd
-  public :: renameChildren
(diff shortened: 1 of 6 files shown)


src/dftbp/io/hsdutils2.F90
```diff
@@ -446,4 +446,22 @@ subroutine renameChildren(node, oldName, newName, updateHsdNames)
 
   end subroutine renameChildren
 
+
+  !> Alias for renameChildren, intended to be used for regional localisation spellings in the hsd
+  !> file, without substitutions to British spelling in the resulting dftb_pin file
+  subroutine localisationName(node, localName, anglicisedName)
```
aradi: Let's start with a verb, as this is a subroutine. What about `localiseName`? Or `anlgiciseName`? (Although, in theory also American English is anglicised, isn't it?)

bhourahine: LocaliseName it is...

PR #1785, opened 2026-01-09
Standalone ELPA integration

terminationshock:
Include ELPA directly without the ELSI library. Additionally implements new "redistribute" feature, which allows DFTB+ to be launched with a higher number of MPI ranks than used for the actual ELPA call. This is useful for GPU runs.
Documentation has been updated accordingly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,6 +112,18 @@ else()
   list(APPEND FYPP_FLAGS -DELSI_VERSION=0)
 endif()
 
+if(WITH_ELPA AND NOT TARGET elpa::elpa)
+  set(ELPA_MIN_VERSION "2022.05.001")
+  find_package(elpa ${ELPA_MIN_VERSION} REQUIRED)
+  if(elpa_VERSION VERSION_GREATER "2023.05.001")
+    list(APPEND FYPP_FLAGS -DELPA_HAS_SETUP_GPU)
+  endif()
+
+  if(WITH_GPU)
+    find_package(CUDAToolkit REQUIRED)
+  endif()
+endif()
+
 if(WITH_PLUMED)
   find_package(CustomPlumed REQUIRED)
   list(APPEND PKG_CONFIG_REQUIRES plumed)
--- a/src/dftbp/elecsolvers/elpa.F90
+++ b/src/dftbp/elecsolvers/elpa.F90
@@ -1,0 +1,613 @@
+!--------------------------------------------------------------------------------------------------!
+!  DFTB+: general package for performing fast atomistic simulations                                !
+!  Copyright (C) 2006 - 2025  DFTB+ developers group                                               !
+!                                                                                                  !
+!  See the LICENSE file for terms of usage and distribution.                                       !
+!--------------------------------------------------------------------------------------------------!
+
+#:include 'common.fypp'
+
+!> Contains the interface to the ELPA solver
+module dftbp_elecsolvers_elpa
+  use dftbp_common_accuracy, only : dp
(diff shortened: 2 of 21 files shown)


src/dftbp/elecsolvers/elpa.F90
```diff
@@ -0,0 +1,608 @@
...
+
+    status = this%handle%setup()
+    if (status /= ELPA_OK) then
+      call error("ELPA error: elpa_setup failed")
+    end if
+
+    if (inp%gpu) then
+      call this%setConfig("amd-gpu", 1)
+
+      status = this%handle%setup_gpu()
```
bhourahine, line 239: The build stops here unless compilation with GPUs are enabled.  Probably needs to be guarded with a preprocessor block to set this%gpu false otherwise.

src/dftbp/elecsolvers/elpa.F90
```diff
@@ -0,0 +1,608 @@
...
+    this%contextOrig = env%blacs%orbitalGrid%ctxt
+
+    call scalafx_getdescriptor(env%blacs%orbitalGrid, this%matrixSize, this%matrixSize,&
+        & env%blacs%rowBlockSize, env%blacs%columnBlockSize, this%descOrig)
+
+    if (env%blacs%rowBlockSize /= env%blacs%columnBlockSize) then
+      call error("Error during ELPA initialization: different block sizes for rows and columns")
+    end if
+
+    status = elpa_init(20250131) ! corresponds to ELPA version 2025.01.001
```
bhourahine: Is the intention to pin to this specific API version? (Unfortunately the debian one seems to, at least for current ubuntu-LTS, to be too old).

src/dftbp/elecsolvers/elpa.F90
terminationshock: I cannot reproduce the problem. `inp%gpu` is set either way and initialized as `.false.`. Could you please send me the build flags?

bhourahine: This might be an API change, as building with the latest ELPA resolves the setup_gpu symbol for a non-GPU library version.

CMakeLists.txt
```diff
@@ -82,6 +82,10 @@ else()
   set(WITH_MAGMA FALSE)
 endif()
 
+if(WITH_ELSI AND WITH_ELPA)
+  message(FATAL_ERROR "You must not enable both WITH_ELSI and WITH_ELPA")
+endif()
+
```
terminationshock: @bhourahine Implemented as discussed during the meeting.

terminationshock:
DFTB+ can now be built `WITH_ELSI` and `WITH_ELPA` at the same time. If both has been enabled, DFTB+ will prefer the direct ELPA integration at runtime. The user can overwrite this behavior with a new config file setting `PreferElsi` in the `ELPA` block.

terminationshock:
Rebased onto current main

bhourahine (approved):
LGTM

terminationshock:
I could build DFTB+ and run the tests successfully on `dpdevel` enabling `WITH_ELSI` and `WITH_ELPA`.

bhourahine:
@terminationshock could you add a note to the CHANGELOG file for this feature, then we'll merge (I don't think @aradi has raised any issues).

terminationshock:
> @terminationshock could you add a note to the CHANGELOG file for this feature, then we'll merge (I don't think @aradi has raised any issues).

Done. :heavy_check_mark: Thanks a lot!

bhourahine (approved):
LGTM

PR #1791, opened 2026-01-23
Fix for output unit setting in API

bhourahine:
Remove redundant stdOut variable inside mmapi, as apparently had no function(?)

### DIFF
--- a/src/dftbp/api/mm/mmapi.F90
+++ b/src/dftbp/api/mm/mmapi.F90
@@ -356,8 +356,6 @@ contains
     !> multiple TDftbPlus instances within an MPI-process)
     integer, intent(in), optional :: devNull
 
-    integer :: stdOut
-
   #:if not INSTANCE_SAFE_BUILD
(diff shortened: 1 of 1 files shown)


src/dftbp/api/mm/mmapi.F90
```diff
@@ -369,12 +367,6 @@ subroutine TDftbPlus_init(this, outputUnit, mpiComm, devNull)
       call error("MPI Communicator supplied to initialise a serial DFTB+ instance")
     end if
 
-    if (present(outputUnit)) then
-      stdOut = outputUnit
-    else
-      stdOut = output_unit
-    end if
-
     call initGlobalEnv(outputUnit=outputUnit, mpiComm=mpiComm, devNull=devNull)
```
bhourahine:
```suggestion
    call initGlobalEnv(outputUnit=stdOut, mpiComm=mpiComm, devNull=devNull)
```

PR #1802, opened 2026-02-18
RecomputeAfterDensity

pecchia:
I moved the existent option for the Poisson solver "RecomputeAfterDensity" to the main dftb Hamiltonian block of options. In this way the option is available in general, also when the gamma-functional is used.
Rationale:
In el-phonon scattering calculations we have published works using fixed-charges. This is obtained by setting MaxSCCIterations=1 and readInitialCharges=Yes. 
We want the Forces to be computed with the initialized charges, not those recomputed from the new DM, especially when atoms are displaced. In practice the option is useful whenever the user would like to run calculations with **fixed** atomic charges. 
The option is also useful in the context of transport+gammafunctional (to come).

### DIFF
--- a/src/dftbp/dftbplus/oldcompat.F90
+++ b/src/dftbp/dftbplus/oldcompat.F90
@@ -28,7 +28,7 @@ module dftbp_dftbplus_oldcompat
 
   !> Actual input version <-> parser version maps (must be updated at every public release)
   type(TVersionMap), parameter :: versionMaps(*) = [&
-      & TVersionMap("25.1", 14),&
+      & TVersionMap("26.1", 15), TVersionMap("25.1", 14),&
       & TVersionMap("24.1", 14), TVersionMap("23.1", 13), TVersionMap("22.2", 12),&
       & TVersionMap("22.1", 11), TVersionMap("21.2", 10), TVersionMap("21.1", 9),&
       & TVersionMap("20.2", 9), TVersionMap("20.1", 8), TVersionMap("19.1", 7),&
@@ -101,6 +101,9 @@ contains
       case (13)
         call convert_13_14(root)
         version = 14
+      case (14)
+        call convert_14_15(root)
+        version = 15
       end select
     end do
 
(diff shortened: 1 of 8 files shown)


test/app/dftb+/transport/CH4-poisson/dftb_in.hsd
```diff
@@ -33,6 +32,7 @@ Prefix = {slakos/origin/mio-1-1/}
     readoldbulkpotential = no
     boundaryregion = global {}
     poissonaccuracy = 1e-7
+    Recomputeafterdensity = No
```
bhourahine, line 35: The compatibility conversion to the 26.1 parser version should be taking care of this. Tests for old parser versions should be unchanged.

test/app/dftb+/tagdiff.conf
```diff
@@ -45,9 +45,6 @@ orbital_charges:	@ 1.0e-6
 forces:			@ 1.0e-6
 total_localdos:         @ 1.0e-5
 dEidV:                  @ 5.0e-6
-dipole_moments:         @ 5.0e-5
```
bhourahine: Original global setting should be unchanged, with test specific updates.

cmake/DftbPlusUtils.cmake
```diff
@@ -228,6 +228,9 @@ function (dftbp_ensure_config_consistency)
   endif()
 
   # Check minimal compiler versions
+  # set(fortran_minimal_versions "GNU;12.2" "Intel;20.2" "IntelLLVM;2024.2" "NAG;7.2")
```
bhourahine: As per previous discussion, this is getting outside of our support window for compilers.

src/dftbp/dftbplus/oldcompat.F90
```diff
@@ -938,6 +941,45 @@ subroutine convert_13_14(root)
 
   end subroutine convert_13_14
 
+  !> Converts input from version 14 to 15. (Version 15 August 2026)
```
bhourahine: Not sure where the date came from, probably should be date of the commit.

src/dftbp/dftbplus/oldcompat.F90
```diff
@@ -938,6 +941,45 @@ subroutine convert_13_14(root)
...
+    type(fnode), pointer :: root
+
+    type(fnode), pointer :: ch1, ch2, par, hamil, dummy
+    logical :: tVal, tPoisson=.false.
+
+    ! Move RecomputeAfterDensity to the Hamiltonian block.
+    ! The normal default is RecomputeAfterDensity=.true.
+    ! Check if Poisson block is present. If NOT RecomputeAfterDensity will take the new default. 
+    call getDescendant(root, "Hamiltonian/DFTB/Electrostatics/Poisson", ch1, parent=par)
+    if (associated(ch1)) then
```
bhourahine:
```
    tPoisson = .false.
    if (associated(ch1)) tPoisson = .true.
```

cmake/DftbPlusUtils.cmake
bhourahine: There also a merged change from PR #1808 that changes the GNU minimum version number.

src/dftbp/poisson/poisson.F90
```diff
@@ -336,7 +336,7 @@ subroutine init_poissbox(iErr)
      biasdir = abs(contdir(1))
 
      if (abs(bound(2)-bound(1)).le.(OxLength+dr_eps)) then
-       @:ERROR_HANDLING(iErr, -6, 'Gate insulator is longer than Poisson box!')
+     !  @:ERROR_HANDLING(iErr, -6, 'Gate insulator is longer than Poisson box!')
```
bhourahine: Why is this removed (the if block now does nothing)? Or should this be part of a different PR that does something else?

src/dftbp/poisson/parcheck.F90
```diff
@@ -155,6 +155,7 @@ subroutine check_parameters()
 
    if(OxLength.lt.GateLength_l) OxLength=GateLength_l
    if(Rmin_ins.lt.Rmin_gate) Rmin_ins=Rmin_gate
+   !if(Rmin_ins.le.0.0d0 .or. Rmin_ins.gt.Rmin_gate) Rmin_ins=Rmin_gate
```
bhourahine: Any reason to add a commented out line here?

test/app/dftb+/transport/CH4-poisson/dftb_in.hsd
pecchia: In fact I changed first, but then reverted in a subsequent commit.

src/dftbp/poisson/poisson.F90
pecchia: I don't know why it slipped here. It it part of another fix I am working at with Franco. The commented line should be the correct one but it is not par of this PR. I will delete the change now.

PR #1806, opened 2026-02-28
VelocityVerlet: optional forces print in XYZ  MD trajectory output

jjakowski:
* Adds PrintTrajectoryForces = Yes/No (default No) to Driver = VelocityVerlet to append forces to ${OutputPrefix}.xyz.
* When enabled, atom lines include fx fy fz as columns 9–11 (after velocities), units (Hartree/bohr) match internal forces (detailed.out).
* Independent of Analysis = { PrintForces = Yes }, which is unchanged.
* No output change when keyword is absent/No.
* Tested with MD example: verified 8 vs 11 columns and final-step force match to detailed.out.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,9 @@ Added
 - Option (FromParameters) to automatically set spin-constants for xTB
   models
 
+- VelocityVerlet: new PrintTrajectoryForces option to append fx fy fz to
+  OutputPrefix.xyz (units: eV/Angstrom).
+
 
 Changed
 -------
--- a/src/dftbp/dftbplus/initprogram.F90
+++ b/src/dftbp/dftbplus/initprogram.F90
@@ -510,6 +510,9 @@ module dftbp_dftbplus_initprogram
     !> Are forces being returned
     logical :: tPrintForces
 
+    !> Should per-atom forces be written into the MD trajectory file
+    logical :: writeTrajectoryForces
+
     !> Number of moved atoms
     integer :: nMovedAtom
 
@@ -1978,6 +1981,7 @@ contains
     this%tAppendGeo = input%ctrl%tAppendGeo
     this%isSccConvRequired = input%ctrl%isSccConvRequired
     this%tMD = input%ctrl%tMD
+    this%writeTrajectoryForces = input%ctrl%writeTrajectoryForces
     if (this%tMD) this%mdOutput = input%ctrl%mdOutput
     this%tDerivs = input%ctrl%tDerivs
     this%tPrintMulliken = input%ctrl%tPrintMulliken
--- a/src/dftbp/dftbplus/mainio.F90
+++ b/src/dftbp/dftbplus/mainio.F90
@@ -15,7 +15,8 @@
 !> Various I/O routines for the main program.
 module dftbp_dftbplus_mainio
   use dftbp_common_accuracy, only : dp, lc, mc, sc
-  use dftbp_common_constants, only : au__Debye, au__pascal, au__V_m, Bohr__AA, Boltzmann, gfac,&
+  use dftbp_common_constants, only : au__Debye, au__fs, au__pascal, au__V_m, Bohr__AA, Boltzmann,&
+      & gfac,&
       & Hartree__eV, quaternionName, spinName
   use dftbp_common_environment, only : TEnvironment
   use dftbp_common_file, only : closeFile, openFile, TFileDescr
@@ -4426,7 +4427,7 @@ contains
   !> Write current geometry to disc
   subroutine writeCurrentGeometry(geoOutFile, pCoord0Out, tLatOpt, tMd, tAppendGeo, tFracCoord,&
       & tPeriodic, tHelical, tPrintMulliken, species0, speciesName, latVec, origin, iGeoStep,&
-      & iLatGeoStep, nSpin, qOutput, velocities)
+      & iLatGeoStep, nSpin, qOutput, velocities, writeTrajectoryForces, derivs)
 
     !> File for geometry output
     character(*), intent(in) :: geoOutFile
@@ -4482,8 +4483,13 @@ contains
     !> Atomic velocities
     real(dp), intent(in), allocatable :: velocities(:,:)
 
+    !> Should per-atom forces be printed in the trajectory
+    logical, intent(in) :: writeTrajectoryForces
+
+    !> Energy derivatives with respect to atomic coordinates
+    real(dp), intent(in), allocatable :: derivs(:,:)
+
     integer :: nAtom
-    integer :: ii, jj
     character(lc) :: comment, fname
 
     nAtom = size(pCoord0Out, dim=2)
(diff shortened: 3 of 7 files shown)


src/dftbp/dftbplus/initprogram.F90
```diff
@@ -509,6 +509,9 @@ module dftbp_dftbplus_initprogram
     !> Are forces being returned
     logical :: tPrintForces
 
+    !> Should per-atom forces be written into the MD trajectory file
+    logical :: tPrintTrajectoryForces
```
bhourahine: We are gradually moving away from starting logical variables with 't', as 'T' is used to denote types.
```suggestion
    logical :: areTrajectoryForcesPrinted
```
but this is a little long.

src/dftbp/dftbplus/mainio.F90
```diff
@@ -4481,11 +4482,21 @@ subroutine writeCurrentGeometry(geoOutFile, pCoord0Out, tLatOpt, tMd, tAppendGeo
     !> Atomic velocities
     real(dp), intent(in), allocatable :: velocities(:,:)
 
+    !> Should per-atom forces be printed in the trajectory
+    logical, intent(in), optional :: tPrintTrajectoryForces
+
+    !> Energy derivatives with respect to atomic coordinates
+    real(dp), intent(in), optional :: derivs(:,:)
```
bhourahine: Is this not always present (but might not be allocated)?

src/dftbp/dftbplus/mainio.F90
```diff
@@ -4481,11 +4482,21 @@ subroutine writeCurrentGeometry(geoOutFile, pCoord0Out, tLatOpt, tMd, tAppendGeo
...
+
     integer :: nAtom
-    integer :: ii, jj
     character(lc) :: comment, fname
+    logical :: shouldPrintTrajectoryForces
 
     nAtom = size(pCoord0Out, dim=2)
+    shouldPrintTrajectoryForces = .false.
+    if (present(tPrintTrajectoryForces)) then
```
bhourahine: OK to do this as 1 line:
```suggestion
    if (present(tPrintTrajectoryForces)) shouldPrintTrajectoryForces = tPrintTrajectoryForces
```

src/dftbp/dftbplus/mainio.F90
```diff
@@ -4498,26 +4509,115 @@ subroutine writeCurrentGeometry(geoOutFile, pCoord0Out, tLatOpt, tMd, tAppendGeo
...
-    if (tPrintMulliken) then
-      ! For non-colinear spin without velocities write magnetisation into the velocity field
-      if (nSpin == 4 .and. .not. allocated(velocities)) then
-        call writeXYZFormat(fname, pCoord0Out, species0, speciesName,&
-            & charges=sum(qOutput(:,:,1), dim=1),&
-            & vectors=transpose(sum(qOutput(:,:,2:4), dim=1)), comment=comment,&
-            & append=tAppendGeo)
+    if (shouldPrintTrajectoryForces) then
+      if (.not. allocated(velocities)) then
+        call error("Internal error: trajectory force output requires velocities in MD trajectory")
```
bhourahine: If possible, please pass errors upwards to the main.F90 level with the @:RAISE_ERROR() mechanism used elsewhere (relevant for use of the code via the API).
Also in this specific case @:ASSERT() could probably be used instead.

CHANGELOG.rst
```diff
@@ -11,6 +11,8 @@ Unreleased
 Added
 -----
 
+- VelocityVerlet: new PrintTrajectoryForces option appends fx fy fz to
```
bhourahine:
```suggestion
- VelocityVerlet: new PrintTrajectoryForces option to append fx fy fz to
```

bhourahine (commented):
As the rest of the xyz is in AA and or AA/ps, probably forces should be in something other than a.u. Not sure what energy base unit is most appropriate, eV possibly?

src/dftbp/dftbplus/mainio.F90
jjakowski: Addressed in 40f431e.  I’ve updated writeCurrentGeometry so tPrintTrajectoryForces is  no longer optional  and  derivs is allocatable and tested  for allocated(derivs)

src/dftbp/dftbplus/initprogram.F90
jjakowski: why not just use "printTrajectoryForces" ?

src/dftbp/dftbplus/mainio.F90
jjakowski: Addressed in 219dcd2.  Replaced the trajectory-force call error(...) checks with @:ASSERT(...) in writeCurrentGeometry (mainio.F90).

CHANGELOG.rst
jjakowski: Addressed in db8f14a (Changelog cleanup).

src/dftbp/dftbplus/initprogram.F90
jjakowski: Done in ce188d1 . Renamed "tPrintTrajectoryForces" to "printTrajectoryForces", ).  consistently with your suggestion

src/dftbp/dftbplus/mainio.F90
jjakowski: Addressed in 40f431e. Removed  "if (present(tPrintTrajectoryForces)) then ...."

src/dftbp/dftbplus/mainio.F90
```diff
@@ -4498,26 +4504,115 @@ subroutine writeCurrentGeometry(geoOutFile, pCoord0Out, tLatOpt, tMd, tAppendGeo
...
+
+    !> Optional Mulliken charges
+    real(dp), intent(in), optional :: charges(:)
+
+    type(TFileDescr) :: fd
+    character(1) :: mode
+    integer :: ii, nAtom, nSpecies
+    real(dp) :: forceConv
+
+200 format(I5)
```
bhourahine: As these are only used once, could just put them in the relevant write statements.

src/dftbp/dftbplus/mainio.F90
```diff
@@ -4498,26 +4504,115 @@ subroutine writeCurrentGeometry(geoOutFile, pCoord0Out, tLatOpt, tMd, tAppendGeo
...
+205 format(A5,10F16.8)
+206 format(A5,9F16.8)
+
+    nAtom = size(coords, dim=2)
+    nSpecies = maxval(species)
+    forceConv = Hartree__eV / Bohr__AA
+    @:ASSERT(size(coords, dim=1) == 3)
+    @:ASSERT(size(species) == nAtom)
+    @:ASSERT(size(speciesNames) == nSpecies)
+    @:ASSERT(all(shape(velocities) == (/ 3, nAtom /)))
```
bhourahine:
```suggestion
    @:ASSERT(all(shape(velocities) == [ 3, nAtom ]))
```

src/dftbp/dftbplus/mainio.F90
```diff
@@ -4481,8 +4482,13 @@ subroutine writeCurrentGeometry(geoOutFile, pCoord0Out, tLatOpt, tMd, tAppendGeo
     !> Atomic velocities
     real(dp), intent(in), allocatable :: velocities(:,:)
 
+    !> Should per-atom forces be printed in the trajectory
+    logical, intent(in) :: printTrajectoryForces
+
+    !> Energy derivatives with respect to atomic coordinates
```
bhourahine: Could make this argument optional, and use it presence to signal whether to print

src/dftbp/dftbplus/mainio.F90
```diff
@@ -4498,26 +4504,115 @@ subroutine writeCurrentGeometry(geoOutFile, pCoord0Out, tLatOpt, tMd, tAppendGeo
...
+    real(dp), intent(in), optional :: charges(:)
+
+    type(TFileDescr) :: fd
+    character(1) :: mode
+    integer :: ii, nAtom, nSpecies
+    real(dp) :: forceConv
+
+200 format(I5)
+205 format(A5,10F16.8)
+206 format(A5,9F16.8)
```
aradi: Pls, use character constants to store reusable format strings, like
```
character(*), parameter :: formWithCharges = "(a5, 10f16.8)"
```

aradi (changes requested):
Thanks a lot. I am fine with that apart of the numeric labels for format statements. We should probably name the name the keyword `WriteTrajectoryForces` instead of `PrintTrajectoryForces`, as they are written to a file not printed to into a human readable out. (I think we do not need to be analogous to the  `PrintForces` keyword...)

bhourahine:
@jjakowski This is now merged with the other suggestions incorporated (input keyword change being the only user-facing difference).

PR #1817, opened 2026-03-24
Atom derivatives wrt MM atoms

bhourahine:
Atom derivatives wrt MM atoms
* API request for atom charge derivs wrt atomic coordinates
* API request for atom charge derivs wrt external charge coords
* Third order QM derivs wrt atom coordinates
* Unit test for coordinate derivative of potential
  Not fully working for periodic case (not used at moment)
* API tester for derivatives (WIP)
* Optionally select specific atom/charge derivatives to evaluate

Note, API init for perturbation needs logging

Co-authored-by: Tomáš Kubař <tomas.kubar@kit.edu>

Contains #1829 and #1832


tomaskubar:
I get a problem with the C-header file when compiling with G++:
error: ‘_Bool’ does not name a type
This is the return type c_bool of 2 functions, dftbp_is_instance_safe() and dftbp_is_hs_real().
The compiler does understand "bool" at this place, though. But will the value then be passed from Fortran correctly?

bhourahine:
> I get a problem with the C-header file when compiling with G++: error: ‘_Bool’ does not name a type This is the return type c_bool of 2 functions, dftbp_is_instance_safe() and dftbp_is_hs_real(). The compiler does understand "bool" at this place, though. But will the value then be passed from Fortran correctly?

This is a variant of the problem described [here](https://blog.bytellect.com/software-development/c/whats-the-difference-between-_bool-and-bool-in-c/). I'll add it to the agenda for the next DFTB+ dev meeting to discuss whether we should use the header or check for __cplusplus instead.

bhourahine:
> I get a problem with the C-header file when compiling with G++: error:
There is a fix in #1839 which will be discussed later this week before merging.

PR #1831, opened 2026-04-08
Automatically download test data via CMake FetchContent

elv3rs:
Replace the get_opt_externals script with FetchContent to automatically download slakos and gbsa test data during cmake configuration.


bhourahine (approved):
Other than adding a note to CHANGELOG.rst it looks ready to merge.

PR #1853, opened 2026-05-22
Fix WITH_API=OFF build: guard isASICallbackEnabled in ensureSolverCompatibility call

vnquanvuong:
When building with -DWITH_API=OFF, compilation fails in src/dftbp/dftbplus/initprogram.F90

This PR wraps the call so WITH_API=OFF builds pass .false. instead.


bhourahine:
It would probably be simpler to remove the proprocessing around line 155 of inputdata.F90, so that it is false by default.

vnquanvuong:
I've updated it as you suggested.

bhourahine (approved):
LGTM

PR #1854, opened 2026-05-22
Add pairwise bond population, bond energy and Mayer bond order analysis

vnquanvuong:
Adds three new keywords to the `Analysis { }` block so that pairwise bond information can be obtained from a standard SCC ground-state run, without going through the `ElectronDynamics`.

| Keyword                       | Output file        | Quantity                                                              |
|-------------------------------|--------------------|-----------------------------------------------------------------------|
| `WriteBondPopulation = Yes`   | `bond_pop.dat`     | Mulliken bond populations  `B_AB = sum_{mu in A, nu in B} P_{mu nu} S_{mu nu}` |
| `WriteBondEnergy = Yes`       | `bond_energy.dat`  | Non-SCC bond energies      `E_AB = sum_{mu in A, nu in B} P_{mu nu} H0_{mu nu}` |
| `WriteBondOrder = Yes`        | `bond_order.dat`   | Mayer bond orders          `B_AB = sum_{mu in A, nu in B} (PS)_{mu nu} (PS)_{nu mu}` |

Output is plain-text N×N matrices. If `WriteAutotestTag = Yes`, the
matrices are also appended to `autotest.tag` under the labels
`bond_populations`, `bond_energies` and `bond_orders`.


src/dftbp/dftb/bondpopulations.F90
```diff
@@ -76,4 +75,44 @@ subroutine addPairWiseBondInfo(info, rhoPrim, sparseMat, iSquare, iNeighbour, nN
 
   end subroutine addPairWiseBondInfo
 
+
+  !> Computes Mayer-style pairwise bond orders from the (dense) matrix product
+  !> PS, where P is the spin-summed density matrix and S the overlap.
```
bhourahine:
```suggestion
  !! PS, where P is the spin-summed density matrix and S the overlap.
```

src/dftbp/dftbplus/main.F90
```diff
@@ -862,6 +868,60 @@ subroutine processOutputCharges(env, this)
   end subroutine processOutputCharges
 
 
+  !> Computes pairwise bond information.
+  !> Mulliken bond populations, non-SCC bond energies, and/or Mayer bond orders.
```
bhourahine:
```suggestion
  !! Mulliken bond populations, non-SCC bond energies, and/or Mayer bond orders.
```

src/dftbp/dftbplus/mainio.F90
```diff
@@ -2038,6 +2039,85 @@ subroutine writeAutotestTag(fileName, electronicSolver, tPeriodic, cellVol, tMul
   end subroutine writeAutotestTag
 
 
+  !> Writes pairwise Mulliken bond populations, non-SCC bond energies and/or
+  !> Mayer bond orders as plain-text N x N matrices.
```
bhourahine:
```suggestion
  !! Mayer bond orders as plain-text N x N matrices.
```

src/dftbp/dftb/bondpopulations.F90
```diff
@@ -76,4 +75,44 @@ subroutine addPairWiseBondInfo(info, rhoPrim, sparseMat, iSquare, iNeighbour, nN
...
+      iOrb1 = iSquare(iAt1)
+      nOrb1 = iSquare(iAt1+1) - iOrb1
+      ! Upper triangle only; the result is symmetric in (iAt1, iAt2).
+      do iAt2 = iAt1, nAtom
+        iOrb2 = iSquare(iAt2)
+        nOrb2 = iSquare(iAt2+1) - iOrb2
+        acc = sum(prodPS(iOrb1:iOrb1+nOrb1-1, iOrb2:iOrb2+nOrb2-1)&
+            & * transpose(prodPS(iOrb2:iOrb2+nOrb2-1, iOrb1:iOrb1+nOrb1-1)))
+        mayerBO(iAt2, iAt1) = acc
+        mayerBO(iAt1, iAt2) = acc
```
bhourahine: Instead after the loop, could call adjointLowerTriangle from dftbp_math_matrixops. Probably slighly better cache performance.

src/dftbp/dftbplus/initprogram.F90
```diff
@@ -2018,6 +2027,9 @@ subroutine initProgramVariables(this, input, env)
         & this%tFixEf .or. this%tSpinSharedEf .or. this%isHybridXc .or. this%isMdftb .or.&
         & this%electronicSolver%iSolver == electronicSolverTypes%GF
     this%tAtomicEnergy = input%ctrl%tAtomicEnergy
+    this%tBondPopul = input%ctrl%tBondPopul
```
bhourahine: Are these OK for MPI enabled calculations? Looks like not, so should be trapped in the parser (probably WITH_MPI preprocessing)  and also here if withMpi is true.

src/dftbp/dftbplus/parser.F90
```diff
@@ -5424,6 +5424,10 @@ subroutine readAnalysis(node, ctrl, geom)
           call readCM5(child, ctrl%cm5Input, geom)
         end if
       end if
+      call getChildValue(node, "WriteBondPopulation", ctrl%tBondPopul, .false.)
```
bhourahine: Needs to be trapped for MPI enabled build.

test/app/dftb+/tests
```diff
@@ -413,6 +413,10 @@ timedep/propadiene_OscWindow           #? WITH_ARPACK and MPI_PROCS <= 8
 
 analysis/H2O_ESP                       #? MPI_PROCS <= 1
 analysis/H2O_mdESP                     #? MPI_PROCS <= 2
+analysis/bond_pop_N2                   #? MPI_PROCS <= 1
```
bhourahine:
```suggestion
analysis/bond_pop_N2                   #? not WITH_MPI
```
in this case.

bhourahine (commented):
Can you also add the keyword and output files to the manual and also add in the CHANGELOG as a new features?
The code should eventually be refactored to use the sparse real-space structures, which would allow periodic and MPI calculations transparently.

vnquanvuong:
I've updated it as you suggested.

bhourahine:
Squashed to one commit, and swapped SYMM for GEMM in the bond population.

One potential issue, which for the current cluster-only implementation is harmless, is the conversion of the density matrix back from sparse to dense, as is used for periodic cases, this is only an exact transformation for an absolutely converged k-point integration.

vnquanvuong:
Thanks for! I'll keep that in mind whenever periodic Mayer gets revisited

bhourahine:
@vnquanvuong I've now blocked use of the bond orders with MPI, as there are available solvers that only produce the density matrix masked by the overlap, so would give incorrect results. Also writing of results.tag now enabled and the internal logical variable names changed to match current convention.

PR #1856, opened 2026-06-01
Fix bugs in helpsetupgeom

pecchia:
I realized today creating a transport geometry that I had a pending fix in setupgeom that might be relevant for the tutorial and in general. 
- A bug was found in the processing of the second contact PL that was occasionally missing the correct atom and the minimal distance of the second neighbour PLs.
- A new subroutine reordering the PLs was introduced. 

Certainly the code is now working on a relatively complex Au/MoS2/Au system with random defects whereas the current version is failing with strange errors.


bhourahine (approved):
Other than missing comments on routines and some commented out lines that should just be removed (can clean all of this up later), looks OK.

PR #1858, opened 2026-06-09
Speed up RPA charge calls a bit

bhourahine:
Remove if branches inside loops and some comment clean-up.

Gives about a 10% speed-up over the timedep test set.


thomas-niehaus:
Very useful!

thomas-niehaus (approved):

PR #1864, opened 2026-06-26
Each atom pdos

bhourahine:
Atom specific PDOS projection, part of #1857 merge after #1863


pecchia:
I think this is ok. I do have a script to slice a structure along the transport direction and spit out all the Regions{}, essentially atomic planes.  Then, with various post-processing, it is possible to obtain the LDOS as contour plot.

PR #1865, opened 2026-06-26
Clean-up species external potentials

bhourahine:
Allow for multiple blocks with different species potentials

Also add missing documentation for speciesPotential feature

We should probably also settle on a more general syntax that includes atom ranges and species in the same notation for a future PR, before the release to simplify backward compatibility.


doc/dftb+/manual/dftbp.tex
```diff
@@ -1695,6 +1695,9 @@ \subsection{Density Functional based Tight Binding}
 \item[\is{AtomSitePotential}] Specifies an external potential at an atom. See
   p.~\pref{sec:dftbp.AtomSitePotential}.
 
+\item[\is{SpeciesPotential}] Specifies an external potential at an atom. See
```
fbonafe: Specifies an external potential for all atoms of a given species

fbonafe (commented):
Thanks for these improvements! Only a minor typo needs to be changed

PR #1869, opened 2026-06-27
Fix for uninitialised values

bhourahine:
As these can be used directly in a print statement.

Also cleaned up some source formatting and use of depreciated logical operators

This fails running with compiler undefined variable checking, as the gate length is otherwise only set on specific code paths.


bhourahine:
@fbonafe probably also the gate information should only be printed when they are set up (which is what tripped the NaN behaviour here).

fbonafe (approved):
Everything looks OK.

PR #1872, opened 2026-06-30
Use symmetric rank-k update for distributed density matrix

RizzoHou:
## Summary

`makeDensityMtxRealBlacs` builds the (non energy-weighted) distributed density matrix with a full `PGEMM`. For non-negative occupations, `rho = V diag(f) V^T = W W^T` with `W = V sqrt(f)`, so a symmetric rank-k update (`PSYRK`) forms only one triangle and roughly halves the floating-point work. The lower triangle is then mirrored into the upper one and the doubled diagonal halved. The serial (`herk`) and GPU (`syrk`) density-matrix builds already use a rank-k update; this brings the ScaLAPACK path in line.

## Correctness / fallbacks

The rank-k form assumes non-negative weights, so the existing `PGEMM` is kept where that does not hold:

- the energy-weighted matrix — occupied eigenvalues may be negative;
- negative occupations, e.g. Methfessel-Paxton filling, where `sqrt(f)` is not real.

Both fallback paths are byte-for-byte unchanged.

## Notes

- First of several upstreaming contributions discussed by email with Ben Hourahine; this is the self-contained, internal-only one (no input or interface change).
- Re-exports `pblasfx_psyrk` from the `scalapackfx` wrapper module.
- `AUTHORS.rst` updated.

### DIFF
--- a/src/dftbp/dftb/densitymatrix.F90
+++ b/src/dftbp/dftb/densitymatrix.F90
@@ -20,8 +20,8 @@ module dftbp_dftb_densitymatrix
   use dftbp_math_blasroutines, only : herk
   use dftbp_type_commontypes, only : TParallelKS
 #:if WITH_SCALAPACK
-  use dftbp_extlibs_scalapackfx, only : blacsgrid, blocklist, M_, pblasfx_pgemm, pblasfx_psyrk,&
-      & pblasfx_ptran, pblasfx_ptranc, scalafx_islocal, size
+  use dftbp_extlibs_scalapackfx, only : blacsgrid, blocklist, CSRC_, NB_, pblasfx_pgemm,&
+      & pblasfx_psyrk, pblasfx_ptran, pblasfx_ptranc, scalafx_indxl2g, scalafx_islocal, size
 #:endif
 #:if WITH_MAGMA
   use iso_fortran_env, only : int64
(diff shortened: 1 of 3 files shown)


src/dftbp/dftb/densitymatrix.F90
```diff
@@ -842,24 +844,49 @@ subroutine makeDensityMtxRealBlacs(myBlacs, desc, filling, eigenVecs, densityMtx
...
+      ! so the rank-k update below is not applicable. Use a matrix product.
       do ii = 1, size(blocks)
         call blocks%getblock(ii, iGlob, iLoc, blockSize)
         do jj = 0, blockSize - 1
           work(:, iLoc + jj) = eigenVecs(:, iLoc + jj) * eigenVals(iGlob + jj) * filling(iGlob + jj)
         end do
       end do
-    else
+      call pblasfx_pgemm(eigenVecs, desc, work, desc, densityMtx, desc, transb="T")
+    else if (minval(filling) < 0.0_dp) then
```
bhourahine:
```suggestion
    else if (any(filling < 0.0_dp)) then
```
This will probably complete faster than checking all values

src/dftbp/dftb/densitymatrix.F90
```diff
@@ -842,24 +844,49 @@ subroutine makeDensityMtxRealBlacs(myBlacs, desc, filling, eigenVecs, densityMtx
     ! Scale a copy of the eigenvectors
     call blocks%init(myBlacs, desc, "c")
     if (present(eigenVals)) then
+      ! Energy-weighted matrix: eigenvalues of occupied states may be negative,
```
bhourahine: A further optimisation would be to test if all of the filling * eigenvals products are <= 0 (which happens quite often), then use a rank-k update with alpha=-1.

bhourahine (commented):
One suggestion and one proposed change.

src/dftbp/dftb/densitymatrix.F90
```diff
@@ -842,27 +843,93 @@ subroutine makeDensityMtxRealBlacs(myBlacs, desc, filling, eigenVecs, densityMtx
...
+    real(dp), intent(inout) :: work(:,:)
+
+    integer :: ii, iLocRow, iLocCol
+    logical :: isLocal
+
+    ! Mirror the lower triangle into the upper triangle by adding the transpose;
+    ! this doubles the diagonal, which is halved afterwards.
+    work(:,:) = matrix
+    call pblasfx_ptran(work, desc, matrix, desc, alpha=1.0_dp, beta=1.0_dp)
+    do ii = 1, desc(M_)
```
bhourahine: I think this can be slightly sped-up. Currently it loops over the full order of the matrix on each rank before testing for local elements. So all ranks do the same operations and effectively behave like a serial operation in the scaling. Instead, something like the loop below will only check the range of the local part of the matrix for diagonal elements:

`do ii = 1, size(matrix, dim=2)`
`      iGlob = scalafx_indxl2g(ii, desc(NB_), myBlacs%mycol, desc(CSRC_), myBlacs%ncol)`
`      call scalafx_islocal(myBlacs, desc, iGlob, iGlob, tLocal, iLocRow, iLocCol)`
`      if (tLocal) then`

bhourahine (commented):
Ready to merge, but one final suggestion for optimisation.

bhourahine (approved):
Ready to merge. For 1000 sp atoms, I get an ~8% total code speed up on 4 ranks. We should also change the complex matrix cases as well in a future pull request.

PR #1874, opened 2026-07-01
Use Hermitian rank-k update for distributed complex density matrix

RizzoHou:
Follow-up to #1872 (symmetric rank-k update for the *real* distributed density matrix), extending the same optimization to the complex path as suggested in that review.

`makeDensityMtxCplxBlacs` built the density matrix with a full PGEMM. For non-negative occupations ρ = V diag(f) Vᴴ equals W Wᴴ with W = V √f, a Hermitian rank-k update (PHERK) that forms only one triangle and roughly halves the floating-point work. This mirrors the PSYRK change to the real path in #1872.

The energy-weighted matrix W = V diag(f e) Vᴴ uses the same rank-k update with a prefactor of −1 when every occupied product f·e is non-positive (the common case); mixed signs fall back to PGEMM. Occupations that may themselves be negative (e.g. Methfessel–Paxton filling) also keep the PGEMM, which makes no sign assumption.

The two PHERK branches fill only the lower triangle and complete it with the existing `adjointLowerTriangle_BLACS` utility (`dftbp_math_matrixops`), which copies the lower triangle into the upper one and forces a real diagonal. The PGEMM branches already build a full matrix and need no completion — matching the placement in `makeDensityMtxRealBlacs`. `pblasfx_pherk` is re-exported from the scalapackfx wrapper module (as `pblasfx_psyrk` was in #1872).

*(Note: the real path in #1872 hand-rolls the equivalent lower→upper completion; it could be simplified to use `adjointLowerTriangle_BLACS` too in a follow-up, so both paths share one implementation.)*

### Validation

Built with Intel oneAPI (MPI + ELSI/ELPA) and run on 2 MPI ranks against the autotest references (generated with the original PGEMM code). Every complex-density-matrix path reproduces the references to ≤ ~2e-11:

| test | complex path exercised | worst deviation |
|------|------------------------|-----------------|
| `md/SiC64-xlbomdfast_kpts` | k-points; DM + energy-weighted DM | 9.6e-12 |
| `md/SiC64-xlbomdfast-T0_kpts` | k-points, T = 0 | 2.0e-11 |
| `geoopt/GaAs_8_latconst` | k-points; forces + stress | 1.5e-11 |
| `analysis/Ga4As4_ESP` | k-points | 3.4e-12 |
| `spinorbit/As4S4` | spin-orbit; forces | OK (within test tolerance) |
| `spinorbit/Fe2`, `spinorbit/EuN` | spin-orbit | 2e-12 |
| `spin/Fe4_noncolinear` | noncollinear | 1.5e-11 |

### DIFF
--- a/src/dftbp/dftb/densitymatrix.F90
+++ b/src/dftbp/dftb/densitymatrix.F90
@@ -20,8 +20,8 @@ module dftbp_dftb_densitymatrix
   use dftbp_math_blasroutines, only : herk
   use dftbp_type_commontypes, only : TParallelKS
 #:if WITH_SCALAPACK
-  use dftbp_extlibs_scalapackfx, only : blacsgrid, blocklist, pblasfx_pgemm, pblasfx_psyr,&
-      & pblasfx_psyrk, pblasfx_ptranc, size
+  use dftbp_extlibs_scalapackfx, only : blacsgrid, blocklist, pblasfx_pgemm, pblasfx_pher,&
+      & pblasfx_pherk, pblasfx_psyr, pblasfx_psyrk, pblasfx_ptranc, size
   use dftbp_math_matrixops, only : adjointLowerTriangle_BLACS
 #:endif
 #:if WITH_MAGMA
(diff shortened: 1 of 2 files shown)


RizzoHou:
Pushed the same tiny-occupation-weight handling for the Hermitian rank-k paths here (mirror of #1877, which carries the full diagnosis of the NaN failures reported against the real path from #1872): weights with magnitude in [ε, √ε) go through exact rank-1 PZHER updates (real α, no square root) instead of the square-rooted PHERK; weights below ε are dropped; more than 32 rank-1 levels falls back to the previous matrix product; and sub-ε products no longer veto the rank-k paths. The `.not. (w < √ε)` test form keeps non-finite occupations propagating through `sqrt`.

Validated against a stock PGEMM build of the same source (Intel oneAPI 2025.3 + MKL): a hot-Fermi variant of `spinorbit/As4S4` at 2400 K puts 24 degenerate spinor levels into the rank-1 window and agrees to 1.9e-14; at 2000 K, 62 levels trigger the matrix-product fallback and agree to 2.4e-14; the original 8-test battery is unchanged (bit-identical where no tiny weights occur).

RizzoHou:
Mirrored the #1877 review changes here in 58d0927b: `dropTol = 16.0_dp * epsilon(1.0_rdp)` applied to every negligibility gate, thresholds anchored to double precision (this path already used `adjointLowerTriangle_BLACS`).

While re-checking the branch I noticed the split into pherk-plus-fallback had silently dropped the unconditional Hermitian symmetrisation that the original single-PGEMM path applied at the end. b4457df6 restores it on the two matrix-product fallback branches, so consumers receive an exactly Hermitian matrix regardless of which branch built it (the pherk/rank-1 path already gets this from `adjointLowerTriangle_BLACS`).

Re-validated on Intel oneAPI 2025.3 + MKL ScaLAPACK: the complex battery (SiC64 k-point MD, GaAs_8 lattice optimisation, Ga4As4 ESP, As4S4/Fe2/EuN spin-orbit, Fe4 non-collinear) passes, and As4S4 heated into the rank-1 window (T = 2400 K) and past the cap into the fallback (T = 2000 K) matches the stock GEMM build to 2e-14.

RizzoHou:
Rebased onto current `main` and force-pushed, as GitHub had started reporting this branch as conflicting.

The conflicts came from the merge of #1877: it rewrote the same two `use` blocks this PR touches — pruning `CSRC_`, `NB_`, `scalafx_indxl2g`, `scalafx_islocal` and `pblasfx_ptran` from `densitymatrix.F90` (unused once the local helper was replaced by `adjointLowerTriangle_BLACS`) and adding `pblasfx_psyr` and `rdp` — while `7e1ceaff` separately added `CTXT_` and `scalafx_numroc` to the wrapper module's re-export list. This branch still carried the pre-#1877 form of `makeDensityMtxRealBlacs` and its import list. The conflicts were confined to the import lists; there was no overlap in the routine bodies.

Resolution takes `main`'s import lists as they stand and adds `pblasfx_pher` and `pblasfx_pherk`. After the rebase, `makeDensityMtxRealBlacs` is byte-identical to `main`, so the change is confined to `makeDensityMtxCplxBlacs` plus the two `use` statements, and the resulting `densitymatrix.F90` is byte-identical to the tree the earlier validation runs were done against — no source change beyond the integration itself.

The four commits are squashed into one, with the message describing the final state (the first of them predated the restored Hermitian symmetrisation on the matrix-product branches and claimed no triangle completion was needed there).

bhourahine:
One regression test is failing (geoopt/diamond_isotropic for Intel one API 2025.3 1 MPI rank and 2 OMP threads). The test has default convergence tolerances and numerically only just fails. I will change the test to have a tighter geometry convergence and merge if that passes.

PR #1875, opened 2026-07-01
Unified and simplified external potentials

bhourahine:
Also removed one of the regression tests, as the other covers all keyword options.


test/app/dftb+/extpot/CH3OH_species/dftb_in.hsd
```diff
@@ -12,13 +12,17 @@ Methanol
...
     SCCTolerance = 1.0E-8
-    SpeciesPotential {
+    AtomSitePotential {
         Gross {
-            Species = C H
-            Vext [eV] = 1.5
+            Atoms = C 2 # all C atoms and first H atom
+            Vext [eV] = 1.5 # all with same potential
+        }
+        Gross {
```
fbonafe, line 20: Would be good to test for the net potential as well.

fbonafe (approved):

PR #1877, opened 2026-07-04
Handle tiny occupation weights in distributed rank-k density matrix

RizzoHou:
Fixes the regression-test failures @bhourahine reported against the rank-k density-matrix build from #1872: with some compiler/library combinations, `scc/C60_Fermi` ends with NaN outputs in the final SCC steps, set off by a Fermi occupation that is greater than zero but smaller than √ε, with a `dstegr2()` call inside the following diagonalisation failing with an impossible argument number.

### Diagnosis

Reproduced in part with Intel oneAPI 2025.3 + MKL ScaLAPACK. `scc/C60_Fermi` has six levels with occupations in [ε, √ε) (a triply degenerate level with f ≈ 5.5e-10 and one with 4.2e-14) plus ~114 Fermi-tail occupations below ε that never underflow to zero at the test's fixed Fermi level. With the rank-k build, MKL's PDSYEVR deterministically reports

```
DSTEGR2 parameter number -202 had an illegal value
```

on 1 MPI rank, in the diagonalisation right after these fillings settle (20/20 runs; 0/20 with the pre-#1872 PGEMM build of the same source). MKL recovers and still converges to the reference results (deviations ≤ 1e-12), while the NAG stack turns the same MRRR fragility into NaN. The DSTEGR2 defect itself appears to be library-side — under MKL it keeps triggering on the bit-level DM differences that remain after this fix — but square-rooting near-underflow occupations is the part the failing toolchains cannot tolerate (zeroing them was already shown to make the NAG failure disappear).

### Change

- Occupation weights with magnitude in [ε, √ε) are excluded from the square-rooted rank-k update and applied exactly through rank-1 updates (PDSYR, real α = f or f·e — no square root involved).
- Weights below ε are dropped: they contribute below double-precision resolution to any matrix element.
- Should more than a handful (32) of levels fall into the rank-1 window — each rank-1 update is a memory-bound level-2 pass over the distributed matrix, and a strongly smeared metallic system could put hundreds of levels there — the previous full matrix product is used instead.
- Products below ε in magnitude no longer veto the rank-k paths (previously a single virtual level with f·e = +1e-16 was enough to force the energy-weighted build onto PGEMM permanently).
- The weight tests use the `.not. (w < √ε)` form, so non-finite occupations still propagate through `sqrt` instead of being silently zeroed.
- `pblasfx_psyr` is re-exported from the scalapackfx wrapper module (as `pblasfx_psyrk` was in #1872).

### Validation (Intel oneAPI 2025.3 + MKL ScaLAPACK)

| check | result |
|---|---|
| `scc/C60_Fermi` vs reference, np = 1…16 | pass, worst deviation 1.2e-12 (forces now on the rank-k + rank-1 energy-weighted path) |
| same source built with the original PGEMM code, cross-compared at np = 1, 2, 4 | all fields agree to ≤ 1e-11 |
| vs the plain #1872 rank-k build at np = 2, 8 | ≤ 1.1e-12 |
| `scc/C60`, `scc/GaAs_2` | pass |

We have no NAG compiler available, so the confirmation for the originally failing combinations has to come from that side. The same treatment for the complex (PHERK) paths has been pushed to #1874.

### DIFF
--- a/src/dftbp/dftb/densitymatrix.F90
+++ b/src/dftbp/dftb/densitymatrix.F90
@@ -13,15 +13,16 @@
 !! Caveat: The routines create the transposed and complex conjugated of the density matrices
 !! (cc* instead of the conventional c*c).
 module dftbp_dftb_densitymatrix
-  use dftbp_common_accuracy, only : dp, lc
+  use dftbp_common_accuracy, only : dp, lc, rdp
   use dftbp_common_constants, only : imag, pi
   use dftbp_common_status, only : TStatus
   use dftbp_elecsolvers_dmsolvertypes, only : densityMatrixTypes
   use dftbp_math_blasroutines, only : herk
   use dftbp_type_commontypes, only : TParallelKS
 #:if WITH_SCALAPACK
-  use dftbp_extlibs_scalapackfx, only : blacsgrid, blocklist, CSRC_, NB_, pblasfx_pgemm,&
-      & pblasfx_psyrk, pblasfx_ptran, pblasfx_ptranc, scalafx_indxl2g, scalafx_islocal, size
+  use dftbp_extlibs_scalapackfx, only : blacsgrid, blocklist, pblasfx_pgemm, pblasfx_psyr,&
+      & pblasfx_psyrk, pblasfx_ptranc, size
+  use dftbp_math_matrixops, only : adjointLowerTriangle_BLACS
 #:endif
 #:if WITH_MAGMA
   use iso_fortran_env, only : int64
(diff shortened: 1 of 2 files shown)


src/dftbp/dftb/densitymatrix.F90
```diff
@@ -833,33 +834,63 @@ subroutine makeDensityMtxRealBlacs(myBlacs, desc, filling, eigenVecs, densityMtx
...
+
+    ! Square-rooting occupation weights close to underflow in the rank-k updates below has
+    ! been observed to destabilise subsequent ScaLAPACK eigensolver (MRRR) calls for some
+    ! compiler/library combinations. Levels with weight magnitude in [eps, sqrt(eps)) are
+    ! therefore applied through exact rank-1 updates instead, while weights below eps are
+    ! dropped, as they contribute below double precision resolution to any matrix element.
+    ! Should more than maxRank1 levels fall into the rank-1 window (each update is a
+    ! memory-bound level-2 pass over the distributed matrix), the plain matrix product is
+    ! used instead.
+    real(dp), parameter :: sqrtEps = sqrt(epsilon(1.0_dp))
```
bhourahine:
```suggestion
    real(dp), parameter :: sqrtEps = sqrt(epsilon(1.0_rdp))
```
Which is specifically the real double precision, instead of the working kind. This would also need that variable to be used from dftbp_common_accuracy

src/dftbp/dftb/densitymatrix.F90
```diff
@@ -869,29 +900,45 @@ subroutine makeDensityMtxRealBlacs(myBlacs, desc, filling, eigenVecs, densityMtx
...
+            if (.not. (filling(iGlob + jj) < sqrtEps)) then
+              work(:, iLoc + jj) = eigenVecs(:, iLoc + jj) * sqrt(filling(iGlob + jj))
+            else
+              work(:, iLoc + jj) = 0.0_dp
+            end if
+          end do
+        end do
+        call pblasfx_psyrk(work, desc, densityMtx, desc, uplo="L", trans="N")
+        do iLev = 1, size(filling)
+          if (filling(iLev) >= epsilon(1.0_dp) .and. filling(iLev) < sqrtEps) then
```
bhourahine:
```suggestion
          if (filling(iLev) >= 16.0_dp * epsilon(1.0_rdp) .and. filling(iLev) < sqrtEps) then
```
Truncating at a few times eps seems stable for all the compilers I've tested (one fails for 1*eps).

src/dftbp/dftb/densitymatrix.F90
```diff
@@ -869,29 +900,45 @@ subroutine makeDensityMtxRealBlacs(myBlacs, desc, filling, eigenVecs, densityMtx
...
+            call pblasfx_psyr(eigenVecs, desc, densityMtx, desc, uplo="L",&
+                & alpha=filling(iLev), jx=iLev)
+          end if
+        end do
+        call addLowerTriangleTranspose(myBlacs, desc, densityMtx, work)
```
bhourahine: May as well switch to the adjointLowerTriangle_BLACS routine and reduce code duplication (that routine is probably slightly slower, but ignores any non-zero noise in the upper triangle).

bhourahine (commented):
Your PR deals with the symptoms of the C60_Fermi fail and passes on the compilers I've tested (but note the comment about truncation at a few times eps, which was needed for one of them to pass), but unfortunately the timedep/C60_OscWindow regression case is still a problem.
A partial work-around for that test might be to perturb the resulting density matrix by adding values of order epsilon to the elements. But obviously this does not fix the underlying library problem and seems to potentially be problematic for other tests. I _think_ the issue is the MRRR eigenvectors, as these are then used in the timedep routines.

RizzoHou:
All three suggestions applied in 74961db7:

- Thresholds are now anchored to `epsilon(1.0_rdp)`, with `rdp` taken from `dftbp_common_accuracy`.
- The truncation floor is a named constant, `dropTol = 16.0_dp * epsilon(1.0_rdp)`. I applied it to every gate that classifies a weight as negligible — the two `nTiny` counts, both rank-1 loop bounds, the energy-weighted `all()` exemption and the negative-occupation test — not only the suggested line, so the counted window and the applied window stay identical. Happy to narrow that if you would rather keep any individual gate at `1 * eps`.
- The local `addLowerTriangleTranspose` helper is gone; both call sites now use `adjointLowerTriangle_BLACS`, matching the complex path in #1874 (the threshold changes are mirrored there).

Revalidated on Intel oneAPI 2025.3 + MKL ScaLAPACK: scc/C60_Fermi passes at np = 1–8, 12, 16 (worst deviation 1.2e-12 vs the reference, 1.0e-12 vs stock GEMM at np = 1, 2, 4), and As4S4 heated to put levels into the rank-1 window (T = 2400 K) and past the cap into the matrix-product fallback (T = 2000 K) matches stock to 2e-14.

One data point in favour of the wider floor: with the floor at `1 * eps`, MKL's PDSYEVR deterministically printed `DSTEGR2 parameter number -202 had an illegal value` at np = 1 on C60_Fermi (results still correct, 20/20 repeats); with this version the message is gone (0/10 repeats). Given how sensitive the underlying failure is to any bit-level change I would not claim causality, but the residual MRRR complaint on our stack has disappeared too. It may also be worth noting for the library reports that at `1 * eps` the instability persisted even though the tiny levels were applied through exact rank-1 updates with no square root involved — the trigger seems to be the mere presence of O(eps) structure in the matrix rather than anything specific to the rank-k arithmetic.

On timedep/C60_OscWindow: agreed that perturbing the density matrix would only mask the library problem — better left to the MRRR reports.

bhourahine:
@RizzoHou I've updated your branch to include the recent commits (including the sort changes) and will now run some more regression tests. If all work OK, this should be merged later today.

PR #1879, opened 2026-07-06
Use pxgemm from ELPA for density matrix creation

terminationshock:


bhourahine (commented):
#1877 had to add a work around for numerically small weights for rank-k which may be relevant here (this also is to mitigate a bug/bugs in ScaLAPACK). Also, there is a relevant ELPA bug fix in #1881

terminationshock:
This merge request will remain in 'draft' for a bit of time, because the pxgemm interface in ELPA will change with the next release. It only makes sense to include it in DFTB+ after the update.

PR #1881, opened 2026-07-09
Fix for ELPA input being allocatable

bhourahine:
Otherwise fails for non-ELPA solvers in ELSI


terminationshock:
Thanks @bhourahine for this fix!

PR #1884, opened 2026-07-18
Remove binary GF and output test files

bhourahine:


pecchia:
I do not remember how the soure- drain-SurfaceGF.mgf binaries have slip into the repository. It seems to be a dephasing stuff. I do not think these files are created by the current version of libnegf. I'm also fine removing all _tunneling.dat files since now this is written and checked in autotest.tag.

On the other hand, I am a bit in doubt about the removal of _output files. Even though these are not checked, I think I used them occasionally when trying to neil down possible bugs or reasons of different outputs. In general these are small text files. But if you think they are really polluting the repository you can delete them.

pecchia (approved):
As said in the general comment, I am fine removing the binaries and _tunneling.dat but perhaps I would leave _output in case they might help understanding some misbehaviour. But I leave the decision to you. It is always possible to reproduce an output with an older working version, so ok. Let's delete the outputs.

PR #1885, opened 2026-07-19
Add SinglePrecisionCycles option for the ELPA solver (via ELSI)

ZMYsamuel:
### Motivation

ELSI has long provided a mixed-precision mode for its ELPA interface
(`elsi_set_elpa_n_single`, part of the ELPA-AEO work, Kus et al., Parallel
Computing 85 (2019) 167, arXiv:1806.01036): the first *n* solves after
(re)initialisation are carried out by ELPA in single precision, all later
solves in double precision. Since the early SCC iterations operate on charges
that are still far from self-consistency, solving them to ~7 significant
digits is enough — the SCC loop itself irons out the difference, and the
final iterations (and hence the converged result) are full double precision.
In the ELPA-AEO paper the effect of the mixed-precision startup on converged
total energies was below ~1e-6 eV.

DFTB+ currently never calls `elsi_set_elpa_n_single`, so this lever is
inaccessible to users. This PR exposes it as an option of the `ELPA{}` solver
block.

Why it is worth having: on large dense problems the eigensolve dominates the
SCC iteration cost, and a single-precision ELPA solve runs substantially
faster than the double-precision one (roughly 1.5–1.9x per solve in the
ELPA-AEO paper). We have not benchmarked `elsi_set_elpa_n_single` itself in
DFTB+ (that is exactly what this PR enables); the closest measurement we do
have is an analogous FP32-early/FP64-tail SCC schedule implemented on a GPU
eigensolver backend for the same system class (8000-atom 3C-SiC, n = 32,000
basis functions, DFTB+ 25.1): 382 s -> 269 s (~1.4x) for the precision change
alone, converged total energy within 1.4e-9 H of the all-double reference.
(The 382 s is itself down from a 420 s FP64 baseline, but that first step came
from moving the density-matrix build onto the GPU and is a separate lever, so
it is not counted here.) For a run with 10+ SCC iterations, solving the early
ones in single precision saves a significant fraction of the total wall clock
at no cost to the converged result.

(Caveat on that measurement: the 8000-atom cell is an antisite-defected
geometry that came with an input deck I cannot redistribute, so the figure is
not independently reproducible as it stands. I will rebuild the case on a
generated geometry with a CPU-eigensolver reference and re-measure.)

### What changed

- `ELPA{ SinglePrecisionCycles = <int> }` HSD option (default `0` =
  unchanged current behaviour). Parsed in `parser.F90` in the same way as the
  other ELPA-block options, stored in `TElpaInp`, copied into `TElsiSolver`,
  and applied via a new `elsi_set_elpa_n_single` wrapper in
  `dftbp_extlibs_elsiiface` (with the usual stub for non-ELSI builds).
- The ELSI call is made only when the value is positive, so default inputs
  produce a bit-identical call sequence to current `main`.
- Validation:
  - negative values are rejected (parser `detailedError` + a defensive check
    in `TElsiSolver_init` for API-driven inputs);
  - builds without ELSI reject a non-zero value at parse time;
  - if the standalone ELPA path is dispatched (built `WITH_ELPA` and
    `PreferElsi = No`), `TElpa_init` stops with an error pointing the user to
    `PreferElsi = Yes` — the option is deliberately not implemented for the
    standalone path in this PR (the standalone `TElpa` type is double
    precision only at the moment; can be done as a follow-up if there is
    interest).
- Manual: new entry in the ELPA solver block section (including the pointer
  that a few extra SCC iterations may be needed depending on `SCCTolerance`),
  with the ELPA-AEO reference added to `manual.bib`.
- `CHANGELOG.rst`: entry under Unreleased/Added.

Files touched: `src/dftbp/dftbplus/parser.F90`,
`src/dftbp/elecsolvers/elpa.F90`, `src/dftbp/elecsolvers/elsisolver.F90`,
`src/dftbp/extlibs/elsiiface.F90`, `doc/dftb+/manual/dftbp.tex`,
`doc/dftb+/manual/manual.bib`, `CHANGELOG.rst`.

### Accuracy discussion

- Only the first `SinglePrecisionCycles` diagonalisations are affected. As
  long as the SCC loop still converges (it does in all our tests of the
  GPU-eigensolver analogue, sometimes with 1–2 extra iterations), the
  converged energies/forces are produced by double-precision solves and match
  the reference to the usual SCC tolerance; ELPA-AEO reports total-energy
  deviations below ~1e-6 eV.
- Results within the *first* iterations (e.g. non-SCC runs printing the first
  diagonalisation) would be single precision — hence the conservative default
  of 0, and the manual wording that later cycles/final results stay double
  precision.
- One thing reviewers may want to weigh in on: ELSI applies `elpa_n_single`
  to the first *n* solves after `elsi_init`/`elsi_reinit`. Since DFTB+ calls
  `elsi_reinit` on geometry changes, the single-precision startup applies to
  the first iterations of each geometry step of an MD/relaxation run. That
  falls out of ELSI resetting its call counter on reinit rather than from a
  deliberate choice on our part, so whether it is the desired semantics is a
  decision for the maintainers rather than something we should assert.
- Requirement on the ELSI side: the linked ELPA must provide single-precision
  routines (ELSI's internal ELPA does; an external ELPA needs
  `--enable-single-precision`). If it does not, ELSI fails the solve at run
  time; we did not add a configure-time check — input welcome on whether one
  is wanted.

### Test plan

- Serial (no MPI/ELSI/ELPA) Debug build with GNU 13.2 compiles cleanly with
  this branch; fypp expansion checked for the WITH_ELSI / WITH_ELPA-only /
  serial variants. [Done locally.]
- Not yet validated in an ELSI build. The serial build above does not compile
  the ELSI code path, so this change has not yet been exercised. Before taking
  this out of draft I intend to do a `WITH_ELSI=1` MPI build and run `ctest`
  over the existing `solvers/*_ELPA*` cases to confirm the defaults are
  regression-free, plus a manual run of `Si384_ELPA2` with the new option set.
  I will report those results here rather than ask for review before they
  exist.
- Offer to maintainers: add an autotest case (clone of an existing ELPA test
  with `SinglePrecisionCycles = 2`); its references should agree with the
  double-precision ones within autotest tolerances since the final iterations
  are double precision. Depends on the CI ELSI build having single-precision
  ELPA support.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -38,6 +38,10 @@ Added
   with optional MPI redistribution to a lower number of ranks for all
   ELPA calls
 
+- SinglePrecisionCycles option for the ELPA solver (via the ELSI
+  library), solving the eigenvalue problem in the first few SCC cycles in
+  single precision
+
 
 Changed
 -------
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -93,3 +93,5 @@ contributed to DFTB+ :
 * Bo Thomsen (Japan Atomic Energy Agency, Japan)
 
 * Victor Yu (Duke University, USA)
+
+* Samuel Zheng (National Tsing Hua University, Taiwan)
--- a/doc/dftb+/manual/dftbp.tex
+++ b/doc/dftb+/manual/dftbp.tex
@@ -2522,6 +2522,7 @@ It accepts following options
   \kw{PreferElsi} & l& & No & \\
   \kw{RedistributeFactor} & i& & 1 & \\
   \kw{RedistributeRanks} & p& RedistributeFactor = 1 & \cb & \\
+  \kw{SinglePrecisionCycles} & i& & 0 & \\
   \kw{Sparse} & l& & No & \\
 \end{ptable}
 \begin{description}
@@ -2547,6 +2548,17 @@ It accepts following options
   available if ELPA is included via the ELSI library.
 \item[\is{RedistributeRanks}] Explicit list of MPI ranks to be used to call
   ELPA, for example, \is{\{0, 2\}}. The matrix will be redistributed accordingly.
+\item[\is{SinglePrecisionCycles}] Number of initial SCC cycles for which the
+  eigenvalue problem is solved in single precision~\cite{kus2019optimizations}
+  (must be greater or equal to \is{0}, default: \is{0}, i.e., all cycles are
+  solved in double precision). All later SCC cycles, and therefore the final
+  results, are obtained in double precision, while the early cycles, where the
+  charges are far from self-consistency anyway, become considerably
+  cheaper. Depending on the requested \is{SCCTolerance}, a few additional SCC
+  cycles may be needed to reach self-consistency. This feature is only
+  available if ELPA is included via the ELSI library, and requires the
+  ELPA library within ELSI to have been built with single precision support
+  enabled.
 \item[\is{Sparse}] Whether the sparse matrix interface of ELPA should be used or
   the dense one (default: dense). The sparse interface does not reduce memory
   usage and is mainly for testing purposes.
--- a/doc/dftb+/manual/manual.bib
+++ b/doc/dftb+/manual/manual.bib
@@ -550,6 +550,20 @@ year = {2020}
   doi = {https://doi.org/10.1016/j.cpc.2020.107459}
 }
 
+@article{kus2019optimizations,
+  title = {Optimizations of the eigensolvers in the {ELPA} library},
+  author = {K\r{u}s, Pavel and Marek, Andreas and K{\"o}cher, Simone
+                  S. and Kowalski, Hermann-Herbert and Carbogno,
+                  Christian and Scheurer, Christoph and Reuter,
+                  Karsten and Scheffler, Matthias and Lederer,
+                  Hermann},
+  journal = {Parallel Computing},
+  volume = {85},
+  pages = {167-177},
+  year = {2019},
+  doi = {https://doi.org/10.1016/j.parco.2019.04.003}
+}
+
 @article{vuong2023accelerating,
   title={Accelerating the density-functional tight-binding method
                   using graphical processing units},
--- a/src/dftbp/dftbplus/parser.F90
+++ b/src/dftbp/dftbplus/parser.F90
@@ -2751,6 +2751,17 @@ contains
               & the GPU acceleration for the ELPA solver")
         end if
       #:endif
+      call getChildValue(value1, "SinglePrecisionCycles", ctrl%solver%elpa%nSinglePrecCycles, 0,&
+          & child=child)
+      if (ctrl%solver%elpa%nSinglePrecCycles < 0) then
+        call detailedError(child, "SinglePrecisionCycles must not be negative")
+      end if
+      #:if not WITH_ELSI
+        if (ctrl%solver%elpa%nSinglePrecCycles /= 0) then
+          call detailedError(child, "Single precision cycles are only possible if ELPA is&
+              & included via the ELSI library")
+        end if
+      #:endif
       call getChildValue(value1, "RedistributeFactor", ctrl%solver%elpa%redistributeFactor, 1,&
           & child=child)
       #:if not WITH_ELPA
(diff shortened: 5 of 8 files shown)


ZMYsamuel:
Two corrections and two scope fixes to the PR description, which I will correct in the body as well.

The motivation quotes the analogous GPU-eigensolver measurement as 420 s → 269 s (~1.56x). That
folds two independent levers into one number: 420 s is the FP64 baseline from before the
density-matrix build was moved onto the GPU, and that move alone accounts for 420 → 382 s. The
isolated precision lever — the one this option is an analogue of — is 382 → 269 s, about 1.4x.

Second, the description says the 8000-atom 3C-SiC benchmark geometry "is generated by a small
script we can share, so the case is reproducible". It is not. The cell measured is
antisite-defected (3992 C / 4008 Si); the generator emits the stoichiometric 4000/4000 crystal in
a different atom order, converging in a different number of SCC iterations and to a different
energy. Those numbers are from a competition input deck I cannot redistribute. I will replace them
with measurements on a redistributable case carrying reference values.

Third, a softening rather than an error. The description says of the ELSI behaviour on
`elsi_reinit` that the resulting single-precision restart at each geometry step is "the desired
semantics". That is not mine to assert: it falls out of ELSI resetting its call counter on reinit
rather than from a deliberate choice, and whether it is desirable is a decision for the
maintainers. I am rewording it to describe the behaviour and leave the judgement open.

Fourth, in the same spirit. The accuracy discussion says the SCC loop still converges "in all our
tests". Those tests are of the GPU-eigensolver analogue, not of `SinglePrecisionCycles` itself —
as the body says elsewhere, `elsi_set_elpa_n_single` has not been benchmarked in DFTB+. I am
scoping that sentence to say so, since as written it reads as evidence for an option that has not
been measured here.

CHANGELOG.rst
```diff
@@ -38,6 +38,10 @@ Added
   with optional MPI redistribution to a lower number of ranks for all
   ELPA calls
 
+- SinglePrecisionCycles option for the ELPA solver (via the ELSI
+  library), solving the eigenvalue problem in the first SCC cycles in
```
bhourahine:
```suggestion
  library), solving the eigenvalue problem in the first few SCC cycles in
```

bhourahine (commented):
See a minor suggestion, but also, isElpaStandalone should be tested if the cycles are set, as direct ELPA linkage is also possible.

bhourahine:
@terminationshock would you mind having a quick look at this?

terminationshock:
LGTM. As far as I can see, there is a safeguard in `src/dftbp/elecsolvers/elpa.F90` that will throw an error in the direct integration case.

bhourahine:
@ZMYsamuel You should also update the AUTHORS.rst file

ZMYsamuel:
Thanks both — suggestion applied, and I have added myself to `AUTHORS.rst`.

On `isElpaStandalone`: @terminationshock is right. Worth adding that it also holds in the case the compile-time check misses — built `WITH_ELPA` *and* `WITH_ELSI` with `PreferElsi` at its `.false.` default, where `parser.F90:2759` is compiled out but `isElpaStandalone` is still `.true.` (`elecsolvers.F90:128` tests `.not. (withElsi .and. preferElsi)`), so `TElpa_init` runs and `elpa.F90:198` errors as intended.

Marking it ready for review. To be explicit about what is and is not tested: I have not run `ctest` against a `WITH_ELSI=1` build myself, so if you would rather have that before merging, say so and I will do it first.

bhourahine:
@ZMYsamuel we probably need at least one test case in the regression system.

I've taken the liberty of squashing and rebasing to current main.

After this PR is merged, I can see some possible follow-ups in this space (cases that optimize to a fixed point, such as geometry optimization probably can run early iterations fully in single precision with loose tolerances, but also for modes like molecular dynamics perhaps there are some things that can be done with predictor-corrector methods). Also, working out some heuristics to choose the number of elpa_n_single steps might be possible given either multiple steps or on the basis of convergence rates.

ZMYsamuel:
Thanks, Ben — understood. I’ll add at least one regression test and follow up once I’ve had a chance to work through it. Thanks as well for squashing and rebasing the PR onto the current `main`.

PR #1893, opened 2026-07-27
Add partial diagonalisation for the ELPA solver

RizzoHou:
This implements the generalisation of the partial diagonalisation you sketched
in your last two mails: start from a specified number of empty states, increase
it during the self-consistent cycle if it turns out to be too small, and do not
let the cycle finish while there is no completely empty state.

## Interface

```
Solver = ELPA {
  Mode = 2
  PreferElsi = Yes
  EmptyStates = 8
}
```

The default is a negative value, which calculates the whole spectrum exactly as
before, so nothing changes for existing inputs.

## How the count is chosen

`EmptyStates` is a starting value only. After each diagonalisation the highest
calculated state is compared with the chemical potential. If it does not lie far
enough above it, the number of empty states is doubled and `tConverged` is
forced false for that iteration, so early iterations tolerate too few empty
states but the converged result cannot be missing occupied ones.

The distance uses your energy-cutoff heuristic rather than an occupation
threshold. **I did not use the multipliers from your mail, and I would like your
view on this.** You suggested ~7 widths for Fermi and ~4 for Methfessel-Paxton
and Gaussian. Those two are not equivalent in the size of what they discard: the
Fermi-Dirac occupation at 7 kT is 1/(1+e^7) = 9.1e-4, while erfc(4)/2 = 7.7e-9.
Empirically 7 kT was too loose — `spin/Fe4_ELPA_2group` (Methfessel-Paxton,
500 K) stalled at an SCC error of ~1.5e-7 against a tolerance of 1e-8, because
the neglected tail leaves a systematic charge error the cycle cannot remove.

So I took the principle from your mail (the point where the filling becomes
numerically insignificant) and solved for it at a target occupation of 1e-10,
giving 24 widths for Fermi-Dirac and 5 for the Gaussian family. These are named
constants in one place and easy to change if you would rather have different
values. For gapped systems the larger Fermi window costs nothing: SiC at 1000 K
is still satisfied by a single empty state.

## Structure

- `elecsolvers/partialdiag.F90` (new) holds the state counting and the
  sufficiency test, with no solver or MPI dependency, so it is unit tested
  directly. The same machinery could later drive a range-limited ScaLAPACK
  solver.
- `TElsiSolver_setNState` rebuilds the ELSI handle, since the number of states
  is fixed at `elsi_init`. The electronic broadening is remembered and applied
  to the new handle, as it is set outside the solver setup and would otherwise
  fall back to the library default.
- `TElectronicSolver_setNState` additionally clears the cached Cholesky
  factorisation. This was a real bug during development: the freshly built
  handle factorises the overlap itself, but DFTB+ was handing it the stored
  factorisation from the previous handle, and ELPA failed with a Cholesky error.
- The charge mixer keeps its history across an increase. Restarting it was
  measured to cost iterations without changing the result.

## Restrictions

Rejected with an explanatory error for a non-SCC calculation, since the count
can only grow during the self-consistent cycle, and when combined with anything
that needs the states left out: excited state and perturbation calculations,
electron dynamics, Delta DFTB, REKS, custom occupations, non-collinear spin,
eigenvector output, and projecting the eigenstates onto regions. Also rejected
for `IndependentKFilling`, which gives a chemical potential per k-point while
the sufficiency test only receives their weighted average. Also rejected when
ELPA is called directly rather than through ELSI, since only the ELSI path is
wired up here; the direct path could follow with the same machinery.
`EmptyStates = 0` is rejected too, as the stop criterion itself needs one empty
state.

## Validation

Built with Intel oneAPI 2025.3 + MKL against ELSI 2.12.0 and ELPA 2024.05.001.

Full test suites: 404/407 serial and 314/316 MPI pass; the 3 (2) failures are
`solvation/gbsa_param*`, which fail on this machine for a missing parameter file
and are unrelated.

Every case below was run with the full spectrum and then with a partial
diagonalisation, and the autotest tags compared:

| case | EmptyStates | growth steps | worst deviation |
|---|---|---|---|
| `scc/C60_Fermi` (Fermi 1000 K, fixed Ef) | 1 / 4 / 32 | 2 / 0 / 0 | 1.3e-9 / 6.6e-14 / 1.0e-12 |
| `scc/C60` (Fermi 400 K) | 1 / 4 / 32 | 0 / 0 / 0 | 4.3e-14 / 4.3e-14 / 4.3e-14 |
| `solvers/Si32C32_ELPA2` (0 K) | 1 / 4 / 32 | 0 / 0 / 0 | 3.0e-13 / 2.0e-13 / 3.0e-13 |
| `solvers/Fe4_ELPA_2group` (MP 500 K, spin, 2 groups) | 1 / 4 / 32 | 2 / 0 / 0 | 4.8e-8 / 1.7e-9 / 0.0 |

The Fe4 case at `EmptyStates = 1` converges in 48 iterations against 47 for the
full spectrum, to the same energy, after two adaptive increases.

New `solvers/Si32C32_ELPA2_partial` regression test: same system at 2000 K
starting from one empty state, which takes four increases and ends up
calculating 144 of 256 states. Its reference tag is the output of the identical
input with the `EmptyStates` line removed, so the test asserts exactly the
property of interest. Verified at 1, 2 and 4 MPI processes.

New unit test collection `test/src/dftbp/unit/elecsolvers` covering the occupied
state count for the three spin cases, the energy windows, the doubling and its
saturation, and the sufficiency test across k-points and spin channels.

## Open question

Eigenvalues above the calculated range are left at zero, since the arrays are
sized for the full basis. They have zero occupation so nothing physical depends
on them, but they are written to `band.out` and `detailed.out` as a block of
levels at 0 eV, which is misleading. I have documented it in the manual for now.
Slicing the output to the calculated states, or filling the remainder with a
clearly invalid value, are both easy to add if you have a preference.


doc/dftb+/manual/dftbp.tex
```diff
@@ -2530,6 +2531,33 @@ \subsubsection{ELPA}
 \item[\is{AutotuneFile}] The ELPA autotune state is stored in this file.
   If present at startup, the autotuning process is resumed using its previous
   state. (default: \is{elpa\_autotune\_state.out})
+\item[\is{EmptyStates}] Number of empty states to calculate in addition to the
```
bhourahine:
```suggestion
\item[\is{EmptyStates}] Number of empty states to calculate, in addition to the
```

doc/dftb+/manual/dftbp.tex
```diff
@@ -2530,6 +2531,33 @@ \subsubsection{ELPA}
...
+  states needed to hold the electrons of the system. A negative value (the
+  default) requests the complete spectrum, which is the usual behaviour.
+
+  A positive value switches on partial diagonalisation, where the solver only
+  returns the lowest states. Since the states left out are unoccupied,
```
bhourahine:
```suggestion
  returns the lowest occupied states. Since the states left out are unoccupied,
```

doc/dftb+/manual/dftbp.tex
```diff
@@ -2530,6 +2531,33 @@ \subsubsection{ELPA}
...
+  they do not contribute to the density matrix, and calculating them is
+  avoidable work. The count given here is only a starting value: whenever the
```
bhourahine:
```suggestion
  avoidable work. The count you provide here is only a starting value: whenever the
```

doc/dftb+/manual/dftbp.tex
```diff
@@ -2530,6 +2531,33 @@ \subsubsection{ELPA}
...
+  highest calculated state is not far enough above the chemical potential for
```
bhourahine:
```suggestion
  highest calculated state is not far enough above the electronic chemical potential for
```

doc/dftb+/manual/dftbp.tex
```diff
@@ -2530,6 +2531,33 @@ \subsubsection{ELPA}
...
+  its occupation to be negligible, the number of empty states is doubled and
+  the self-consistent cycle continues, so that the calculation cannot converge
+  while states which might be occupied are missing. The distance required above
```
bhourahine:
```suggestion
  while states which should be (partially) occupied are missing. The energy of states required above
```

doc/dftb+/manual/dftbp.tex
```diff
@@ -2530,6 +2531,33 @@ \subsubsection{ELPA}
...
+  the chemical potential is the point at which the occupation of a state falls
+  below $10^{-10}$, which is 24 times the broadening width for \is{Fermi}
```
bhourahine:
```suggestion
  below $10^{-10}$ e, which is 24 times the broadening width for \is{Fermi}
```

doc/dftb+/manual/dftbp.tex
```diff
@@ -2530,6 +2531,33 @@ \subsubsection{ELPA}
...
+  filling and 5 times for the \is{Gaussian} and \is{MethfesselPaxton} schemes.
+
+  This option is currently only available when ELPA is called through the ELSI
+  library (see \is{PreferElsi}), and cannot be combined with calculations which
+  need the states that would be left out. These are excited state and
```
bhourahine:
```suggestion
  need the states that would be left out. These are currently excited state and
```

doc/dftb+/manual/dftbp.tex
```diff
@@ -2530,6 +2531,33 @@ \subsubsection{ELPA}
...
+  perturbation calculations, electron dynamics, $\Delta$DFTB, custom
+  occupations, non-collinear spin, and writing the eigenvectors. Note also that
+  eigenvalue output above the calculated states is not meaningful.
```
bhourahine:
```suggestion
  eigenvalues outputted above the calculated states are not meaningful.
```

doc/dftb+/manual/dftbp.tex
```diff
@@ -2530,6 +2531,33 @@ \subsubsection{ELPA}
...
+
+  The value given should be a sensible guess for the system, rather than the
+  smallest number which happens to work. Starting from too few states costs
+  extra iterations, and for systems with more than one self-consistent solution
```
bhourahine:
```suggestion
  extra iterations, and for systems with more than one possible self-consistent solution,
```

test/app/dftb+/solvers/Si32C32_ELPA2_partial/dftb_in.hsd
```diff
@@ -0,0 +1,116 @@
...
+  }
+  MaxAngularMomentum = {
+    Si = "p"
+    C = "p"
+  }
+  Filling = Fermi {
+    Temperature [Kelvin] = 2000.0
+  }
+  SlaterKosterFiles = {
+    Si-Si = "./Si-Si.skf"
```
bhourahine, line 93:
```suggestion
    Prefix = "Prefix = "slakos/origin/pbc-0-3/""
    Si-Si = "./Si-Si.skf"
```
This would allow the symlinks for the individual files to be removed.

test/app/dftb+/solvers/Si32C32_ELPA2_partial/dftb_in.hsd
```diff
@@ -0,0 +1,116 @@
...
+Analysis = {
+  CalculateForces = Yes
+}
+
+Options = {
+  WriteAutotestTag = Yes
+}
+
+ParserOptions = {
+  ParserVersion = 5
```
bhourahine:
```suggestion
  ParserVersion = 14
```
as the current version

test/app/dftb+/solvers/Si32C32_ELPA2_partial/dftb_in.hsd
```diff
@@ -0,0 +1,116 @@
...
+    C-Si = "./C-Si.skf"
+    C-C = "./C-C.skf"
+  }
+  KPointsAndWeights = {
+   0.0 0.0 0.0 1.0
+  }
+}
+
+Analysis = {
+  CalculateForces = Yes
```
bhourahine:
```suggestion
  PrintForces = Yes
```

src/dftbp/dftbplus/initprogram.F90
```diff
@@ -7589,4 +7601,71 @@ end subroutine overrideContactCharges
 #:endif
 
 
+  !> Sets up the state counting for a partial diagonalisation, checking that the requested
+  !> calculation does not need the states which would be left out.
```
bhourahine:
```suggestion
  !> calculation does not need the states which would be left out in partial diagonalisation.
```

src/dftbp/dftbplus/initprogram.F90
```diff
@@ -7589,4 +7601,71 @@ end subroutine overrideContactCharges
...
+  subroutine initPartialDiag_(partialDiag, ctrl, nOrb, nEl, nSpin, isEigenSolver)
```
bhourahine, line 7606: Should this not also test if the calculation is required to be self-consistent? As the algorithm requires a self-consistent cycle.

src/dftbp/dftbplus/main.F90
```diff
@@ -1547,6 +1532,13 @@ subroutine processGeometry(this, env, iGeoStep, iLatGeoStep, tWriteRestart, tSto
             & tStopScc, errStatus)
         if (errStatus%hasError()) call error(errStatus%message)
 
+        if (.not. this%electronicSolver%partialDiag%isSufficient) then
+          ! the states calculated so far do not reach far enough above the chemical potential for
```
bhourahine:
```suggestion
          ! the states calculated so far do not reach high enough above the chemical potential for
```

src/dftbp/dftbplus/main.F90
```diff
@@ -3394,6 +3397,54 @@ subroutine getDensityFromDenseDiag(env, denseDesc, ints, neighbourList, symNeigh
   end subroutine getDensityFromDenseDiag
 
 
+  !> Asks the solver for more empty states, after the last diagonalisation was found not to have
+  !> returned enough of them.
+  !>
+  !> The charge mixer is restarted as well: its history was built from densities which were missing
```
bhourahine: Possibly using incorrect densities is not so bad, this should eventually be tested.

src/dftbp/elecsolvers/elecsolvers.F90
```diff
@@ -197,6 +202,26 @@ pure function providesEigenvalues(iSolver)
...
+    !> Number of states to calculate from now on
+    integer, intent(in) :: nState
+
+    if (this%isElsiSolver) then
+      call this%elsi%setNState(nState)
+    end if
+
+    ! The solver instance is built again for the new number of states, so it will factorise the
+    ! overlap once more. That needs the original matrix, not the stored factorisation.
+    this%hasCholesky(:) = .false.
```
bhourahine, line 220: I'm not sure if this is needed, as the Choleskii factorisation is distinct from the regular eigensolution (I will check the ELPA documentation).

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+  !> otherwise leave a systematic error behind which the self-consistent cycle cannot remove.
+  real(dp), parameter :: negligibleFilling = 1.0e-10_dp
+
+  !> Multiple of the broadening width above the chemical potential at which the Fermi-Dirac
+  !> occupation 1 / (1 + exp(x)) falls below negligibleFilling, which is ln(1 / f - 1) = 23.0,
+  !> rounded up.
+  real(dp), parameter :: nWidthsFermi = 24.0_dp
+
+  !> Same for the Gaussian and Methfessel-Paxton schemes, where the occupation follows
+  !> erfc(x) / 2, reaching negligibleFilling at 4.6 widths.
```
bhourahine: It is actually easier to obtain from the underlying gaussian broadening function of the level, which is just exp(-x**2),  instead of the filling function which as you note requires the erfc.
Also:
```suggestion
  !! erfc(x) / 2, reaching negligibleFilling at 4.6 widths.
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+    !> Number of states to calculate
+    integer :: nState
+
+    nState = min(this%nOccupied + this%nEmpty, this%nOrb)
+
+  end function TPartialDiag_getNState
+
+
+  !> Tests whether the calculated states extend far enough above the chemical potential that the
+  !> states which were not calculated are certainly empty.
```
bhourahine:
```suggestion
  !! states which were not calculated are certainly empty.
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+  !>
```
bhourahine:
```suggestion
  !!
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+  !> The test is applied to the highest calculated state of every k-point and spin channel, as the
```
bhourahine:
```suggestion
  !! The test is applied to the highest calculated state of every k-point and spin channel, as the
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+  !> spectra of these are independent of each other.
```
bhourahine:
```suggestion
  !! spectra of these are independent of each other.
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+!  DFTB+: general package for performing fast atomistic simulations                                !
+!  Copyright (C) 2006 - 2025  DFTB+ developers group                                               !
+!                                                                                                  !
+!  See the LICENSE file for terms of usage and distribution.                                       !
+!--------------------------------------------------------------------------------------------------!
+
+#:include 'common.fypp'
+
+!> Bookkeeping for partial diagonalisation, where the eigensolver is only asked for the occupied
+!> states together with a limited number of empty states.
```
bhourahine:
```suggestion
!! states together with a limited number of empty states.
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+!>
```
bhourahine:
```suggestion
!!
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+!> The number of empty states is chosen by the user as a starting value and then increased during
```
bhourahine:
```suggestion
!! The number of empty states is chosen by the user as a starting value and then increased during
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+!> the self-consistent cycle whenever the states obtained are found to be insufficient. The test for
```
bhourahine:
```suggestion
!! the self-consistent cycle whenever the states obtained are found to be insufficient. The test for
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+!> sufficiency is that the highest calculated state lies far enough above the chemical potential for
```
bhourahine:
```suggestion
!! sufficiency is that the highest calculated state lies far enough above the chemical potential for
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+!> its occupation to be numerically negligible, the distance being a multiple of the broadening
```
bhourahine:
```suggestion
!! its occupation to be numerically negligible, the distance being a multiple of the broadening
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+!> width of the filling function in use.
```
bhourahine:
```suggestion
!! width of the filling function in use.
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+  use dftbp_dftb_etemp, only : fillingTypes
+  implicit none
+
+  private
+  public :: TPartialDiag, TPartialDiag_init
+  public :: getEmptyStateWindow, getNOccupied
+
+
+  !> Occupation small enough to be neglected when deciding whether a state is empty. It has to be
+  !> well below the accuracy to which the charges are converged, as the neglected states would
```
bhourahine:
```suggestion
  !! well below the accuracy to which the charges are converged, as the neglected states would
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+  !> otherwise leave a systematic error behind which the self-consistent cycle cannot remove.
```
bhourahine:
```suggestion
  !! otherwise leave a systematic error behind which the self-consistent cycle cannot remove.
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+  real(dp), parameter :: negligibleFilling = 1.0e-10_dp
+
+  !> Multiple of the broadening width above the chemical potential at which the Fermi-Dirac
+  !> occupation 1 / (1 + exp(x)) falls below negligibleFilling, which is ln(1 / f - 1) = 23.0,
```
bhourahine:
```suggestion
  !! occupation 1 / (1 + exp(x)) falls below negligibleFilling, which is ln(1 / f - 1) = 23.0,
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+  !> rounded up.
```
bhourahine:
```suggestion
  !! rounded up.
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+      return
+    end if
+
+    this%nEmpty = min(max(2 * this%nEmpty, 1), this%nOrb - this%nOccupied)
+
+  end subroutine TPartialDiag_grow
+
+
+  !> Distance above the chemical potential beyond which the occupation of a state is numerically
+  !> negligible for the given filling function.
```
bhourahine:
```suggestion
  !! negligible for the given filling function.
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+      window = nWidthsFermi * tempElec
+    else
+      window = nWidthsGaussian * tempElec
+    end if
+
+  end function getEmptyStateWindow
+
+
+  !> Number of states required to hold the electrons of the system.
+  !>
```
bhourahine:
```suggestion
  !!
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+  !> For collinear spin the two channels are diagonalised separately, so the larger of the two
```
bhourahine:
```suggestion
  !! For collinear spin the two channels are diagonalised separately, so the larger of the two
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+  !> channels determines the count. Where the Fermi level is shared between the channels, electrons
```
bhourahine:
```suggestion
  !! channels determines the count. Where the Fermi level is shared between the channels, electrons
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+  !> can move between them during the self-consistent cycle, which the adaptive increase of the
```
bhourahine:
```suggestion
  !! can move between them during the self-consistent cycle, which the adaptive increase of the
```

src/dftbp/elecsolvers/partialdiag.F90
```diff
@@ -0,0 +1,230 @@
...
+  !> number of empty states takes care of.
```
bhourahine:
```suggestion
  !! number of empty states takes care of.
```

bhourahine (commented):
Looks good, but I have some comments (mostly minor stylistic about Doxygen comments). But in particularly see the one about checking for tScc is false, as this probably needs to be blocked or handled separately.
Decay of the filling functions to a value of order machine double precision epsilon for Fermi filling is around 36 * kT and for gaussian (i.e. the slowest of the Methfessle-Paxton family) 6 * kT. So it might be a good idea to make a follow-up PR to either make this is an input parameter instead of hard coded or to set the minimum required filling as a tolerance.

ZMYsamuel:
Not a review — just flagging that I posted a static empty-states sweep on #1886 (8000-atom SiC, 32k basis, 40 K Fermi: full spectrum 2792 s vs 2267 s at 64 empty states, converged energy identical) in case a measured data point is useful while you settle the growth heuristic. The one thing it cannot tell you is where the lower bound is — I never found the failure point, which is the gap your `TPartialDiag_check` closes. Happy to run anything specific on that case if it would help.

src/dftbp/dftbplus/initprogram.F90
RizzoHou: You are right, it needed blocking. `EmptyStates` is now rejected for a non-SCC calculation, and the manual says so.

It would have been worse than an unsupported case. Without the self-consistent cycle the count can never grow, so too small a value would have silently truncated the occupied states. On top of that `growPartialDiag` is still reached on the single iteration, and in the version you reviewed it restarted the charge mixer, which is only allocated when `requiresMixer` holds in `initprogram.F90` — so the non-SCC path would have gone through an unallocated mixer.

The other two conditions in `requiresMixer` were already rejected here: REKS explicitly, and `tRestartNoSC` is only set for the electron dynamics restart, which is rejected too. With the SCC test added, nothing that reaches `growPartialDiag` can be missing the mixer, so I have not added a separate guard for it. (The mixer restart itself is gone now, see the other thread.)

Supporting the non-SCC case properly would need a re-diagonalisation loop around the single solve, rather than relying on the SCC cycle. I am happy to add that as a follow-up if you think it is worth having.

Going back over the guard list after your comment turned up two more holes, both now closed the same way:

- **`ProjectStates`** was not rejected, although it is the same class of problem as `WriteEigenvectors`, which was. `writeProjectedEigenvectors` projects all `nOrb` states, the solver only fills the calculated ones, and `eigVecsReal` is allocated without initialisation — so the projections above `nState` came from uninitialised memory rather than from anything merely meaningless.
- **`IndependentKFilling`** was not rejected, and it defeats the sufficiency test rather than the output. With it set, `getFillingsAndBandEnergies` accumulates `Ef(:) = Ef + EfTmp * kWeights(iK)`, so what `check` receives is the weighted average over k-points. A k-point whose own chemical potential sits above that average could then satisfy the test with occupied states missing, and the count would never grow. Handling it would mean passing the per-k Fermi levels down; rejecting seemed the honest thing to do for now.

All four rejections are exercised on a machine with ELSI: each is refused with its own message and a non-zero exit, while the unmodified input still runs and takes its four growth steps.

src/dftbp/elecsolvers/elecsolvers.F90
RizzoHou: The factorisation is distinct from the eigensolution, but the flag also decides which matrix is handed to the solver. In `diagDenseRealMtxBlacs`, a set `hasCholesky` makes `getCholesky` overwrite `SSqr` with the stored factor before the solver is called, and on the ELSI path there is no `skipchol` equivalent to tell it — as the comment there notes, ELSI factorises only on the first call for a given handle.

Rebuilding the handle makes the next call a first call again, so ELSI does factorise, but with `hasCholesky` still set it receives the previous factor instead of the overlap. That is what failed with `Cholesky factorization failed ierr -1`, on the first growth step, every time. Clearing the flag lets the freshly unpacked overlap through, and the new handle stores its own factor on the way out.

It could equally live inside `TElsiSolver_setNState`, next to the handle being destroyed, if you would rather the invalidation sat with its cause. The only reason it is in the caller is that `hasCholesky` belongs to `TElectronicSolver` rather than to the ELSI solver.

src/dftbp/dftbplus/main.F90
RizzoHou: Tested, and you were right — the restart is not needed, so it is gone.

I built the same code with and without the `resetChargeMixer` call and ran the three cases which actually grow. Keeping the history is not merely harmless, it is cheaper:

| case (starting from 1 empty state) | growth steps | SCC iterations with restart | without |
|---|---|---|---|
| `solvers/Si32C32_ELPA2_partial` (Fermi 2000 K) | 4 | 10 | 8 |
| `scc/C60_Fermi` (Fermi 1000 K, fixed Ef) | 2 | 15 | 11 |
| `solvers/Fe4_ELPA_2group` (MP 500 K, spin, 2 groups) | 2 | 2294 | 2266 |

The Fe4 figures are summed over the geometry optimisation, which takes the same
50 steps either way, so they are comparing like with like.

The autotest tags of the two builds agree in every field (worst deviation 1.0e-7 on Fe4, 4.4e-9 and 1.3e-9 on the other two), so the extrapolation across an increase converges to the same fixed point and simply gets there sooner. The intuition that the residuals are not comparable is presumably outweighed by the fact that both sets of densities are heading for the same place, and the increase happens early, while the history is short anyway.

Dropping the restart also let me revert the extraction of `resetChargeMixer` out of the SCC setup, so the patch no longer touches that part of `main.F90` at all.

Caveat worth stating: this is three cases, and none of them is a difficult metal. If you would rather keep the restart as the conservative choice until something harder has been tried, say so and I will put it back.

src/dftbp/elecsolvers/partialdiag.F90
RizzoHou: Much better, thank you — changed. Bounding by the broadening function itself rather than its integral also happens to be the conservative direction: `exp(-x**2)` reaches 1e-10 at 4.8 widths against 4.6 for `erfc(x)/2`, so the window is slightly wider than strictly needed and the rounded constant is unchanged at 5.

src/dftbp/dftbplus/initprogram.F90
RizzoHou: You are right, it needed blocking. `EmptyStates` is now rejected for a non-SCC calculation, and the manual says so.

It would have been worse than an unsupported case. Without the self-consistent cycle the count can never grow, so too small a value would have silently truncated the occupied states. On top of that `growPartialDiag` is still reached on the single iteration, and in the version you reviewed it restarted the charge mixer, which is only allocated when `requiresMixer` holds in `initprogram.F90` — so the non-SCC path would have gone through an unallocated mixer.

The other two conditions in `requiresMixer` were already rejected here: REKS explicitly, and `tRestartNoSC` is only set for the electron dynamics restart, which is rejected too. With the SCC test added, nothing that reaches `growPartialDiag` can be missing the mixer, so I have not added a separate guard for it. (The mixer restart itself is gone now, see the other thread.)

Supporting the non-SCC case properly would need a re-diagonalisation loop around the single solve, rather than relying on the SCC cycle. I am happy to add that as a follow-up if you think it is worth having.

Going back over the guard list after your comment turned up two more holes, both now closed the same way:

- **`ProjectStates`** was not rejected, although it is the same class of problem as `WriteEigenvectors`, which was. `writeProjectedEigenvectors` projects all `nOrb` states, the solver only fills the calculated ones, and `eigVecsReal` is allocated without initialisation — so the projections above `nState` came from uninitialised memory rather than from anything merely meaningless.
- **`IndependentKFilling`** was not rejected, and it defeats the sufficiency test rather than the output. With it set, `getFillingsAndBandEnergies` accumulates `Ef(:) = Ef + EfTmp * kWeights(iK)`, so what `check` receives is the weighted average over k-points. A k-point whose own chemical potential sits above that average could then satisfy the test with occupied states missing, and the count would never grow. Handling it would mean passing the per-k Fermi levels down; rejecting seemed the honest thing to do for now.

All four rejections are exercised on a machine with ELSI: each is refused with its own message and a non-zero exit, while the unmodified input still runs and takes its four growth steps.

src/dftbp/elecsolvers/elecsolvers.F90
RizzoHou: The factorisation is distinct from the eigensolution, but the flag also decides which matrix is handed to the solver. In `diagDenseRealMtxBlacs`, a set `hasCholesky` makes `getCholesky` overwrite `SSqr` with the stored factor before the solver is called, and on the ELSI path there is no `skipchol` equivalent to tell it — as the comment there notes, ELSI factorises only on the first call for a given handle.

Rebuilding the handle makes the next call a first call again, so ELSI does factorise, but with `hasCholesky` still set it receives the previous factor instead of the overlap. That is what failed with `Cholesky factorization failed ierr -1`, on the first growth step, every time. Clearing the flag lets the freshly unpacked overlap through, and the new handle stores its own factor on the way out.

It could equally live inside `TElsiSolver_setNState`, next to the handle being destroyed, if you would rather the invalidation sat with its cause. The only reason it is in the caller is that `hasCholesky` belongs to `TElectronicSolver` rather than to the ELSI solver.

src/dftbp/dftbplus/main.F90
RizzoHou: Tested, and you were right — the restart is not needed, so it is gone.

I built the same code with and without the `resetChargeMixer` call and ran the three cases which actually grow. Keeping the history is not merely harmless, it is cheaper:

| case (starting from 1 empty state) | growth steps | SCC iterations with restart | without |
|---|---|---|---|
| `solvers/Si32C32_ELPA2_partial` (Fermi 2000 K) | 4 | 10 | 8 |
| `scc/C60_Fermi` (Fermi 1000 K, fixed Ef) | 2 | 15 | 11 |
| `solvers/Fe4_ELPA_2group` (MP 500 K, spin, 2 groups) | 2 | 2294 | 2266 |

The Fe4 figures are summed over the geometry optimisation, which takes the same
50 steps either way, so they are comparing like with like.

The autotest tags of the two builds agree in every field (worst deviation 1.0e-7 on Fe4, 4.4e-9 and 1.3e-9 on the other two), so the extrapolation across an increase converges to the same fixed point and simply gets there sooner. The intuition that the residuals are not comparable is presumably outweighed by the fact that both sets of densities are heading for the same place, and the increase happens early, while the history is short anyway.

Dropping the restart also let me revert the extraction of `resetChargeMixer` out of the SCC setup, so the patch no longer touches that part of `main.F90` at all.

Caveat worth stating: this is three cases, and none of them is a difficult metal. If you would rather keep the restart as the conservative choice until something harder has been tried, say so and I will put it back.

src/dftbp/elecsolvers/partialdiag.F90
RizzoHou: Much better, thank you — changed. Bounding by the broadening function itself rather than its integral also happens to be the conservative direction: `exp(-x**2)` reaches 1e-10 at 4.8 widths against 4.6 for `erfc(x)/2`, so the window is slightly wider than strictly needed and the rounded constant is unchanged at 5.

RizzoHou:
Thanks for the review. Pushed, with the individual points answered in their own threads. Summary of what changed:

- **Non-SCC calculations are now rejected.** This was the real bug you spotted — details in that thread, including why it would have crashed rather than merely being unsupported. Re-reading the guard list because of your comment turned up two more that had to go the same way: `ProjectStates`, which was reading uninitialised eigenvector columns, and `IndependentKFilling`, which quietly defeats the sufficiency test because the chemical potential reaching it is the k-weighted average. Both are described in that thread, and all four rejections are now exercised on a machine with ELSI.
- **Rebuilding the ELSI handle no longer loses the electronic broadening.** `elsi_set_mu_broaden_*` is applied from `updateElectronicTemp`, which runs once per geometry step, before the SCC loop — so after the first increase the new handle was back on the library default. The width is now remembered and re-applied on a fresh handle. Dormant for the ELPA eigensolver, which fills the states itself, but `setNState` is solver agnostic and this PR is what makes a second `elsi_init` reachable at all.
- **The charge mixer restart is gone**, having measured it as you suggested. Without it the three growing cases converge in *fewer* iterations to the same answer, so the restart was only costing work. That also let me revert the `resetChargeMixer` extraction, so `main.F90` no longer touches the SCC setup.
- **Doxygen continuation lines** changed to `!!` throughout the new code, not only the lines you marked.
- **The Gaussian window comment** now derives the multiplier from `exp(-x**2)` as you suggested, which is both simpler and slightly conservative relative to the erfc form.
- **Manual and test input** wording and keyword suggestions applied.

Two things worth flagging.

**The test input needed more than the parser version bump.** Moving `ParserVersion` from 5 to 14 also required renaming `Eigensolver` to `Solver` — the version-5 input was relying on the 6→7 conversion to do that, and at 14 the block is simply ignored and the run stops. The `Prefix` change worked as you described and the four symlinks are gone. Both are verified by running the test, since it needs ELSI and so never executes in CI; it passes against the committed reference tag at 1, 2 and 4 MPI processes, taking the same four growth steps as before. Note that 14 is not quite current — `oldcompat` maps 26.1 to 15 — so the run prints a 14→15 conversion notice. Nothing in the input is affected by that conversion, and 14 is what the rest of the test suite uses, so I have left it at your suggested value; say the word if you would prefer 15.

**On the multipliers**, I agree with making them adjustable, and your machine-epsilon figures (36 kT and 6 kT) line up with the 24 and 5 here — same construction, just a less strict target than epsilon. What I would suggest exposing is the target occupation rather than the multiplier itself, i.e. the `negligibleFilling = 1e-10` constant, since each filling function turns that one number into its own window, and a user picking "how small an occupation may I neglect" is choosing something with a physical meaning, whereas a multiplier only means something once you know which filling function is in use. Happy to add it here or to keep it for a follow-up — whichever you prefer.

The open question from the description still stands, by the way: eigenvalues above the calculated range are written to `band.out` and `detailed.out` as a block of levels at 0 eV. I can slice the output to the calculated states, or fill the remainder with something clearly invalid, if you have a preference.

RizzoHou:
@ZMYsamuel thanks for posting it — a measured point at that size is useful, and the honesty about what the sweep does not establish is more useful still.

The part I would find valuable, if you are offering: run the same cell from `EmptyStates = 1` and report how many growth steps it takes and where it lands. That is the lower bound neither of us has — the growth is logged as `Too few empty states, increasing to N`, so the sequence of increases tells you the smallest buffer the check would have accepted, without having to bisect for a failure. At 40 K with a gap I would expect it to stop almost immediately, in which case your 64 was already generous by a wide margin and the interesting comparison is against whatever the check settles on.

If you are willing to also report the highest calculated eigenvalue against the chemical potential at that point, that is the quantity the sufficiency test uses, and a data point at 32,000 basis functions would be better than anything I can produce here.

ZMYsamuel:
Thanks — understood. I’ll find time to run the 32k-basis case starting from `EmptyStates = 1` and report the growth sequence, the final accepted state count, and the highest calculated eigenvalue relative to the chemical potential.

PR #1895, opened 2026-07-30
Added missing initialization in setRefCharges

terminationshock:
The API call to `dftbp_set_ref_charges` lead to an inconsistent state if the lattice had been changed before.

Altering the lattice and then changing the charges with a sequence `dftbp_get_ref_charges` and `dftbp_set_ref_charges` makes the SCC cycles fail to converge. The bug does not appear if the lattice is kept constant. This is the reason why it has not been covered by tests. To test this now, I altered the `treeinit` test. Let me know if this is acceptable or whether you rather would like to have a separate test.


bhourahine (commented):
Looks good as an initial fix for changing the reference charge with setElectronNumber(). Unfortunately it also resets the charge of the system to be neutral, irrespective of any prior call(s) to setElectronNumber(), so order of these two then matter.

src/dftbp/dftbplus/mainapi.F90
```diff
@@ -414,6 +414,10 @@ subroutine setRefCharges(env, main, q0)
 
     call updateReferenceShellCharges(main%qShell0, main%q0, main%orb, main%nAtom, main%species0)
 
+    if (allocated(main%qInput)) then
```
bhourahine:
```suggestion
```

src/dftbp/dftbplus/mainapi.F90
```diff
@@ -414,6 +414,10 @@ subroutine setRefCharges(env, main, q0)
...
+      main%qInput(:,:,:) = main%q0
```
bhourahine:
```suggestion
```

src/dftbp/dftbplus/mainapi.F90
```diff
@@ -414,6 +414,10 @@ subroutine setRefCharges(env, main, q0)
...
+    end if
```
bhourahine:
```suggestion
```

src/dftbp/dftbplus/mainapi.F90
bhourahine, line 419:
```suggestion
    if (allocated(main%qInput)) then
      main%qInput(:,:,:) = main%qInput + q0 - main%q0
    end if

    main%q0(:,:,:) = q0
```

bhourahine (commented):
One possible way to address this in the suggestions. The main place where this would be noticeable would be small changes in reference charges when there is already a self-consistent solution.

terminationshock:
> Looks good as an initial fix for changing the reference charge with setElectronNumber(). Unfortunately it also resets the charge of the system to be neutral, irrespective of any prior call(s) to setElectronNumber(), so order of these two then matter.

Thanks for this hint. I could introduce a flag for this, which is not very clean, though. Do you have any suggestions?

bhourahine:
I'm now seeing some strange behaviour with my suggested changes (your original choice to initialise qInput = q0 seems to be OK, but obviously then requires a few SCC cycles).

terminationshock:
I just tested your suggestion. It seems to be fine. I don't see any issues on my side.

bhourahine:
> I just tested your suggestion. It seems to be fine. I don't see any issues on my side.

I've just pushed a fix to this branch: there are also reduced input charges used for the SCC convergence test, so these need to be updated as well. I'll squash and merge shortly.

bhourahine:
@terminationshock I've just pushed a version rebased on upstream main, the treeinit test needed to be recalculated as part of the changes for the reduced charges. Can you have a quick look at the current state of the branch, and if happy I will merge.

terminationshock:
It looks good to me. I also tested this again in my setup. Can be merged from my point of view.

bhourahine (approved):

PR #1900, opened 2026-08-04
Remove redundant index sum in transcharges

bhourahine:
The work array should only be over one spin channel. The array was reset to zero each time in the loop, so was working correctly, but only the current spin is referenced. So it was using twice the memory and MPI reducing over a matrix which was half full of zeros.


bhourahine:
There is probably room to optimise further and avoid the nAtom x full matrix GEMM operations to set up the charges. MPI windows and one sided communications might be the best option. But putting the sTimesGrndEigVecs into a globally replicated array and then locally making the products with the distributed grndEigVecs, followed by a global sum over the qCache arrays would also probably work faster than the current structure.

thomas-niehaus (approved):
Thanks for fixing this.

thomas-niehaus:
Yes, I agree,  sTimesGrndEigVecs should be distributed.

PR #1901, opened 2026-08-05
Clean out redundant env use in RPA

bhourahine:


bhourahine:
@thomas-niehaus was there a plan to use env in some way for these routine? Otherwise, makes sense to remove it, as is doing nothing.

thomas-niehaus:
I can't remember why it was necessary at some point. Can be removed.

thomas-niehaus (approved):

PR #1903, opened 2026-08-16
More accurate kWeight normalisation

bhourahine:
For large k-grids, underflow affects the normalisation of sum(kWeights). Also block compiler optimisation of Kahan sum.


bhourahine:
@aradi Any views about the use of volatile, as the more accurate sum would be useful elsewhere.

aradi (approved):

PR #1909, opened 2026-09-04
Fix pump probe bug

fbonafe:
In restart runs, initializeTDVariables() received ham0 as intent(out) and unconditionally reset it from H0, which is built for the geometry given in the input file. On a restart the geometry instead comes from the dump file, and updateH0S() had just rebuilt ham0 correctly for those coordinates (only for that to be overwritten again). The stale ham0 was then used to build the first Hamiltonian, the one that bootstraps the leapfrog propagator, so it was evaluated at the wrong geometry. Only the inter-atomic blocks are affected (onsite blocks do not depend on geometry) and the error is sub-percent, but the bootstrap's imaginary displacement is a near-cancelling commutator [S⁻¹H, Re ρ], which amplifies it by roughly an order of magnitude; rho and rhoOld are left mutually inconsistent and this excites the leapfrog odd–even mode at period 2dt. It shows up as a large spurious ripple on the probe dipole (~30x the physical signal here) and requires ion dynamics (with fixed nuclei the restart geometry equals the input geometry and the stale ham0 happens to be correct). 

Fix: ham0 becomes intent(inout) and is only allocated and set from H0 when not restarting, so the values computed by updateH0S() for the restarted geometry survive; since it now persists across calls (runDynamics() calls the routine once per kick polarisation direction), the allocation is guarded with if (.not. allocated(ham0)).


bhourahine:
@fbonafe can you add this to the fixed section of the CHANGELOG.rst, as it will impact users.

src/dftbp/timedep/timeprop.F90
```diff
@@ -2469,6 +2468,12 @@ subroutine initializeTDVariables(this, env, densityMatrix, rho, H1, Ssqr, Sinv,
     end if
 
     if (.not. this%tReadRestart) then
+      ! ham0 is intent(inout), so it survives between calls (runDynamics() calls this routine
+      ! once per kick polarisation direction) and must only be allocated the first time.
```
bhourahine, line 2472: Is it possible to check the kick direction is consistent with allocation status?

bhourahine (commented):
Plus suggestion about the CHANGELOG

src/dftbp/timedep/timeprop.F90
fbonafe: No, because the kick direction could be set independently of this. I just changed the comment because it was misleading. The thing is actually not to override with the ham0 that was just built from restarted coordinates with the GS one.

bhourahine: I fixed the line length, and added the mentioned check above the routine in the call tree.

PR #1913, opened 2026-09-09
Recalculate xTB references with tigthened SCC tolerance

aradi:
Needed as scc convergence changed in current tblite, (v0.7.0) causing false positives for the test cases with default scc tolerance (1-5).


bhourahine (approved):
Other than a typo in the log message (tigthened -> tightened), LGTM.
Probably when TBLITE 0.7 is merged, will need a note in the CHANGELOG about their tolerances.

PR #1916, opened 2026-09-10
Download test data only if necessary

aradi:
Fixes current automatic test-data, which is triggered at each new configuration, even if the data is present.
The current approach only downloads the data when not present yet.

### DIFF
--- a/cmake/DftbPlusUtils.cmake
+++ b/cmake/DftbPlusUtils.cmake
@@ -722,3 +722,49 @@ function (dftbp_check_minimal_compiler_version lang compiler_versions)
     endif()
   endwhile()
 endfunction()
+
+
+# Download the external data (SK-files, GBSA-data) needed for testing, if not present yet.
+function (dftbp_download_test_data)
+  _dftbp_download_test_data_from_github(
+    dftbplus
+    testparams
+    1f5aefe45b7951f3fd1c0b63e40b80f80fdab612
+    external/slakos
+    "SK-files"
+  )
+  _dftbp_download_test_data_from_github(
+    grimme-lab
+    gbsa-parameters
+    6836c4d997e4135e418cfbe273c96b1a3adb13e2
+    external/gbsa
+    "GBSA-data"
+  )
+endfunction ()
+
+
+# Helper function to download external test data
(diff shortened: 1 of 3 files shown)


cmake/DftbPlusUtils.cmake
```diff
@@ -722,3 +722,49 @@ function (dftbp_check_minimal_compiler_version lang compiler_versions)
...
   endwhile()
 endfunction()
+
+
+# Download the external data (SK-files, GBSA-data) needed for testing, if not present yet.
+function (dftbp_download_test_data)
+  _dftbp_download_test_data_from_github(
+    dftbplus
+    testparams
+    1f5aefe45b7951f3fd1c0b63e40b80f80fdab612
```
bhourahine, line 732: Doesn't this require the same tag as also present in test/CMakeLists.txt ?
Can this and the gbsa tags not be defined in a single place?

bhourahine: Ah, just read further, please ignore.

cmake/DftbPlusUtils.cmake
```diff
@@ -722,3 +722,49 @@ function (dftbp_check_minimal_compiler_version lang compiler_versions)
...
+      if (EXISTS "${targetdirpath}/origin.tar.gz")
+        file(REMOVE "${targetdirpath}/origin.tar.gz")
+      endif ()
+      message(FATAL_ERROR "Download of ${content} failed: ${status_message} (code: ${status_code})")
+    endif ()
+    file(ARCHIVE_EXTRACT INPUT "${targetdirpath}/origin.tar.gz" DESTINATION "${targetdirpath}")
+    file(RENAME "${targetdirpath}/${repository}-${commit_id}" "${targetdirpath}/origin")
+    file(REMOVE "${targetdirpath}/origin.tar.gz")
+  endif ()
+endfunction ()
```
bhourahine, line 770: Missing newline character on last line.

bhourahine:
```suggestion
endfunction ()

```

bhourahine (approved):

PR #1917, opened 2026-09-10
Upgrade tblite software stack to latest releases

aradi:
- tblite v0.7.0
- dftd4 v4.2.0
- mctc v0.5.2
- mstore v0.3.0
- multicharge v0.5.0
- s-dftd3 v1.6.0
- toml-f v0.5.2

Currently, mctc v0.5.2 is not standard conforming (due to a `sleep()` routine), which does not exist in Fortran. (See https://github.com/grimme-lab/mctc-lib/pull/134)

### DIFF
--- a/src/dftbp/extlibs/tblite.F90
+++ b/src/dftbp/extlibs/tblite.F90
@@ -440,17 +440,22 @@ contains
     !> Parametrisation data
     type(xtb_calculator), intent(out) :: calc
 
+    type(error_type), allocatable :: exc
+
     select case(method)
     case default
       call error("Unknown method selector")
     case(tbliteMethod%gfn2xtb)
-      call new_gfn2_calculator(calc, mol)
+      call new_gfn2_calculator(calc, mol, exc)
+      if (allocated(exc)) call error("Failed to initialize gfn2 calculator")
     case(tbliteMethod%gfn1xtb)
-      call new_gfn1_calculator(calc, mol)
+      call new_gfn1_calculator(calc, mol, exc)
+      if (allocated(exc)) call error("Failed to initialize gfn2 calculator")
     case(tbliteMethod%ipea1xtb)
-      call new_ipea1_calculator(calc, mol)
+      call new_ipea1_calculator(calc, mol, exc)
+      if (allocated(exc)) call error("Failed to initialize gfn2 calculator")
     end select
-
+    
   end subroutine getCalculator
 #:endif
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -250,14 +250,14 @@ endif()
 
 if(WITH_TBLITE OR WITH_SDFTD3)
   set(MCTC_LIB_GIT_REPOSITORY "https://github.com/grimme-lab/mctc-lib.git")
-  set(MCTC_LIB_GIT_TAG "87a46cdf5281ad75909083c4f72354d54abdf95d")  # do not change manually!
+  set(MCTC_LIB_GIT_TAG "e9de066d89f250d1cfb6de3a33f0c27c0e2f855d")  # do not change manually!
   dftbp_config_hybrid_dependency(mctc-lib mctc-lib::mctc-lib "${HYBRID_CONFIG_METHODS}" "QUIET"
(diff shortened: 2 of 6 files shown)


src/dftbp/extlibs/tblite.F90
```diff
@@ -440,17 +440,22 @@ subroutine getCalculator(method, mol, calc)
...
 
+    type(error_type), allocatable :: exc
+
     select case(method)
     case default
       call error("Unknown method selector")
     case(tbliteMethod%gfn2xtb)
-      call new_gfn2_calculator(calc, mol)
+      call new_gfn2_calculator(calc, mol, exc)
+      if (allocated(exc)) call error("Failed to initialize gfn2 calculator")
```
awvwgk, line 452:
```suggestion
      if (allocated(exc)) call error("Failed to initialize gfn2 calculator: "//exc%message)
```

src/dftbp/extlibs/tblite.F90
```diff
@@ -440,17 +440,22 @@ subroutine getCalculator(method, mol, calc)
...
     case(tbliteMethod%gfn1xtb)
-      call new_gfn1_calculator(calc, mol)
+      call new_gfn1_calculator(calc, mol, exc)
+      if (allocated(exc)) call error("Failed to initialize gfn2 calculator")
```
awvwgk, line 455:
```suggestion
      if (allocated(exc)) call error("Failed to initialize gfn2 calculator: "//exc%message)
```

src/dftbp/extlibs/tblite.F90
```diff
@@ -440,17 +440,22 @@ subroutine getCalculator(method, mol, calc)
...
     case(tbliteMethod%ipea1xtb)
-      call new_ipea1_calculator(calc, mol)
+      call new_ipea1_calculator(calc, mol, exc)
+      if (allocated(exc)) call error("Failed to initialize gfn2 calculator")
```
awvwgk, line 458:
```suggestion
      if (allocated(exc)) call error("Failed to initialize gfn2 calculator: "//exc%message)
```

src/dftbp/extlibs/tblite.F90
```diff
@@ -440,17 +440,22 @@ subroutine getCalculator(method, mol, calc)
...
     end select
-
+
```
awvwgk, line 460:
```suggestion

```

awvwgk (commented):
An allocated `error_type` will always have the message field set. It is best to propagate this to the user.

aradi:
I've fixed d-orbital order, but still have problems with the 3 periodic tests, which differ considerably with current tblite. A quick attempt to calculate them with the xtb command line tool seems to deliver differing total energy values from our stored references as well as the value which I obtain witin DFTB+ with tblite v0.7. (@awvwgk any insights here?)

awvwgk:
Yes we changed a couple of things for periodic calculations in tblite:

1. D3 and D4 now use a smooth taper function at the cutoff radius, enabled by default
2. Klopman-Ohno ist now done in the Wigner-Seitz cell only instead of the tapered Coulomb variant Ewald sum, since this introduced artifacts
3. Missing contributions in the stress tensor of the multipole Ewald sum were added

We might revise 2 for the full Klopman-Ohno Ewald sum in the future as well rather than using the tapered version. We also have an PME for the dispersion which we might make available for tblite, but probably not by default since it is a little more expensive than the tapered real space summation.

aradi:
OK, thanks, so tests should recalculated I guess. I would like to be sure, though, that everything is fine at our side. Is there any explanation, why xtb gives different results as dftb+/tblite(v0.7)? Does xtb use a different periodic model as tblite?

aradi:
(And, is there any way to reproduce old results with the current version of tblite? I find it somewhat problematic, if rerunning an older DFTB+ input results in different total energies without the possibility to obtain a previous result.)

awvwgk:
The native implementation of the xTB methods for PBCs in xtb is incomplete, if used with tblite make sure that xtb is actually using tblite 0.7 and not an earlier version (note that the last xtb release 6.7.1 only works with 0.5).

Regarding reproducibility, I agree for the tapering function in 1 you can explicitly opt out. For the tapered Ewald sum in 2, we can reintroduce it if desired for reproducibility, but for 3 we fixed an actual bug in the multipole Ewald sum, so previous results were not correct.

PR #1920, opened 2026-09-18
Project lattice steps onto FixAngles, FixLengths, and Isotropic

### LINKED ISSUE #1870: GeometryOptimization fix angles broken
**Describe the bug**

Keyword can be set, but has no effect.

**To Reproduce**

Standard GeometryOptimization block with FixAngles = Yes 

**Expected behaviour**

Original geometry lattice vector orientation retained during geometry optimization.

**Additional context**

Makes driver almost unusable for various crystalline structures.

shaneraphel:
Closes #1870

`GeometryOptimization` already reads `FixAngles`, `FixLengths`, and `Isotropic` into the cartesian lattice filter. `transformDerivative` then zeros some Cartesian stress entries and forms `lat_grad = matmul(sigma, inv_lat)`, after which `transformStructure` adds the full 3×3 lattice displacement. The keywords therefore do not constrain the lattice-vector trajectory: a sheared trial step still changes cell angles.

This applies the same allowed modes on both the lattice gradient and the structure step. `FixAngles` keeps each lattice vector parallel to itself, `FixLengths` removes that parallel part, and `Isotropic` keeps a single common scale. Skipping `LatticeOpt`, or handling only one keyword, would not keep that identity.

Measured on DFTB+ `4a1c70b` with a 3×3 lattice and a sheared displacement:

| path | angle delta (deg) |
| --- | --- |
| unconstrained | −0.937, +2.017, −5.434 |
| `FixAngles` | 0, 0, 0 |
| `Isotropic` | 0, 0, 0 |


src/dftbp/geoopt/filter.F90
```diff
@@ -214,10 +218,51 @@ subroutine transformDerivative(this, coord0, latVec, gradient, stress, deriv)
...
 
 
+  !> Project a lattice increment onto FixAngles / FixLengths / Isotropic.
+  subroutine project_lattice_mode(fixAngles, fixLength, isotropic, latVec, mode)
+
+    logical, intent(in) :: fixAngles
+    logical, intent(in) :: fixLength(3)
+    logical, intent(in) :: isotropic
+    real(dp), intent(in) :: latVec(:, :)
+    real(dp), intent(inout) :: mode(:, :)
```
bhourahine, line 245: Please comment interface variables, as per the doxygen style used in other routines, particularly for variables like mode where the name is a) relatively unique to this routine and b) the name does not clearly indicate what the variable is for/does.

bhourahine (commented):
Thank you for this PR, please add your details to the AUTHORS.rst file.
The regression test for the fixed angle rational functional optimizer fails with your modification, apparently with a change in the cell volume just larger than the tolerance. Does this count as a consequence of a bug fix or is it a regression?

shaneraphel (commented):
Thanks for the review. The follow-up commit adds the doxygen comments (file convention, `!>` above each dummy) and the AUTHORS.rst entry.

On Si_6_rf: I read this as a consequence of the bug fix, not a regression:

- The test runs with FixAngles=Yes, and #1870 reported that keyword had no effect, with lattice orientations drifting. The old reference was therefore generated with drifting angles.
- With this PR the final lattice keeps the input angles exactly (120/90/90, to 4e-10 in geo_end.gen) — the behaviour #1870 asked for.
- The only tag that moves is cell_volume (+2.99e-05 against the 1e-6 tagdiff tolerance); forces, stress, energy, coords and charges all still match, worst 3.5e-07. Only this test fails (512/513 gcc, 457/458 intel); the openmpi leg runs a different 178-test subset without it.

So the optimizer now converges on the properly constrained manifold to a slightly different volume. I left the reference tag untouched — happy to update `_autotest.tag` to the new volume if you want that, or leave it for you to regenerate (note gcc-Debug and intel-RelWithDebInfo may produce slightly different volumes, so one updated value would need CI to confirm both legs).

AUTHORS.rst
```diff
@@ -94,4 +94,6 @@ contributed to DFTB+ :
 
 * Bo Thomsen (Japan Atomic Energy Agency, Japan)
 
+* Shan Yu
```
bhourahine, line 97: Do you have an affiliation you want to use for this?

shaneraphel:
I have no employer.thanks for review,i am honored to accept it.

